PR #8, opened 2021-10-29
Init package files

vojtechszocs:
In this PR:

- add Yarn latest files (using classic [`node_modules` linker strategy](https://yarnpkg.com/configuration/yarnrc#nodeLinker) for now)
- add `.gitignore` with appropriate Yarn entries (not using [Zero-Install strategy](https://next.yarnpkg.com/features/zero-installs) for now)
- add minimal `package.json` representing a monorepo with packages located under `packages` directory
- add EditorConfig file
- update OWNERS and README files

---

Follow-up - adding monorepo packages:
- core lib - adapted from `@openshift-console/dynamic-plugin-sdk`
- webpack lib - adapted from `@openshift-console/dynamic-plugin-sdk-webpack`
- sample dynamic plugin
- sample host application


OWNERS
```diff
@@ -9,4 +9,5 @@ approvers:
   - christianvogt
   - andrewballantyne
   - spadgett
-component: Dynamic Plugin SDK
\ No newline at end of file
+  - vojtechszocs
```
andrewballantyne, line 12: Ah good point! Sorry about that, I made up the owners files. We'll need to be sure we get work together on changes when new things happen or changes to existing things happen (that are public facing).

andrewballantyne (approved):

PR #10, opened 2021-11-02
Add runtime package with initial code

vojtechszocs:
- add package containing code to process dynamic plugins at runtime
- add Yarn `workspace-tools` plugin to run scripts across packages in the workspace
- add Prettier & VS Code config files
- add a minimal `PluginStore`

Following [Yarn 2+ lifecycle scripts](https://yarnpkg.com/advanced/lifecycle-scripts) and [recommendations](https://yarnpkg.com/advanced/rulebook/#packages-should-use-the-prepack-script-to-generate-dist-files-before-publishing):
- package build is executed as part of `prepack`
- package unit tests are executed as part of `prepublish`


andrewballantyne (commented):
@vojtechszocs what's the intended 'packages' structure? I gather that `lib-runtime` is the run-time library for setting up plugins, but what are the rough outlines for the other aspects?

vojtechszocs:
> @vojtechszocs what's the intended 'packages' structure? I gather that `lib-runtime` is the run-time library for setting up plugins, but what are the rough outlines for the other aspects?

@andrewballantyne My idea was to have the following monorepo package structure:
```
packages/
├── lib-runtime
├── lib-webpack
├── sample-plugin
└── sample-host-app
```

"runtime" should contain all code necessary to process plugins at runtime (assuming browser env. unless we want this to be completely env-agnostic) by the host app. This includes loading plugins, resolving their code references and managing them via plugin store.

"webpack" should contain all code necessary to build plugins based on webpack 5+ native module federation support. This has no impact on host app itself - plugins should work with host app built with webpack 4 or better.

The other two packages are for reference and e2e verification purposes.

andrewballantyne (approved):
Great first steps, thanks @vojtechszocs ! 

Once the CI has been fixed in another PR, this should be good to merge.

package.json
```diff
@@ -1,8 +1,10 @@
 {
+  "name": "monorepo",
   "license": "Apache-2.0",
-  "private": true,
   "workspaces": [
     "packages/*"
   ],
-  "packageManager": "yarn@3.1.0"
+  "scripts": {
+    "test": "yarn workspaces foreach run test"
```
florkbr, line 8: @vojtechszocs @andrewballantyne if we are planning on using a mono-repo structure, should we investigate using something like [Lerna](https://github.com/lerna/lerna)? It'll make running things like this easier IMO.

vojtechszocs: @florkbr For now, I've used [`yarn workspaces foreach`](https://yarnpkg.com/cli/workspaces/foreach) CLI (provided via Yarn `workspace-tools` plugin) instead of Lerna for the sake of simplicity.

I agree that we should look into tools like Lerna in the future, based on what we need to do with our packages. This monorepo should be quite simple - some packages are libraries (to be published to npm package registry) while others are helpers. It makes sense to have `build` and `test` scripts for all of these packages.

andrewballantyne:
@florkbr Looks like the test failure here is the lack of a dependency install. 

```
➤ YN0000: Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)
```

build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+
+yarn install
+yarn build
```
andrewballantyne, line 5: 👍

test.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+
+# TODO lint
+yarn test
```
andrewballantyne, line 5: 👍 
This is obv expecting a `build.sh` to be executed first.

andrewballantyne (commented):
@florkbr FYI for the CI setup.

vojtechszocs:
> @florkbr Looks like the test failure here is the lack of a dependency install.
> 
> ```
> ➤ YN0000: Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)
> ```

FYI `yarn install` is run as part of `build.sh` script.

vojtechszocs:
Setup might be borked, will create a new PR.

PR #13, opened 2021-11-03
[WIP]  Do not merge - Add initial code - testing pipeline

florkbr:


package.json
```diff
@@ -1,11 +1,10 @@
...
-  "private": true,
-  "scripts": {
-    "test": "echo 'Write some tests!'"
-  },
   "workspaces": [
     "packages/*"
   ],
-  "packageManager": "yarn@3.1.0"
+  "scripts": {
+    "test": "yarn install && yarn workspaces foreach run install && yarn workspaces foreach run test"
```
andrewballantyne, line 8: As discussed on slack, but for tying it to this PR... this won't work for us. We'll likely need to follow the effort done in OpenShift Console and create a `build.sh` that handles things like this.

vojtechszocs: #10 updated - it now comes with `build.sh` and `test.sh` scripts.

florkbr: :+1: closing this PR then

PR #15, opened 2021-11-03
Add runtime pkg with initial code

vojtechszocs:
- add package containing code to process dynamic plugins at runtime
- add Yarn `workspace-tools` plugin to run scripts across packages in the workspace
- add Prettier & VS Code config files
- add a minimal `PluginStore`

Following [Yarn 2+ lifecycle scripts](https://yarnpkg.com/advanced/lifecycle-scripts) and [recommendations](https://yarnpkg.com/advanced/rulebook/#packages-should-use-the-prepack-script-to-generate-dist-files-before-publishing):
- package build is executed as part of `prepack`
- package unit tests are executed as part of `prepublish`


andrewballantyne (approved):
/lgtm

PR #16, opened 2021-11-10
Add initial ESLint configuration

fhlavac:
Took the ESLint approach from Console project. It puts all ESLint related dependencies, configs and recommended rule overrides into a single monorepo pkg, e.g. eslint-plugin-openshift. This means all ESLint stuff is centralized in a single pkg which is easier to manage and upgrade if necessary.

Added two ESLint presets; each preset combines rules of many ESLint plugins (based on their recommended configs) into a single, focused linting rule set: `react-typescript-prettier` (for lib-runtime, sample-plugin and sample-host-app) and `node-typescript-prettier` for (lib-webpack,  sample-plugin and sample-host-app).

**Latest changes:**
- use global `.eslintignore` file
- don't use `{}` type since it's not what it appears to be (see here for detailed explanation)
- rename package `eslint-plugin-openshift` to `eslint-plugin-internal`
- all packages are now linted, including the ESLint one
- Yarn 2+ allows sharing scripts between packages if script name includes `:` character (see here for reference)
- add `react` dependency to `lib-runtime` pkg due to "detect auto" default in React related lint rules
- for now, assume Node.js version v14 (due to current CI builder image) - however note that current active Node.js LTS is already v16


andrewballantyne (commented):
This looks like a wip, so I won't do a real review here, just a couple quick thoughts:

- Use yarn (we shouldn't have a package-lock.json)
- Let us scrub out all references to OpenShift; we are in that org on GitHub but I think it's best if we don't bring over any of that into the SDK

fhlavac:
@andrewballantyne yeah, that's just an early draft. Thanks for the comments, I will work on it today.

andrewballantyne (commented):
Not sure if you have fixed your error in the description or not... but I think it's related to the parsing of your lint file... I just `.eslintignore` it in hac-core-frontend: https://github.com/RedHatInsights/hac-core-frontend/blob/master/.eslintignore#L7

packages/eslint-plugin-openshift/rule-overrides/react-typescript-prettier.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
```
vojtechszocs, line 1: Nit: please use 2 spaces for indentation, this is part of EditorConfig we use.

I recommend using VS Code since we already have `.vscode/settings.json` config in the repo.

Also, I'd recommend installing the relevant VS Code extensions to take full advantage of ^^ config:

```sh
code --install-extension EditorConfig.EditorConfig
code --install-extension dbaeumer.vscode-eslint
code --install-extension esbenp.prettier-vscode
```

packages/lib-runtime/src/PluginStore.ts
```diff
@@ -77,11 +83,13 @@ export class PluginStore<TLoadedExtension extends LoadedExtension> {
    */
   addPlugin(metadata: PluginMetadata, extensions: TLoadedExtension[]) {
     if (this.loadedPlugins.has(metadata.name)) {
+      // eslint-disable-next-line no-console
```
vojtechszocs, line 86: The `no-console` rule can be disabled once, for the whole file.

vojtechszocs (changes requested):

packages/lib-runtime/src/PluginStore.ts
andrewballantyne: Do not recommend whole file disabling of lints unless it is a legacy file that predates configurations or there is a unique use-case for it (nothing comes to the top of my head). However, uniquely, `no-console` should always be on for a file. It forces the calls of `console.` to be validated by the developer. Debug statements always will be things we want to vet so we don't leak those into the build.

vojtechszocs:
I have one more change incoming, please don't merge yet.

packages/lib-runtime/src/PluginStore.ts
vojtechszocs: Fair enough, I'll move back to disable-per-line for `no-console` rule.

vojtechszocs:
I think we addressed all comments, let's remove the WIP label :smiley:

@types/sdk-common/index.d.ts
```diff
@@ -0,0 +1 @@
+declare type AnyObject = Record<string, unknown>;
```
vojtechszocs, line 1: Reference: https://github.com/typescript-eslint/typescript-eslint/issues/2063#issuecomment-675156492

.eslintignore
```diff
@@ -0,0 +1,3 @@
+**/node_modules
+**/dist
+**/.eslintrc.js
```
vojtechszocs, line 3: Despite being ignored, `.eslintrc.js` files should still get properly formatted by the editor, assuming it has support for Prettier and EditorConfig.

packages/eslint-plugin-internal/configs/base-node.js
```diff
@@ -0,0 +1,18 @@
+const ecmaVersion =
+  require('../../../tsconfig-bases/node-cjs.json').compilerOptions.target.substring(2);
```
vojtechszocs, line 2: The idea behind this is to avoid duplicating ECMA version between tsconfigs vs. ESLint configs.

packages/eslint-plugin-internal/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "eslint-plugin-internal",
+  "version": "0.0.0-fixed",
+  "description": "Internal ESLint configs and scripts",
+  "main": "index.js",
+  "scripts": {
+    "lint:run": "cd $INIT_CWD && eslint --ignore-path $PROJECT_CWD/.eslintignore --max-warnings 0 --color",
```
vojtechszocs, line 7: Reference: https://yarnpkg.com/advanced/lifecycle-scripts/#environment-variables

packages/eslint-plugin-internal/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "eslint-plugin-jsx-a11y": "^6.5.1",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-prettier": "^4.0.0",
+    "eslint-plugin-promise": "^5.1.1",
+    "eslint-plugin-react": "^7.27.0",
+    "eslint-plugin-react-hooks": "^4.3.0",
+    "prettier": "^2.4.1"
+  },
+  "engines": {
+    "node": ">=14"
```
vojtechszocs, line 29: FYI this is the recommended way for ESLint `node` plugin to specify compatible Node.js version.

packages/lib-runtime/package.json
```diff
@@ -1,14 +1,20 @@
...
+  "name": "@openshift/dynamic-plugin-sdk-runtime",
   "version": "0.0.0-alpha1",
+  "description": "Allows loading, managing and interpreting plugins at runtime",
   "scripts": {
     "prepack": "yarn build",
     "prepublish": "yarn test",
     "build": "rm -rf dist && tsc --outDir dist",
-    "test": "echo 'Write some tests!'"
+    "test": "echo 'Write some tests!'",
+    "lint": "yarn lint:run --ext .ts,.tsx ."
```
vojtechszocs, line 10: Here we assume the lib packages will be written purely in TypeScript.

andrewballantyne (changes requested):
I wonder if we want to add what we did in hac-core-frontend to prevent odd bundling problems with PF React-Icons and lodash. It's my understanding (2nd hand learned, haven't tested directly) that they don't tree-shake well and thus increase bundle sizes if we use them.

From [hac-core-frontend/.eslintrc.js](https://github.com/RedHatInsights/hac-core-frontend/blob/master/.eslintrc.js#L49-L58)
```
    'no-restricted-imports': ['error',
      {
        name: '@patternfly/react-icons',
        message: 'Don\'t use group imports. Use @patternfly/react-icons/dist/js/icons/(kebab-case-name) instead.'
      },
      {
        name: 'lodash',
        message: 'Don\'t use group imports. Use lodash/(funcName) instead.'
      }
    ],
```

Which effectively stops group imports:
```
import { add, filter, map } from 'lodash';
// or
import { StarIcon, OtherIcon } from '@patternfly/react-icons';
```
And lints you down to a singular direct import of each one.

.eslintignore
```diff
@@ -0,0 +1,3 @@
+**/node_modules
+**/dist
```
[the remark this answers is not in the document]
andrewballantyne: Feel free to ignore if want to just push it down the road

packages/lib-runtime/package.json
andrewballantyne: I see no reason we would do anything else. I think OpenShift Console only has `.js` in there because of the legacy code before TS was added.

packages/eslint-plugin-openshift/rule-overrides/react-typescript-prettier.js
fhlavac: added them, thank you :-)

fhlavac:
@vojtechszocs thanks for all the comments, looks good!

andrewballantyne (approved):

PR #17, opened 2021-11-22
Add a linter to stop the 'any' type being used

andrewballantyne:
/cc @vojtechszocs 

Adding no explicit `any` now so we don't have any slip up from our definitions as we grow. We are an SDK, not an App... luxury on types should not be granted to us.


andrewballantyne:
Closing PR. We have it configured as a warning and do not allow warnings in the SDK.

PR #18, opened 2021-11-24
Add PluginLoader and connect it with PluginStore

vojtechszocs:
- `PluginLoader` is responsible for loading plugin assets from remote sources
- `PluginStore` is responsible for managing plugins and providing access to runtime plugin information and extensions
- `PluginStoreClient` is the client interface for `PluginStore` consumers

### Example usage

```ts
// 1. host application: set up your plugin store

const loader = new PluginLoader(/* custom loader options */);
loader.registerPluginEntryCallback();

const store = new PluginStore(/* custom store options */);
store.setLoader(loader);
```

```ts
// 2. host application: start loading some plugins

store.loadPlugin(pluginBaseURL);
```

```ts
// 3. plugin store consumers: subscribe to relevant events

store.subscribe(() => {
  const extensions = store.getExtensions();
  // ...
}, [PluginStoreEventType.ExtensionsChanged]);
```


packages/lib-runtime/src/utils/fetch.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+/**
+ * Basic implementation of `FetchResource` that uses the `fetch` API.
+ */
+export const basicFetch: FetchResource = async (url, method) => {
+  const response = await fetch(url, { method });
+
+  if (response.ok) {
+    const responseText = await response.text();
```
andrewballantyne, line 21: Intent is to have a JSON Fetch? I'd expect very rarely are we wanting raw text.

packages/lib-runtime/src/utils/logger.ts
```diff
@@ -0,0 +1,17 @@
...
+
+type Logger = Record<'info' | 'warn' | 'error', LogFunction>;
+
+const isProdEnv = process.env.NODE_ENV === 'production';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop: LogFunction = () => {};
+
+export const consoleLogger: Logger = {
+  info: isProdEnv ? noop : console.info,
```
andrewballantyne, line 14: You have lodash, why not use it? `_.noop`

andrewballantyne (commented):
Will review the meat tomorrow... just what I saw skimming the code.

packages/lib-runtime/src/utils/fetch.ts
vojtechszocs: @andrewballantyne The intent here was that plugin SDK (its "runtime" library specifically) would work with a generic `FetchResource` compatible function, allowing to fetch arbitrary resources over HTTP and return their contents as plain text.

```ts
export type FetchResource = (url: string, method: string) => Promise<string>;
```

The `basicFetch` implementation is used as a default fallback. Specific host apps, like Console, will probably want to provide a custom implementation, like Console's `coFetch` which adds some HTTP headers and has retry & timeout logic built into it.

For now, `FetchResource` is used only in `PluginLoader.getPluginManifest`.

Is there a specific reason for doing `response.json()` instead of calling `JSON.parse` on `response.text()`? (The latter allows us to have a single `FetchResource` interface for fetching & processing all kinds of resources, not only JSON.)

packages/lib-runtime/src/utils/logger.ts
vojtechszocs: Good point, will do the change.

packages/lib-runtime/src/PluginLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * Register the global function used by plugin entry scripts.
+   */
+  registerPluginEntryCallback(getWindow: () => typeof window = _.constant(window)) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const anyWindow = getWindow() as any;
+    const callback = this.options.entryCallbackName;
+
+    if (anyWindow[callback] === 'function') {
```
vojtechszocs, line 201: Missing `typeof` here, will fix.

packages/lib-runtime/src/store/PluginStore.ts
```diff
@@ -0,0 +1,309 @@
...
+type LoadedPluginInfo = {
+  pluginName: string;
+  status: 'loaded';
+  metadata: LoadedPlugin['metadata'];
+  enabled: boolean;
+};
+
+type NotLoadedPluginInfo = {
+  pluginName: string;
+  status: 'pending' | 'failed';
```
christianvogt, line 25: Create an enum for the status.

packages/lib-runtime/src/store/PluginStore.ts
```diff
@@ -0,0 +1,309 @@
...
+import { consoleLogger } from '../utils/logger';
+import { PluginLoader } from './PluginLoader';
+
+/**
+ * Client interface for `PluginStore` consumers.
+ */
+export interface PluginStoreClient<TLoadedExtension extends LoadedExtension> {
+  subscribe: (listener: VoidFunction, eventTypes: PluginStoreEventType[]) => VoidFunction;
+  getExtensions: () => TLoadedExtension[];
+  getPluginInfo: () => (LoadedPluginInfo | NotLoadedPluginInfo)[];
```
christianvogt, line 13: I think any type exposed on a public type should themselves be exposed.

packages/lib-runtime/src/store/PluginStore.ts
```diff
@@ -0,0 +1,309 @@
...
+   */
+  setPluginEnabled(pluginName: string, enabled: boolean) {
+    if (!this.loadedPlugins.has(pluginName)) {
+      consoleLogger.warn(
+        `Attempt to ${enabled ? 'enable' : 'disable'} plugin ${pluginName} which is not loaded yet`,
+      );
+      return;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
christianvogt, line 231: Disabling eslint rule shouldn't be necessary

packages/lib-runtime/src/store/PluginStore.ts
christianvogt: Might be good to establish some guidelines when to export types for consistency across the SDK. That could be a longer discussion and shouldn't hold up this PR in the interim.

christianvogt: cc @andrewballantyne  ^^

christianvogt:
overall it lgtm

packages/lib-runtime/src/store/PluginStore.ts
christianvogt: ha missed the `!` at the _end_ of the line. So nm!

rawagner (approved):
looks good

packages/lib-runtime/src/utils/fetch.ts
andrewballantyne: No this is fair... this was before your rundown of the code.

PR #19, opened 2021-11-29
Update test.sh script

vojtechszocs:
Removed `build.sh` and updated `test.sh` to follow the standard CI project test flow:
1. install dependencies
2. post-install checks, e.g. ensure that the committed `yarn.lock` file is not outdated
3. analyze code for potential problems
4. run unit tests
5. build all packages
6. run end-to-end tests

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-set -exuo pipefail
-
-yarn install
-yarn build
--- a/test.sh
+++ b/test.sh
@@ -1,8 +1,25 @@
 #!/bin/bash
 set -exuo pipefail
 
-# TODO remove this once CI runs build script prior to test script
-./build.sh
+print_error() { printf "%s\n" "$*" >&2; }
 
+# Install dependencies
+yarn install
+
+# Check for outdated yarn.lock file
+if [[ -n "$(git status --porcelain -- yarn.lock)" ]]; then
+  print_error "Outdated yarn.lock file, commit changes to fix"
+  exit 1
+fi
+
+# Analyze code for potential problems
 yarn lint
+
+# Run unit tests
 yarn test
+
+# Build packages
+yarn build
+
+# Run end-to-end tests
+# TODO


fhlavac:
@vojtechszocs looks good to me, thank you :slightly_smiling_face:

andrewballantyne (commented):
Two comments:

1. Is the concern that minor versions will break the semver contract and thus break us? Minor versions are supposed to be backwards compatible
2. I'm kinda surprised `./build.sh` didn't cause this to break (cc @florkbr)... We are [using it theoretically](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/dynamic-plugin-sdk/openshift-dynamic-plugin-sdk-main.yaml#L1)

vojtechszocs:
> 1. Is the concern that minor versions will break the semver contract and thus break us? Minor versions are supposed to be backwards compatible

In theory, yes :smiley: in practice, minor version bumps might cause breakages because people do make mistakes.

Also, dependencies are not isolated units of vendor code - they can, and often do depend on each other. It might happen, for example, that `react` version will get out of sync with `@types/react`. Always bumping to latest stable version (especially when done on _per-single-dependency_ basis, instead of _per-logical-dependency-group_ basis) can lead to problems.

(Sadly, `package.json` has no concept of logical dependency groups that would guard consistency _across_ multiple related vendor packages. This responsibility therefore falls on the developer.)

So my main concern was to be very conservative when it comes to dependencies, given that we're developing a library to be used by multiple applications and lots of plugins. If you think I'm over-cautious, we can keep using the `^` ranges, it's just that I personally prefer the `~` ranges for better stability.

> 2. I'm kinda surprised `./build.sh` didn't cause this to break (cc @florkbr)... We are [using it theoretically](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/dynamic-plugin-sdk/openshift-dynamic-plugin-sdk-main.yaml#L1)

Right, but the `build.sh` script is not called by CI when testing our PRs, so I've expected this would pass :smile: 

Anyway, I think it's fine to have a single script like `test.sh` to verify all packages within the monorepo, including unit and e2e tests. We can update the CI job config and remove the `build.sh` part, but it seems it's not in effect anyway.

andrewballantyne:
> RE Number 1

I think this is a weird stance to take, personally. Nothing at all, ever, stops them from breaking API. Patches can break API. I think NPM stops zero version update of code, but if they can do that... THAT can break API. 

I think we either agree to the semver, or not. If we don't... not sure what our alternative is... maybe just go very specific with the version `1.1.1` instead of `~1.1.1` or `^1.1.1`.

Also, we have a lock file. Isn't the purpose behind this is to never update unless we touch the dependencies in some way? Like update another dependency that overlaps with ours or like delete the lock and rebuild it? Because I don't ever recall it auto updating as I have ran `yarn install` on its own.

> RE Number 2

Quite bizarre if you ask me -- but if it works it works. We should 100% definitely delete the reference in the release repo though. Since there will not be anything called `build.sh` anymore.

andrewballantyne (approved):

PR #20, opened 2021-11-29
Add PluginStoreContext

vojtechszocs:
Code in `PluginStoreContext.tsx` is inspired by article [How to use React Context effectively](https://kentcdodds.com/blog/how-to-use-react-context-effectively).

> Note that I'm _NOT_ exporting `CountContext`. This is intentional. I expose only one way to provide the context value and only one way to consume it. This allows me to ensure that people are using the context value the way it should be and it allows me to provide useful utilities for my consumers.

`react-esm.json` TypeScript config base adapted to support JSX syntax (applies to `.tsx` files).

Note that by default, TypeScript does _not_ allow plain JavaScript code to be imported ([`allowJs`](https://www.typescriptlang.org/tsconfig#allowJs) defaults to `false`).

### DIFF
--- a/packages/lib-runtime/package.json
+++ b/packages/lib-runtime/package.json
@@ -17,6 +17,7 @@
   "devDependencies": {
     "@types/lodash-es": "^4.17.5",
     "@types/node": "^12.20.36",
+    "@types/react": "^17.0.37",
     "eslint-plugin-internal": "0.0.0-fixed",
     "typescript": "^4.4.4"
   }
--- a/packages/lib-runtime/src/store/PluginStore.ts
+++ b/packages/lib-runtime/src/store/PluginStore.ts
@@ -7,10 +7,10 @@ import { PluginLoader } from './PluginLoader';
 /**
  * Client interface for `PluginStore` consumers.
  */
-export interface PluginStoreClient<TLoadedExtension extends LoadedExtension> {
+export interface PluginStoreClient<TLoadedExtension extends LoadedExtension = LoadedExtension> {
   subscribe: (listener: VoidFunction, eventTypes: PluginStoreEventType[]) => VoidFunction;
   getExtensions: () => TLoadedExtension[];
-  getPluginInfo: () => (LoadedPluginInfo | NotLoadedPluginInfo)[];
+  getPluginInfo: () => PluginInfo[];
 }
 
 type LoadedPluginInfo = {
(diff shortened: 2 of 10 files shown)


yarn.lock
```diff
@@ -135,6 +135,7 @@ __metadata:
   dependencies:
     "@types/lodash-es": ^4.17.5
     "@types/node": ^12.20.36
+    "@types/react": ^17.0.37
```
vojtechszocs, line 138: Note to self: change to `~` after #19 merges.

packages/lib-runtime/package.json
```diff
@@ -17,6 +17,7 @@
   "devDependencies": {
     "@types/lodash-es": "^4.17.5",
     "@types/node": "^12.20.36",
+    "@types/react": "^17.0.37",
```
vojtechszocs, line 20: Note to self: change to `~` after #19 merges.

vojtechszocs:
cc @christianvogt

packages/lib-runtime/src/store/PluginStore.ts
```diff
@@ -25,6 +25,8 @@ type NotLoadedPluginInfo = {
   status: 'pending' | 'failed';
 };
 
+export type PluginInfo = LoadedPluginInfo | NotLoadedPluginInfo;
```
christianvogt, line 28: Do we want to enforce define before use?  https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-use-before-define.md

vojtechszocs: You can inspect the actual ESLint config for this file:
```sh
cd packages/lib-runtime
yarn lint:print-config src/store/PluginStore.ts
```

The relevant bits:
```jsonc
"no-use-before-define": [
  "off",
  {
    "functions": true,
    "classes": true,
    "variables": true
  }
],
"@typescript-eslint/no-use-before-define": [
  "error"
],
```

[`@typescript-eslint/no-use-before-define`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-use-before-define.md) rule is active, using its default configuration.

By default, it doesn't care about type-related symbols, but it _does_ care about JS symbols, for example:
```ts
const a = x(); // triggers error: @typescript-eslint/no-use-before-define
const x = () => true;
```

Personally, I'm in favor of not enforcing this for type-related symbols.

christianvogt: fair. Just noticed it.

PR #21, opened 2021-12-03
Add React hooks for working with PluginStore

vojtechszocs:
### Major updates
- remove `PluginStoreClient` interface
- add `PluginConsumer` interface with methods: `subscribe`, `getExtensions`, `getPluginInfo`
- add `PluginManager` interface with methods: `loadPlugin`, `setPluginEnabled`
- make `PluginStore` implement both `PluginConsumer` and `PluginManager`
- replace `usePluginStore` hook with `usePluginConsumer` and `usePluginManager` hooks
- add `useExtensions` and `usePluginInfo` hooks, move common bits into `usePluginSubscription` hook
- helper `usePluginSubscription` hook supports data equality comparison, re-renders only on data change

### Minor updates
- remove generic type args on `PluginStore` since consumers will use typed hooks anyway
- update `extensionType` object schema (must have `<app-or-plugin-specifier>.` prefix)
- [ESLint] prefer using [type-only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export)
- add more code docs


packages/lib-runtime/src/store/usePluginInfo.ts
```diff
@@ -0,0 +1,20 @@
...
+  const infoEntries = usePluginSubscription(
+    (pluginConsumer) => pluginConsumer.getPluginInfo(),
+    _.isEqual,
+    [PluginEventType.PluginInfoChanged],
+  );
+
+  return infoEntries;
+};
+
+export const noPluginsPending = (infoEntries: PluginInfoEntry[]) =>
```
andrewballantyne, line 19: Interesting, do you see this being something they want to do outside of `usePluginInfo`? I'm thinking we can just make this a hook that internally looks it up through `usePluginInfo`.

andrewballantyne (changes requested):
A couple comments, but looking pretty good from my point of view. I do have a requested change though.

/cc @christianvogt

packages/lib-runtime/src/store/useExtensions.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as React from 'react';
+import type { Extension, LoadedExtension, ExtensionPredicate } from '../types/extension';
+import { PluginEventType } from '../types/store';
+import { usePluginSubscription } from './usePluginSubscription';
+
+/**
+ * TODO docs
+ */
+export const useExtensions = <TExtension extends Extension>(
+  predicate?: ExtensionPredicate<TExtension>,
```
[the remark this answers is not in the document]
vojtechszocs: I've discussed this with @christianvogt in the past.

Originally, I've advocated for multiple predicates / type guards that operate on the same extension type (like `E` in code snippet below):

```ts
export const useExtensions = <E extends Extension>(
  ...typeGuards: ExtensionTypeGuard<E>[]
): LoadedExtension<E>[] => {
  // ...
};
```

Looking at the latest Console codebase, among all (~70) `useExtensions` calls that target specific extensions, there's only one that uses multiple type guards:

```ts
// in public/components/nav/perspective-nav.tsx
const allItems = useExtensions<PluginNavSection | NavItem | Separator>(isNavSection, isNavItem);
```

but that ^^ code can be reworked to use a single logical type guard representing a union of "nav section" and "nav item" extensions, basically a composite type guard, see my comment https://github.com/openshift/console/pull/10476#discussion_r752565954 for details.

**TL;DR** - I see multiple predicates / type guards as an unnecessary complexity.

If we need to consume a union of (multiple types of) extensions, we can do e.g.

```ts
// composite type guard
const isNavItemOrSection = (e: Extension): e is NavItem | NavSection =>
  isNavItem(e) || isNavSection(e);

// inferred type of navExtensions: LoadedExtension<NavItem | NavSection>[]
const navExtensions = useExtensions<NavItem | NavSection>(isNavItemOrSection);
```

packages/lib-runtime/src/store/usePluginInfo.ts
vojtechszocs: FYI, I've removed the helper `noPluginsPending` function in the latest commit, since code like

```ts
infoEntries.every((e) => e.status !== 'pending')
```

isn't really infra or SDK related, it's just processing of the `usePluginInfo` hook result, which can be done directly in the consuming code.

vojtechszocs: > Interesting, do you see this being something they want to do outside of `usePluginInfo`?

It's code that post-processes the result of `usePluginInfo` hook.

If we think it has high value to consumers, we can make a hook out of it. Personally, I'd leave this to the consumers.

packages/lib-runtime/src/store/useExtensions.ts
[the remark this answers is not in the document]
andrewballantyne: Works for me. Worst case they can always do multiple calls if they are unlike types.

packages/lib-runtime/src/store/PluginStore.ts
```diff
@@ -183,53 +110,76 @@ export class PluginStore<TLoadedExtension extends LoadedExtension = LoadedExtens
...
+        return acc;
+      },
+      [] as PluginInfoEntry[],
+    );
+
+    return [...loadedEntries, ...failedEntries, ...pendingEntries];
+  }
+
+  async loadPlugin(baseURL: string) {
+    if (this.loader === undefined) {
```
christianvogt, line 157: You should be checking for null values instead of explicitly undefined.
```suggestion
    if (this.loader == null) {
```
or
```suggestion
    if (!this.loader) {
```

packages/lib-runtime/src/store/PluginStore.ts
```diff
@@ -183,53 +110,76 @@ export class PluginStore<TLoadedExtension extends LoadedExtension = LoadedExtens
...
+  }
 
-    if (this.loadedPlugins.has(pluginName)) {
-      consoleLogger.warn(`Attempt to re-add an already loaded plugin ${pluginName}`);
-      return false;
-    }
+  getPluginInfo() {
+    const loadedEntries = Array.from(this.loadedPlugins.keys()).reduce((acc, pluginName) => {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const plugin = this.loadedPlugins.get(pluginName)!;
```
christianvogt, line 128: Could you do this instead to remove the need to avoid the `no-non-null-assertion`
```suggestion
    const loadedEntries = Array.from(this.loadedPlugins.values()).reduce((acc, plugin) => {
      const { pluginName } = plugin.metadata;
```

packages/lib-runtime/src/store/PluginStoreContext.tsx
```diff
@@ -14,19 +16,38 @@ export const PluginStoreProvider: React.FC<PluginStoreProviderProps> = ({ store,
...
+  );
+};
+
+/**
+ * React hook for exposing the {@link PluginManager} interface.
+ */
+export const usePluginManager = (): PluginManager => {
+  const store = React.useContext(PluginStoreContext);
+
+  if (store === undefined) {
```
christianvogt, line 45:
```suggestion
  if (store == null) {
```
or
```suggestion
  if (!store) {
```

packages/lib-runtime/src/utils/objects.ts
```diff
@@ -0,0 +1,27 @@
...
+) => {
+  const methods = _.pick(target, methodNames);
+  const delegate: AnyObject = {};
+
+  Object.keys(methods).forEach((methodName) => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    delegate[methodName] = (target as any)[methodName];
+  });
+
+  return delegate as Pick<TObject, TObjectMethods>;
```
christianvogt, line 26: You should probably make an assertion that the `methodNames` exist on `target`.

Not sure why you `pick` from the the `target` only to call `Object.keys(methods)`.

```suggestion
  const delegate = {} as Pick<TObject, TObjectMethods>;

  methodNames.forEach((methodName) => {
    const fn = (target as any)[methodName];
    if (typeof fn !== 'function') {
      throw new Error('...');
    }
    delegate[methodName] = fn;
  });

  return delegate;
```

packages/lib-runtime/src/store/usePluginSubscription.ts
```diff
@@ -0,0 +1,94 @@
...
+        const prevData = hookResultRef.current;
+        const nextData = getDataRef.current(pluginConsumer);
+        const dataChanged = !isSameDataRef.current(prevData, nextData);
+
+        if (dataChanged) {
+          hookResultRef.current = nextData;
+        }
+
+        if (isMountedRef.current && dataChanged) {
+          forceRender();
```
christianvogt, line 56: Could this not be done with `useState`?
```
const [hookResult, setHookResult] = React.useState<TPluginData>(getData(pluginConsumer));
```

packages/lib-runtime/src/store/usePluginInfo.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * The hook's result is guaranteed to be referentially stable across re-renders.
+ */
+export const usePluginInfo = (): PluginInfoEntry[] => {
+  const getData = React.useCallback(
+    (pluginConsumer: PluginConsumer) => pluginConsumer.getPluginInfo(),
+    [],
+  );
+
+  const eventTypes = React.useMemo(() => [PluginEventType.PluginInfoChanged], []);
```
andrewballantyne, line 20: Any reason this isn't just a file constant?

andrewballantyne (commented):
I think I missed this with my last review... odd to use Memo for a constant imo.

packages/lib-runtime/src/store/useExtensions.ts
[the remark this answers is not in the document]
christianvogt: I agree that the current api is an unnecessary complexity.

cc @spadgett

packages/lib-runtime/src/store/usePluginInfo.ts
vojtechszocs: In general, React hooks should be sensitive (responsive) to changes to their arguments. That's because the hook's arguments typically have a direct impact on the hook's result.

Let's take the `usePluginSubscription` hook as an example. It has 3 arguments: `getData` (function), `isSameData` (function) and `eventTypes` (array of strings).

If we call this hook like so:

```ts
const infoEntries = usePluginSubscription(
  (pluginConsumer) => pluginConsumer.getPluginInfo(),
  (prevData, nextData) => _.isEqual(prevData, nextData),
  [PluginEventType.PluginInfoChanged],
);
```

we're effectively creating brand new `getData`, `isSameData` and `eventTypes` argument objects _every time the hook gets called_. And because of that, the `usePluginSubscription` hook will un- and re-subscribe to `PluginStore` every time we call it that way, which is highly inefficient and just plain wrong.

What we want to do is to ensure that all `usePluginSubscription` arguments change (in terms of the referenced object) only when they need to. In our case, there's nothing that impacts these arguments, so we memoize their values using an empty dependency array `[]` which tells React the value never changes.

```
const getData = React.useCallback(
  (pluginConsumer: PluginConsumer) => pluginConsumer.getPluginInfo(),
  [], // no dependencies, getData reference never changes
);
```

From React docs on [`useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback)

> `useCallback(fn, deps)` is equivalent to `useMemo(() => fn, deps)`

packages/lib-runtime/src/store/PluginStore.ts
vojtechszocs: `this.loader` is marked as private and its value is set via `setLoader` method so this check should apply there, I think.

I don't like `!falsy` expressions too much because [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) encompasses a lot of different value types.

I'll update this to `instanceof` check since `this.loader` must be an object created via `PluginLoader` constructor function.

packages/lib-runtime/src/store/usePluginInfo.ts
christianvogt: @vojtechszocs what @andrewballantyne is asking here is why not define this as a constant _outside_ the component instead of creating the new array each time and using `useMemo`

packages/lib-runtime/src/store/PluginStore.ts
christianvogt: Ah `loader` has type `| undefined`. This is ok then. Sorry I missed that.
No need for `instanceof` check since we have typescript to handle that for us.

packages/lib-runtime/src/store/PluginStoreContext.tsx
christianvogt: Also missed here that context has type `| undefined`.

packages/lib-runtime/src/store/usePluginInfo.ts
andrewballantyne: Aww, I came here to say that @christianvogt -- stealing all my thunder! 😛

packages/lib-runtime/src/store/PluginStore.ts
vojtechszocs: ESLint rules can't be 100% accurate and it's perfectly fine to suppress them if necessary.

That said - agreed, will do the change :smiley:

packages/lib-runtime/src/store/PluginStoreContext.tsx
vojtechszocs: Similar to above, I'll update the check to be an `instanceof` check (which also returns `false` on `null` or `undefined` values).

packages/lib-runtime/src/utils/objects.ts
vojtechszocs: Great stuff, will do the change.

packages/lib-runtime/src/store/usePluginSubscription.ts
vojtechszocs: Yes, I think so, good catch!

vojtechszocs: Actually, doing it this way, we'd also need to put `hookResult` as a dependency of `trySubscribe` which would result in unnecessary recomputations of the memoized `trySubscribe` function.

Overall, I'd say `useRef` is great for infra-level hooks that want full control over the value and when to re-render, while `useState` is great for consuming app components that generally want re-render upon value change.

vojtechszocs: Also, there'd be the issue when the `PluginStore` listener would fire but the component is no longer mounted. We still want to keep track of the current data, while not wanting to re-render since the component is not mounted. We can't have this level of control with `useState` hook.

packages/lib-runtime/src/store/usePluginInfo.ts
vojtechszocs: Sorry, seems like I managed to completely misunderstand @andrewballantyne's comment!

Agreed, will do the change.

packages/lib-runtime/src/store/usePluginSubscription.ts
christianvogt: If you unsubscribed in an effect disposer function, how would this listener be invoked?
With `useState` you can pass a function to the `setter` function to get the current state thus removing it as a dependency in your function.

vojtechszocs: Sounds good, we can use the so-called function updater form of `useState` value setter function, as described in [React docs](https://reactjs.org/docs/hooks-reference.html#usestate).

packages/lib-runtime/src/store/useExtensions.ts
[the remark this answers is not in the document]
spadgett: I'm good with the change.

PR #24, opened 2021-12-22
Add commonFetch utilities into lib-utils package

vojtechszocs:
This PR adds basic resource fetch utilities into `lib-utils` package.

- `ResourceFetch` is the basic fetch interface
```ts
type ResourceFetch = (url: string, options?: RequestInit) => Promise<Response>;
```
- add `lib-utils` config API, similar to `configSetup` in @invincibleJai's Console PR [#10707](https://github.com/openshift/console/pull/10707)
```ts
setUtilsConfig({ appFetch, /* k8sBasePath, etc. */ });
```
- add `lib-utils` fetch APIs: `commonFetch`, `commonFetchText`, `commonFetchJSON`


vojtechszocs:
/hold

Need to fix the build error.

packages/lib-utils/src/utils/common-fetch.ts
```diff
@@ -0,0 +1,89 @@
...
+    applyDefaults(options, { headers: { Accept: 'text/plain' } }),
+    timeout,
+  );
+
+  const responseText = await response.text();
+
+  return responseText ?? '';
+};
+
+export const commonFetchJSON: CommonFetchJSON<unknown> = async (
```
vojtechszocs, line 43: Not sure if this should be `CommonFetchJSON<any>`, I'll look into this.

packages/lib-utils/src/config.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+/**
+ * Get the {@link UtilsConfig} reference.
+ *
+ * Throws an error if the reference isn't already set.
+ */
+export const getUtilsConfig = (): UtilsConfig => {
+  if (config === undefined) {
+    throw new Error('setUtilsConfig has not been called');
```
vojtechszocs, line 37: The reason for throwing an error here is to avoid subsequent errors, e.g. `getUtilsConfig().appFetch` being undefined.

I think it's generally better to fail early when it comes to proper SDK utils initialization.

packages/common/src/errors/CustomError.ts
```diff
@@ -22,10 +22,3 @@ export class CustomError extends Error {
     }
   }
 }
-
-export class ErrorWithCause extends CustomError {
```
vojtechszocs, line 26: Removing since there's no usage so far, we can add it later if necessary.

vojtechszocs:
Note to self, adding `"../common/src"` to TS config `include` should fix the build error for both lib packages.

packages/common/rollup-configs.js
```diff
@@ -1,11 +1,13 @@
 import { execSync } from 'child_process';
-/* eslint-disable import/no-extraneous-dependencies */
+/* eslint-disable import/no-extraneous-dependencies -- avoid pulling Rollup dependencies into common package.json */
```
vojtechszocs, line 2: This addresses @andrewballantyne's review comment https://github.com/openshift/dynamic-plugin-sdk/pull/23#discussion_r774043939

packages/common/rollup-configs.js
```diff
@@ -1,11 +1,13 @@
...
 import nodeResolve from '@rollup/plugin-node-resolve';
 import typescript from '@rollup/plugin-typescript';
 import analyzer from 'rollup-plugin-analyzer';
 import dts from 'rollup-plugin-dts';
 /* eslint-enable import/no-extraneous-dependencies */
 
+const rootDir = process.env.PROJECT_CWD;
```
vojtechszocs, line 9: This env. variable is automatically set by Yarn: https://yarnpkg.com/advanced/lifecycle-scripts/#environment-variables

packages/common/rollup-configs.js
```diff
@@ -14,7 +16,7 @@ const getExternalModules = (pkg) => [
 const getExternalModuleTester = (pkg) => {
   const externalModules = getExternalModules(pkg);
   const externalModuleTests = externalModules.map((module) => new RegExp(`^${module}(\\/.+)*$`));
-  return (source) => externalModuleTests.some((regexp) => regexp.test(source));
+  return (moduleID) => externalModuleTests.some((regexp) => regexp.test(moduleID));
```
vojtechszocs, line 19: In Rollup's terminology, this is a "module ID" so I've renamed the argument to stay consistent with that.

FYI - webpack calls them "module requests" while ECMAscript spec calls them "module specifiers".

Wish that everyone just used the same naming :face_exhaling:

packages/common/rollup-configs.js
```diff
@@ -29,20 +31,21 @@ const getBanner = (pkg) => `/*
 /**
  * @returns {import('rollup').RollupOptions}
  */
-export const tsLibConfig = (pkg) => ({
-  input: 'src/index.ts',
+export const tsLibConfig = (pkg, inputFile) => ({
+  input: inputFile,
   output: {
     file: 'dist/index.js',
-    format: 'esm',
```
vojtechszocs, line 36: I've realized that this is already the default Rollup output file format: https://rollupjs.org/guide/en/#outputformat

packages/common/rollup-configs.js
```diff
@@ -29,20 +31,21 @@ const getBanner = (pkg) => `/*
...
     banner: getBanner(pkg),
   },
   plugins: [
     nodeResolve(),
     typescript({
       tsconfig: './tsconfig.json',
```
vojtechszocs, line 43: Fun fact: `tsconfig: 'tsconfig.json'` won't work, need `./` relative path prefix for some reason.

packages/common/rollup-configs.js
```diff
@@ -29,20 +31,21 @@ const getBanner = (pkg) => `/*
...
+      include: ['src/**/*', '../common/src/**/*'],
```
vojtechszocs, line 44: This is equivalent to `tsconfig.json` option
```json
"include": ["src", "../common/src"]
```

but without the above setting, `@rollup/plugin-typescript` will fail to process the code in `common` package as TypeScript code.

The reason for that seems to be

> Source files in these fields are ignored as Rollup's configuration is used instead.

https://github.com/rollup/plugins/tree/master/packages/typescript/#files-include-exclude

packages/common/rollup-configs.js
```diff
@@ -29,20 +31,21 @@ const getBanner = (pkg) => `/*
...
     }),
     analyzer({
       summaryOnly: true,
+      root: rootDir,
```
vojtechszocs, line 48: This makes the `rollup-plugin-analyzer` output module paths relative to monorepo root directory, since e.g. `packages/lib-utils` may include code from `packages/common`.

packages/eslint-plugin-internal/src/rules/all-bases.js
```diff
@@ -23,8 +23,7 @@ module.exports = {
   'import/no-extraneous-dependencies': [
     'error',
     {
-      devDependencies: ['**/packages/common/**', '**/packages/eslint-plugin-internal/**'],
-      optionalDependencies: false,
+      devDependencies: true,
```
vojtechszocs, line 26: I've discussed this with @andrewballantyne - this setting allows using dev dependencies in all monorepo packages.

packages/lib-runtime/src/store/PluginLoader.ts
```diff
@@ -202,8 +202,7 @@ export class PluginLoader {
    * Register the global function used by plugin entry scripts.
    */
   registerPluginEntryCallback(getWindow: () => typeof window = _.constant(window)) {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const windowGlobal = getWindow() as any;
+    const windowGlobal = getWindow() as unknown as AnyObject;
```
vojtechszocs, line 205: This change was done to reduce our existing `@typescript-eslint/no-explicit-any` rule suppressions.

packages/lib-runtime/tsconfig.json
```diff
@@ -2,8 +2,7 @@
   "extends": "@monorepo/common/tsconfig-bases/react-esm.json",
   "compilerOptions": {
     "outDir": "dist",
-    "declarationDir": "types",
-    "emitDeclarationOnly": true
```
vojtechszocs, line 6: `@rollup/plugin-typescript` ignores the `emitDeclarationOnly` option: https://github.com/rollup/plugins/tree/master/packages/typescript/#ignored-options

packages/lib-utils/src/config.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Set the {@link UtilsConfig} reference.
+ *
+ * This must be done before using any of the Kubernetes utilities.
+ */
+export const setUtilsConfig = (c: UtilsConfig) => {
+  if (config !== undefined) {
+    throw new Error('UtilsConfig reference has already been set');
+  }
+
+  config = Object.freeze({ ...c });
```
vojtechszocs, line 27: This ensures that own properties of `UtilsConfig` object aren't modified.

The `getUtilsConfig` function returns the frozen version.

packages/lib-utils/src/utils/common-fetch.ts
```diff
@@ -0,0 +1,90 @@
...
+import { getUtilsConfig } from '../config';
+import type { CommonFetch, CommonFetchText, CommonFetchJSON } from '../types/fetch';
+
+class TimeoutError extends CustomError {
+  public constructor(url: string, ms: number) {
+    super(`Call to ${url} timed out after ${ms}ms`);
+  }
+}
+
+const defaultTimeout = 60_000; // one minute
```
andrewballantyne, line 11:
```suggestion
const defaultTimeout = 60_000;
```
60,000 is well known to be 60 seconds === 1 minute. I don't think we need these kind of comments imo.

andrewballantyne (changes requested):
A few comments but otherwise looking awesome! The picture is starting to come together.

/cc @invincibleJai 
Jai, make sure you're aware of the changes in `packages/lib-utils/src/config.ts` and how that will impact things you're doing (won't directly impact, but best not to deviate too much from each other so it's easier to move over)

packages/lib-utils/src/utils/common-fetch.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const commonFetch: CommonFetch = async (url, options = {}, timeout = defaultTimeout) => {
```
invincibleJai, line 13: do we need this to be an async function?

invincibleJai:
> A few comments but otherwise looking awesome! The picture is starting to come together.
> 
> /cc @invincibleJai Jai, make sure you're aware of the changes in `packages/lib-utils/src/config.ts` and how that will impact things you're doing (won't directly impact, but best not to deviate too much from each other so it's easier to move over)

Thanks @andrewballantyne, as suggested my changes are in sync with changes made in this https://github.com/openshift/console/pull/10707. @vojtechszocs has done a review as well and have worked on it.

packages/common/src/types/fetch.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Fetch a resource over HTTP and return the {@link Response} object.
+ */
+export type ResourceFetch = (url: string, options?: RequestInit) => Promise<Response>;
```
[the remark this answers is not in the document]
vojtechszocs: This is our basic fetch function interface and I think it's generally better to not alias it to `fetch` API signature (even if they're the same). Some degree of duplicity is OK, in this case we're being explicit about our interface.

packages/lib-runtime/src/store/PluginLoader.ts
```diff
@@ -96,11 +95,12 @@ export class PluginLoader {
 
     consoleLogger.info(`Loading plugin manifest from ${manifestURL}`);
 
-    const manifestText = await this.options.fetchImpl(manifestURL, 'GET');
+    const response = await this.options.fetchImpl(manifestURL);
```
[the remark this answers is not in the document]
vojtechszocs: Previously, the `PluginLoader` worked with fetch interface (note that `method` arg is mandatory)
```ts
export type FetchResource = (url: string, method: string) => Promise<string>;
```

but we're changing it now to
```ts
export type ResourceFetch = (url: string, options?: RequestInit) => Promise<Response>;
```

Indeed, `GET` should be the default HTTP method when not specified. So basically what prompted the original explicit definition was that `method: string` wasn't marked as optional :smiley:

packages/lib-runtime/src/store/PluginLoader.ts
```diff
@@ -96,11 +95,12 @@ export class PluginLoader {
...
+    const responseText = await response.text();
 
     let manifest: PluginManifest = await pluginManifestSchema
       .strict()
-      .validate(JSON.parse(manifestText));
+      .validate(JSON.parse(responseText));
```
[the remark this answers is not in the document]
vojtechszocs: Inside `async` functions, runtime errors cause the resulting Promise to be rejected with the given error.

So this shouldn't blow up the main processing thread due to uncaught error.

If a function's contract includes throwing errors, the caller should adapt its code to handle them (in React render flows, this typically means using [error boundaries](https://reactjs.org/docs/error-boundaries.html)). For Promises, the caller should handle the Promise rejections.

packages/lib-runtime/src/utils/basic-fetch.ts
```diff
@@ -0,0 +1,21 @@
+import { CustomError, applyDefaults } from '@monorepo/common';
+import type { ResourceFetch } from '@monorepo/common';
+
+class FetchError extends CustomError {
```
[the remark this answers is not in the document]
vojtechszocs: This isn't exported because it's a custom error object the `basicFetch` implementation works with.

The `basicFetch` function is a fallback used when `PluginLoaderOptions.fetchImpl` is not specified (all properties within the `PluginLoaderOptions` object are optional by design).

Host apps normally want to use a custom fetch impl. and that impl. can throw whatever errors (declared in host app code) necessary.

TL;DR this is tied to `basicFetch` and not a general purpose HTTP fetch error object.

packages/common/src/errors/CustomError.ts
vojtechszocs: Right, PascalCase files typically contain a single React component or a class declaration, and camelCase files typically contain a single React hook or a function.

I know this is a matter of preference, but for better consistency and easier handling of imports, I think that named exports are better overall.

Let's say we change `packages/common/src/errors/CustomError.ts`
```diff
-export class CustomError extends Error {
+export default class CustomError extends Error {
```

we must then change `packages/common/src/index.ts` which re-exports the `CustomError` class
```diff
-export * from './errors/CustomError';
+export { default as CustomError } from './errors/CustomError';
```

to allow consumers importing `CustomError` from `common` package (index) as a named import.

If we wanted to use `CustomError` in the same package, let's say in `packages/common/src/utils/objects.ts`

```diff
-import { CustomError } from '../errors/CustomError';
+import CustomError from '../errors/CustomError';
```

I think it's easier to handle all exports as named than to handle default ones as a special case. But that's just me :smiley:

vojtechszocs: Also, if we wanted to add more exports to e.g. `packages/common/src/errors/CustomError.ts` those would have to be named exports, and the import syntax would look like
```ts
import CustomError, { foo } from '../errors/CustomError';
```

This may cause some people to put `foo` in a separate file, even if it's closely related to `CustomError`. Conventions shouldn't cause loss of flexibility if possible.

packages/lib-runtime/src/store/PluginLoader.ts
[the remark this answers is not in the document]
andrewballantyne: Ah good call, didn't see it was an async func

packages/common/src/errors/CustomError.ts
andrewballantyne: As I mentioned last week, it's a preference that I think has merit, but I won't "force it"... so named exports is fine.

andrewballantyne (approved):

PR #25, opened 2022-01-18
Add logic to process code references

vojtechszocs:
Fixes [HAC-444](https://issues.redhat.com/browse/HAC-444)

- add `EncodedCodeRef` and `CodeRef` types
- add `ResolvedExtension` type, to be used with upcoming `useResolvedExtensions` hook
- modify `Extension` type to allow subtyping its `type` property, for example:
```ts
export type Foo = Extension<
  'core.foo',
  {
    bar: string;
    qux: CodeRef<VoidFunction>;
  }
>;
```
- `decodeCodeRefs` function implements `EncodedCodeRef` :arrow_right: `CodeRef` value transformation (sync)
- `resolveCodeRefValues` function implements `CodeRef<T>` :arrow_right: `T` value transformation (async)
- `PluginStore` now uses `Map<string, CodeRef>` to preserve referential identity of `CodeRef` functions
- `CodeRef` cache key is composed of `pluginName` (e.g. `kubevirt`) and the encoded `$codeRef` value (e.g. `icons.vmIconElement`) so basically a composite key `<pluginName,(exposed)moduleName,exportName>`

### DIFF
--- a/packages/common/src/utils/objects.ts
+++ b/packages/common/src/utils/objects.ts
@@ -18,19 +18,13 @@ export const applyOverrides = <TObject>(obj: TObject, overrides: TObject): TObje
   _.defaultsDeep({}, overrides, obj);
 
 /**
- * Return `true` if `obj` is a plain non-React object.
- */
-export const isPlainNonReactObject = (obj: unknown): obj is AnyObject =>
-  _.isPlainObject(obj) && !(obj as AnyObject).$$typeof;
-
-/**
  * Recursive equivalent of {@link _.forOwn} function that traverses plain objects and arrays.
  */
 export const visitDeep = <TValue>(
   obj: AnyObject,
   predicate: (value: unknown) => value is TValue,
   valueCallback: (value: TValue, key: string, container: AnyObject) => void,
-  isPlainObject = isPlainNonReactObject,
+  isPlainObject: (obj: unknown) => obj is AnyObject = (o): o is AnyObject => _.isPlainObject(o),
 ) => {
   const visitValue = (value: unknown, key: string, container: AnyObject) => {
     if (predicate(value)) {
--- a/packages/common/src/utils/objects.ts
+++ b/packages/common/src/utils/objects.ts
@@ -1,4 +1,5 @@
 import * as _ from 'lodash-es';
+import type { AnyObject } from '../types/common';
 
 /**
  * Create new object by recursively assigning property defaults to `obj`.
@@ -15,3 +16,33 @@ export const applyDefaults = <TObject>(obj: TObject, defaults: TObject): TObject
  */
 export const applyOverrides = <TObject>(obj: TObject, overrides: TObject): TObject =>
   _.defaultsDeep({}, overrides, obj);
+
+/**
+ * Return `true` if `obj` is a plain non-React object.
+ */
+export const isPlainNonReactObject = (obj: unknown): obj is AnyObject =>
+  _.isPlainObject(obj) && !(obj as AnyObject).$$typeof;
+
+/**
+ * Recursive equivalent of {@link _.forOwn} function that traverses plain objects and arrays.
+ */
+export const visitDeep = <TValue>(
+  obj: AnyObject,
+  predicate: (value: unknown) => value is TValue,
+  valueCallback: (value: TValue, key: string, container: AnyObject) => void,
+  isPlainObject = isPlainNonReactObject,
+) => {
+  const visitValue = (value: unknown, key: string, container: AnyObject) => {
+    if (predicate(value)) {
+      valueCallback(value, key, container);
+    } else if (isPlainObject(value)) {
+      visitDeep(value, predicate, valueCallback, isPlainObject);
+    } else if (Array.isArray(value)) {
+      value.forEach((element) => {
+        visitDeep(element, predicate, valueCallback, isPlainObject);
+      });
+    }
+  };
+
+  _.forOwn(obj, visitValue);
+};
--- a/packages/lib-runtime/src/schema/plugin-manifest.ts
+++ b/packages/lib-runtime/src/schema/plugin-manifest.ts
@@ -44,22 +44,21 @@ const extensionType = yup
 /**
  * Schema for `Extension` objects.
  */
-const extension = yup
-  .object({
-    type: extensionType,
-    properties: yup.object().required(),
-  })
-  .required();
+const extension = yup.object().required().shape({
+  type: extensionType,
+  properties: yup.object().required(),
+});
 
 /**
  * Schema for `PluginManifest` objects.
  */
 export const pluginManifest = yup
-  .object({
+  .object()
+  .required()
+  .shape({
     name: pluginName,
     version: semverString,
     extensions: yup.array().of(extension).required(),
-  })
-  .required();
+  });
 
 export { pluginManifest as pluginManifestSchema };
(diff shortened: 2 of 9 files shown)


packages/lib-runtime/src/store/coderefs.ts
```diff
@@ -0,0 +1,150 @@
...
+ */
+export const decodeExtensionCodeRefs = (
+  extension: LoadedExtension,
+  entryModule: PluginEntryModule,
+) => {
+  visitDeep<EncodedCodeRef>(extension.properties, isEncodedCodeRef, (encodedRef, key, obj) => {
+    const codeRef = createCodeRef(
+      encodedRef,
+      entryModule,
+      (message) => `${message} in extension ${extension.uid}`,
```
vojtechszocs, line 127: This adds useful diagnostics to `CodeRef` value resolution errors, so that developers know exactly from where they are coming from.

packages/lib-runtime/src/store/coderefs.ts
```diff
@@ -0,0 +1,150 @@
...
+  entryModule: PluginEntryModule,
+  formatErrorMessage: (message: string) => string = _.identity,
+): CodeRef => {
+  let referencedObject: unknown;
+  let referencedObjectResolved = false;
+
+  const codeRef: CodeRef = async () => {
+    if (referencedObjectResolved) {
+      return referencedObject;
+    }
```
christianvogt, line 77: If this function is called twice before the first one completes, it will be processed twice?
It doesn't seem like there's a downside but the comment suggests that resolution is done only once.

packages/lib-runtime/src/store/coderefs.ts
```diff
@@ -0,0 +1,150 @@
...
+  visitDeep<EncodedCodeRef>(extension.properties, isEncodedCodeRef, (encodedRef, key, obj) => {
+    const codeRef = createCodeRef(
+      encodedRef,
+      entryModule,
+      (message) => `${message} in extension ${extension.uid}`,
+    );
+
+    // Use a sensible function name that reflects the contextual information
+    Object.defineProperty(codeRef, 'name', {
+      value: `$codeRef_${extension.uid}_${encodedRef}`,
```
vojtechszocs, line 132: BTW, this is actually needed in Console - see `sameLoader` in [`async.tsx`](https://github.com/openshift/console/blob/master/frontend/public/components/utils/async.tsx) utils module.

Comparing functions is always tricky, but this should make it a bit more easier, at least for `CodeRef` ones.

cc @rawagner

packages/lib-runtime/src/utils/objects.ts
```diff
@@ -17,7 +17,9 @@ export const createMethodDelegate = <
 ) => {
   const delegate = {} as Pick<TObject, TObjectMethods>;
 
-  const validMethodNames = methodNames.filter((methodName) => _.isFunction(target[methodName]));
+  const validMethodNames = methodNames.filter(
+    (methodName) => typeof target[methodName] === 'function',
```
vojtechszocs, line 21: Using `typeof obj === 'function'` seems generally better than `_.isFunction` since the former is natively supported by the browser.

Please correct me if I'm wrong here :smiley:

packages/lib-runtime/src/store/coderefs.ts
vojtechszocs: The only async bit of this function is
```ts
const moduleFactory = await entryModule.get(moduleName);
```

so if this function gets called twice in a row, it is technically possible for the 2nd execution to get ahead of the 1st execution. In any case, `moduleFactory` would be computed twice, yes. I don't think there's a solution for this, given the async nature of `moduleFactory` computation.

vojtechszocs: > If this function is called twice before the first one completes, it will be processed twice?

Basically yes, due to `moduleFactory` async computation being part of the resolution logic.

vojtechszocs: > It doesn't seem like there's a downside but the comment suggests that resolution is done only once.

Good point, will update the comment.

christianvogt: `codeRef` function could create a shared `Promise` instance and the first call is responsible for resolving that `Promise`.

packages/lib-runtime/src/utils/objects.ts
andrewballantyne: I honesty think Lodash is overrated for a lot of the basic functions of JS. Deep copies is the only thing I think of from a use-case of Lodash. Most of their functions are minutely better or not at all. I benched a few when I first got here because my previous team used lodash for absolutely everything and it was just making it harder to read.

So yes, native 👍

packages/lib-runtime/src/store/coderefs.ts
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * Parse data from the {@link EncodedCodeRef} object.
+ *
+ * Returns `undefined` if the `$codeRef` value is malformed.
+ */
+const parseEncodedCodeRef = (
+  ref: EncodedCodeRef,
+): [moduleName: string, exportName: string] | undefined => {
+  const match = ref.$codeRef.match(/^([^.]+)(?:\.([^.]+)){0,1}$/);
```
andrewballantyne, line 40: FWIW, I think this also works:
```suggestion
  const match = ref.$codeRef.match(/^([^.]+)(?:\.([^.]+))?$/);
```

andrewballantyne (changes requested):

packages/lib-runtime/src/store/coderefs.ts
vojtechszocs: Yup, that's exactly what I was thinking of yesterday as well, will do the change.

packages/common/src/utils/objects.ts
```diff
@@ -17,20 +17,14 @@ export const applyDefaults = <TObject>(obj: TObject, defaults: TObject): TObject
...
-
 /**
  * Recursive equivalent of {@link _.forOwn} function that traverses plain objects and arrays.
  */
 export const visitDeep = <TValue>(
   obj: AnyObject,
   predicate: (value: unknown) => value is TValue,
   valueCallback: (value: TValue, key: string, container: AnyObject) => void,
-  isPlainObject = isPlainNonReactObject,
+  isPlainObject: (obj: unknown) => obj is AnyObject = (o): o is AnyObject => _.isPlainObject(o),
```
vojtechszocs, line 27: FYI - can't do
```ts
isPlainObject: (obj: unknown) => obj is AnyObject = _.isPlainObject,
```
since `_.isPlainObject` return type is just `boolean` (not a type guard).

packages/lib-runtime/src/schema/plugin-manifest.ts
```diff
@@ -45,21 +45,24 @@ const extensionType = yup
...
 const extension = yup
-  .object({
+  .object()
+  .required()
+  .shape({
     type: extensionType,
     properties: yup.object().required(),
   })
-  .required();
+  .noUnknown(false);
```
vojtechszocs, line 54: This reflects `[customProperty: string]: unknown;` properties in `Extension` type, allowing extension declarations to contain custom (direct) properties.

andrewballantyne: `yup` doesn't lock down extra properties. It validates the structure to match and ignores unknown properties by default. We don't need `noUnknown` as `required` covers the effort of additional properties while enforcing the object is there (you don't need the earlier required in this case)

andrewballantyne: Quick clarity of `yup` usage:

`.object()` is needed to do `.shape`
`.required()` is desired any point we want this just for clarity; but honestly the fact that props internal are required, it makes the object required
`.shape({ ... })` is needed from a "we want structure" point of view and thus also makes it required if internal properties are required

andrewballantyne (approved):

PR #28, opened 2022-02-09
HAC-423: Bring AppInitSDK into Core SDK

florkbr:
Migrates the AppInitSDK logic into the HAC dynamic plugin SDK. Does not migrate the api discovery logic nor actions/reducers required by the AppInitSDK logic.

See https://issues.redhat.com/browse/HAC-423

### DIFF
--- a/packages/lib-utils/src/app/AppInitSDK.tsx
+++ b/packages/lib-utils/src/app/AppInitSDK.tsx
@@ -7,12 +7,14 @@ import type { InitAPIDiscovery } from '../types/api-discovery';
 import { initAPIDiscovery } from './api-discovery';
 import { useReduxStore } from './redux';
 
-type AppInitSDKProps = React.PropsWithChildren<{
+type AppInitSDKProps = {
+  /** Only child is your Application */
+  children: React.ReactElement;
   configurations: {
     apiDiscovery?: InitAPIDiscovery;
     appFetch: UtilsConfig['appFetch'];
   };
-}>;
+};
 
 /**
  * Component for providing store access to the SDK.
@@ -25,8 +27,7 @@ type AppInitSDKProps = React.PropsWithChildren<{
  * return (
  *  <Provider store={store}>
  *   <AppInitSDK configurations={{ appFetch, apiDiscovery }}>
- *      <CustomApp />
- *      ...
+ *      <App />
  *   </AppInitSDK>
  *  </Provider>
  * )
@@ -34,18 +35,17 @@ type AppInitSDKProps = React.PropsWithChildren<{
  */
 const AppInitSDK: React.FC<AppInitSDKProps> = ({ children, configurations }) => {
   const { store, storeContextPresent } = useReduxStore();
+  const { appFetch, apiDiscovery = initAPIDiscovery } = configurations;
   React.useEffect(() => {
-    const { appFetch, apiDiscovery = initAPIDiscovery } = configurations;
     try {
       setUtilsConfig({ appFetch });
       apiDiscovery(store);
     } catch (e) {
       consoleLogger.warn(e);
     }
-  }, [configurations, store]);
+  }, [apiDiscovery, appFetch, store]);
 
-  // eslint-disable-next-line react/jsx-no-useless-fragment
-  return !storeContextPresent ? <Provider store={store}>{children}</Provider> : <>{children}</>;
+  return !storeContextPresent ? <Provider store={store}>{children}</Provider> : children;
 };
 
 export default AppInitSDK;
--- a/packages/lib-utils/src/app/AppInitSDK.tsx
+++ b/packages/lib-utils/src/app/AppInitSDK.tsx
@@ -1,0 +1,51 @@
+import { consoleLogger } from '@monorepo/common';
+import * as React from 'react';
+import { Provider } from 'react-redux';
+import type { UtilsConfig } from '../config';
+import { setUtilsConfig } from '../config';
+import type { InitAPIDiscovery } from '../types/api-discovery';
+import { initAPIDiscovery } from './api-discovery';
+import { useReduxStore } from './redux';
+
+type AppInitSDKProps = React.PropsWithChildren<{
+  configurations: {
+    apiDiscovery?: InitAPIDiscovery;
+    appFetch: UtilsConfig['appFetch'];
+  };
+}>;
+
+/**
+ * Component for providing store access to the SDK.
+ * Add this at app-level to make use of app's redux store and pass configurations prop needed to initialize the app, preferred to have it under Provider.
+ * It checks for store instance if present or not.
+ * If the store is there then the reference is persisted to be used in SDK else it creates a new store and passes it to the children with the provider
+ * @component AppInitSDK
+ * @example
+ * ```ts
+ * return (
+ *  <Provider store={store}>
+ *   <AppInitSDK configurations={{ appFetch, apiDiscovery }}>
+ *      <CustomApp />
+ *      ...
+ *   </AppInitSDK>
+ *  </Provider>
+ * )
+ * ```
+ */
+const AppInitSDK: React.FC<AppInitSDKProps> = ({ children, configurations }) => {
+  const { store, storeContextPresent } = useReduxStore();
+  React.useEffect(() => {
+    const { appFetch, apiDiscovery = initAPIDiscovery } = configurations;
+    try {
+      setUtilsConfig({ appFetch });
+      apiDiscovery(store);
+    } catch (e) {
+      consoleLogger.warn(e);
+    }
+  }, [configurations, store]);
+
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  return !storeContextPresent ? <Provider store={store}>{children}</Provider> : <>{children}</>;
+};
+
+export default AppInitSDK;
--- a/packages/lib-utils/src/app/AppInitSDK.tsx
+++ b/packages/lib-utils/src/app/AppInitSDK.tsx
@@ -17,13 +17,13 @@ type AppInitSDKProps = {
 };
 
 /**
- * Component for providing store access to the SDK.
+ * Initializes the host application to work with Kubernetes and related SDK utilities.
  * Add this at app-level to make use of app's redux store and pass configurations prop needed to initialize the app, preferred to have it under Provider.
  * It checks for store instance if present or not.
  * If the store is there then the reference is persisted to be used in SDK else it creates a new store and passes it to the children with the provider
  * @component AppInitSDK
  * @example
- * ```ts
+ * ```tsx
  * return (
  *  <Provider store={store}>
  *   <AppInitSDK configurations={{ appFetch, apiDiscovery }}>
@@ -41,7 +41,7 @@ const AppInitSDK: React.FC<AppInitSDKProps> = ({ children, configurations }) => 
       setUtilsConfig({ appFetch });
       apiDiscovery(store);
     } catch (e) {
-      consoleLogger.warn(e);
+      consoleLogger.warn('Error while initializing AppInitSDK', e);
     }
   }, [apiDiscovery, appFetch, store]);
 
--- a/packages/lib-utils/src/app/redux/index.ts
+++ b/packages/lib-utils/src/app/redux/index.ts
@@ -1,0 +1,43 @@
+import { consoleLogger } from '@monorepo/common';
+import * as React from 'react';
+import { useStore } from 'react-redux';
+import type { Store } from 'redux';
+import { applyMiddleware, combineReducers, createStore, compose } from 'redux';
+import thunk from 'redux-thunk';
+import { setReduxStore, getReduxStore } from '../../config';
+import type { SDKStoreState } from '../../types/redux';
+import { SDKReducers } from './reducers';
+
+/**
+ * `useReduxStore` will provide the store instance if present or else create one along with info if the context was present.
+ *
+ * @example
+ * ```ts
+ * function Component () {
+ *   const {store, storeContextPresent} = useReduxStore()
+ *   return ...
+ * }
+ * ```
+ */
+export const useReduxStore = (): { store: Store<unknown>; storeContextPresent: boolean } => {
+  const storeContext = useStore();
+  const [storeContextPresent, setStoreContextPresent] = React.useState(false);
+  const store = React.useMemo(() => {
+    if (storeContext) {
+      setStoreContextPresent(true);
+      setReduxStore(storeContext);
+    } else {
+      consoleLogger.info('Creating the SDK redux store');
+      setStoreContextPresent(false);
+      const storeInstance = createStore(
+        combineReducers<SDKStoreState>(SDKReducers),
+        {},
+        compose(applyMiddleware(thunk)),
+      );
+      setReduxStore(storeInstance);
+    }
+    return getReduxStore();
+  }, [storeContext]);
+
+  return { store, storeContextPresent };
+};
--- a/packages/lib-utils/src/app/redux/index.ts
+++ b/packages/lib-utils/src/app/redux/index.ts
@@ -8,6 +8,11 @@ import { setReduxStore, getReduxStore } from '../../config';
 import type { SDKStoreState } from '../../types/redux';
 import { SDKReducers } from './reducers';
 
+type UseReduxStoreResult = {
+  store: Store<unknown>;
+  storeContextPresent: boolean;
+};
+
 /**
  * `useReduxStore` will provide the store instance if present or else create one along with info if the context was present.
  *
@@ -19,7 +24,7 @@ import { SDKReducers } from './reducers';
  * }
  * ```
  */
-export const useReduxStore = (): { store: Store<unknown>; storeContextPresent: boolean } => {
+export const useReduxStore = (): UseReduxStoreResult => {
   const storeContext = useStore();
   const [storeContextPresent, setStoreContextPresent] = React.useState(false);
   const store = React.useMemo(() => {
--- a/packages/lib-utils/src/config.ts
+++ b/packages/lib-utils/src/config.ts
@@ -1,4 +1,5 @@
 import type { ResourceFetch } from '@monorepo/common';
+import type { Store } from 'redux';
 
 export type UtilsConfig = {
   /**
@@ -38,4 +39,28 @@ export const getUtilsConfig = () => {
   }
 
   return config;
+};
+
+let reduxStore: Store;
+
+/**
+ * Set the {@link Store} reference.
+ *
+ * This must be done before using the AppInitSDK React entrypoint
+ */
+export const setReduxStore = (storeData: Store) => {
+  reduxStore = storeData;
+};
+
+/**
+ * Get the {@link Store} reference.
+ *
+ * Throws an error if the reference isn't already set.
+ */
+export const getReduxStore = () => {
+  if (reduxStore === undefined) {
+    throw new Error('Redux store reference has not been set');
+  }
+
+  return reduxStore;
 };
--- a/packages/lib-utils/src/types/api-discovery.ts
+++ b/packages/lib-utils/src/types/api-discovery.ts
@@ -1,0 +1,19 @@
+import type { Store, AnyAction } from 'redux';
+import type { ActionType as Action } from 'typesafe-actions';
+import type { K8sVerb } from './k8s';
+
+export type InitAPIDiscovery = (store: Store<unknown, Action<AnyAction>>) => void;
+
+export type APIResourceList = {
+  kind: 'APIResourceList';
+  apiVersion: 'v1';
+  groupVersion: string;
+  resources?: {
+    name: string;
+    singularName?: string;
+    namespaced?: boolean;
+    kind: string;
+    verbs: K8sVerb[];
+    shortNames?: string[];
+  }[];
+};
--- a/packages/lib-utils/src/types/api-discovery.ts
+++ b/packages/lib-utils/src/types/api-discovery.ts
@@ -1,13 +1,13 @@
+import type { K8sVerb } from '@openshift/dynamic-plugin-sdk/src/types/core';
 import type { Store, AnyAction } from 'redux';
 import type { ActionType as Action } from 'typesafe-actions';
-import type { K8sVerb } from './k8s';
+import type { K8sModelCommon } from './k8s';
 
 export type InitAPIDiscovery = (store: Store<unknown, Action<AnyAction>>) => void;
 
-export type APIResourceList = {
+export type APIResourceList = K8sModelCommon & {
   kind: 'APIResourceList';
   apiVersion: 'v1';
-  groupVersion: string;
   resources?: {
     name: string;
     singularName?: string;
--- a/packages/lib-utils/src/types/k8s.ts
+++ b/packages/lib-utils/src/types/k8s.ts
@@ -58,13 +58,3 @@ export type Patch = {
   path: string;
   value?: unknown;
 };
-
-export type K8sVerb =
-  | 'create'
-  | 'get'
-  | 'list'
-  | 'update'
-  | 'patch'
-  | 'delete'
-  | 'deletecollection'
-  | 'watch';
--- a/packages/lib-utils/src/types/k8s.ts
+++ b/packages/lib-utils/src/types/k8s.ts
@@ -58,3 +58,13 @@ export type Patch = {
   path: string;
   value?: unknown;
 };
+
+export type K8sVerb =
+  | 'create'
+  | 'get'
+  | 'list'
+  | 'update'
+  | 'patch'
+  | 'delete'
+  | 'deletecollection'
+  | 'watch';
--- a/packages/lib-utils/package.json
+++ b/packages/lib-utils/package.json
@@ -22,15 +22,20 @@
     "test": "echo 'Write some tests!'"
   },
   "peerDependencies": {
+    "@openshift/dynamic-plugin-sdk": "^1.0.0",
     "lodash-es": "^4.17.21",
     "react": "^17.0.2",
     "react-redux": "^7.2.2",
-    "redux": "^4.0.3",
+    "redux": "^4.1.2",
     "redux-thunk": "^2.4.1"
+  },
+  "dependencies": {
+    "immutable": "^3.8.2",
+    "typesafe-actions": "^4.4.2"
   },
   "devDependencies": {
     "@monorepo/common": "0.0.0-fixed",
-    "@openshift/dynamic-plugin-sdk": "1.0.0-alpha1",
+    "@openshift/dynamic-plugin-sdk": "link:../lib-core",
     "@rollup/plugin-node-resolve": "^13.1.1",
     "@rollup/plugin-typescript": "^8.3.0",
     "@types/lodash-es": "^4.17.5",
@@ -38,16 +43,12 @@
     "lodash-es": "^4.17.21",
     "react": "^17.0.2",
     "react-redux": "^7.2.2",
-    "redux": "^4.0.3",
+    "redux": "^4.1.2",
     "redux-thunk": "^2.4.1",
     "rollup": "^2.61.1",
     "rollup-plugin-analyzer": "^4.0.0",
     "rollup-plugin-dts": "^4.0.1",
     "tslib": "^2.3.1",
     "typescript": "~4.4.4"
-  },
-  "dependencies": {
-    "immutable": "^3.8.2",
-    "typesafe-actions": "^4.4.2"
   }
 }
--- a/packages/lib-utils/package.json
+++ b/packages/lib-utils/package.json
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@monorepo/common": "0.0.0-fixed",
+    "@openshift/dynamic-plugin-sdk": "1.0.0-alpha1",
     "@rollup/plugin-node-resolve": "^13.1.1",
     "@rollup/plugin-typescript": "^8.3.0",
     "@types/lodash-es": "^4.17.5",
--- a/packages/lib-utils/package.json
+++ b/packages/lib-utils/package.json
@@ -23,7 +23,10 @@
   },
   "peerDependencies": {
     "lodash-es": "^4.17.21",
-    "react": "^17.0.2"
+    "react": "^17.0.2",
+    "react-redux": "^7.2.2",
+    "redux": "^4.0.3",
+    "redux-thunk": "^2.4.1"
   },
   "devDependencies": {
     "@monorepo/common": "0.0.0-fixed",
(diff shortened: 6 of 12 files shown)


florkbr:
@andrewballantyne it doesn't look like I have permissions to add reviewers or assignees to PRs 😕

packages/lib-core/package.json
```diff
@@ -24,20 +24,30 @@
...
     "@rollup/plugin-typescript": "^8.3.0",
+    "@types/immutable": "3.x",
     "@types/lodash-es": "^4.17.5",
     "@types/react": "^17.0.37",
+    "@types/react-redux": "6.0.2",
+    "immutable": "3.x",
     "lodash-es": "^4.17.21",
     "react": "^17.0.2",
+    "react-redux": "7.2.2",
+    "redux": "4.0.3",
```
florkbr, line 44: @vojtechszocs FYI I needed to bump the version of Redux to `4.0.3` from `4.0.1` (that the console uses).

This is due to a typescript bug fixed in Typescript `4.4.4` (the console is using Typescript `~3.8`).

We will need to keep this in mind when pulling the HAC sdk back into the console.

packages/lib-core/src/types/k8s.ts
```diff
@@ -0,0 +1,122 @@
+export type K8sVerb =
```
florkbr, line 1: I'm curious on reviewers thoughts here: do we want the type definitions next to the source files (IE `lib-core/src/app/k8s` versus `lib-core/src/types`)? I've also combined some of the definitions we had in `console-types` and `common-types` into this file. I can split them back out if there is disagreement.

packages/lib-core/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { initApiDiscovery } from './k8s/api-discovery';
+import { useReduxStore } from './useReduxStore';
+
+type AppInitSDKProps = {
+  children: React.ReactNode;
+  configurations: {
+    apiDiscovery?: InitApiDiscovery;
+    appFetch: UtilsConfig['appFetch'];
+    /** @deprecated - will be removed later when we have an interface for plugin */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
florkbr, line 16: We use any all over these files. I'll take a stab at making them more explicit.

packages/lib-core/package.json
```diff
@@ -24,20 +24,30 @@
   "peerDependencies": {
     "lodash-es": "^4.17.21",
     "react": "^17.0.2",
+    "react-redux": "7.2.2",
```
florkbr, line 27: TODO add carat range here

packages/lib-core/src/utils/k8s/k8s.ts
```diff
@@ -0,0 +1 @@
+export const k8sBasePath = `${window.SERVER_FLAGS.basePath}api/kubernetes`;
```
andrewballantyne, line 1: Oooh, we don't want this. But I bet this is needed for the apiDiscovery 😕 

We need to rely maybe on the appFetch function in this case so we can fetch to the same API backend that the k8s utils will call 🤔 

cc @vojtechszocs 

---

This can be deleted for now... but we'll need to think on this with the work @vojtechszocs has in https://github.com/openshift/dynamic-plugin-sdk/pull/30

packages/lib-core/src/types/window.ts
```diff
@@ -0,0 +1,11 @@
+import type { compose } from 'redux';
+
+declare global {
+  interface Window {
+    SERVER_FLAGS: {
+      basePath: string;
+    };
```
andrewballantyne, line 7: We won't have a server, so this is not going to be needed.

packages/lib-core/src/utils/k8s/k8s-get-resource.ts
```diff
@@ -0,0 +1,13 @@
...
+ * @deprecated - This is SDK internal function for resource-icon purposes, use Console version instead.
+ * Provides an abbreviation string for given kind with respect to abbrBlacklist.
+ * @param kind Kind for which the abbreviation is generated.
+ * @return Abbreviation string for given kind.
+ * TODO: Use in resource-icon component once it is being migrated to the SDK.
+ * * */
+export const kindToAbbr = (kind: string) => {
+  const abbrKind = (kind.replace(/[^A-Z]/g, '') || kind.toUpperCase()).slice(0, 4);
+  return abbrBlacklist.includes(abbrKind) ? abbrKind.slice(0, -1) : abbrKind;
+};
```
andrewballantyne, line 13: If it's not needed -- we should be adding it yet.

Delete this for now. We need to think on the `@deprecated` state too. We need this internally, and what is and isn't exposed out of the SDK is on us, so this deprecation is weird for the Core SDK.

packages/lib-core/src/types/k8s.ts
```diff
@@ -0,0 +1,122 @@
...
+  resourceVersion?: string;
+  uid?: string;
+};
+
+// Properties common to (almost) all Kubernetes resources.
+export type K8sResourceCommon = {
+  apiVersion?: string;
+  kind?: string;
+  metadata?: ObjectMetadata;
+};
```
andrewballantyne, line 122: This is being added by @vojtechszocs's PR - https://github.com/openshift/dynamic-plugin-sdk/pull/30/files#diff-01541dad1d9bd6352f582e703d8f5a8e5c2b81b6b92e7cab86fd07d542c5d8b6R12

packages/lib-core/src/types/k8s.ts
```diff
@@ -0,0 +1,122 @@
...
+  matchLabels?: MatchLabels;
+  matchExpressions?: MatchExpression[];
+};
+
+enum BadgeType {
+  DEV = 'Dev Preview',
+  TECH = 'Tech Preview',
+}
+
+export type K8sModel = {
```
andrewballantyne, line 44: This is going to be covered & expanded on eventually from Vojtech's PR: https://github.com/openshift/dynamic-plugin-sdk/pull/30/files#diff-01541dad1d9bd6352f582e703d8f5a8e5c2b81b6b92e7cab86fd07d542c5d8b6R7

And this has a lot of weird OpenShift Console use-cases we are not yet ready or ever want to support (ie the badge prop)

packages/lib-core/src/types/redux.ts
```diff
@@ -0,0 +1,27 @@
+import type { Map as ImmutableMap } from 'immutable';
+import type { K8sResourceCommon } from './k8s';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type K8sState = ImmutableMap<string, any>;
```
andrewballantyne, line 5: `any` types are bad... we shouldn't have this. Use `AnyObject` ([link](https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/common/src/types/common.ts#L8)) for now and put a TODO for us to expand it out when we start putting the reducer together.

andrewballantyne (changes requested):
Once again, sorry for the ticket issue... lib-core is not going to be the spot for a lot of this since it has a different goal. App Bootstrapping should be ideally under `lib-utils/.../app`

packages/lib-core/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,59 @@
...
+      setUtilsConfig({ appFetch });
+      if (initPlugins) {
+        initPlugins(store);
+      }
+      apiDiscovery(store);
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.warn(e);
+    }
+  }, [configurations, store]);
```
vojtechszocs, line 53: I think it's better to destructure `configurations` and have the effect depend on the destructured values directly, instead of depending on the `configurations` itself (which is technically just a holder for values).

```ts
const AppInitSDK: React.FC<AppInitSDKProps> = ({
  children,
  configurations: {
    appFetch,
    initPlugins,
    apiDiscovery = initApiDiscovery
  }
}) => {
  // .. React.useEffect can refer directly to destructured values
};
```

packages/lib-core/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,59 @@
...
+    const { appFetch, initPlugins, apiDiscovery = initApiDiscovery } = configurations;
+    try {
+      setUtilsConfig({ appFetch });
+      if (initPlugins) {
+        initPlugins(store);
+      }
+      apiDiscovery(store);
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.warn(e);
```
vojtechszocs, line 51: We should use our `consoleLogger` utility for consistent browser console log behavior.

In my PR #30 I'm moving it out of `lib-core` into `common` package. We can improve this later on.

packages/lib-core/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }, [configurations, store]);
+
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  return !storeContextPresent ? <Provider store={store}>{children}</Provider> : <>{children}</>;
```
vojtechszocs, line 56: Can't we just return `children` in the right-hand side of the ternary operator? (I think that's what `react/jsx-no-useless-fragment` complains about.)

packages/lib-core/src/app/core/reducers/index.ts
```diff
@@ -0,0 +1,11 @@
+import type { ActionType as Action } from 'typesafe-actions';
```
vojtechszocs, line 1: Can we put any Redux related files in `redux` directory, e.g. `redux/reducers/..` ?

packages/lib-core/src/app/storeHandler.ts
```diff
@@ -0,0 +1,12 @@
+import type { Store } from 'redux';
+
+let store: Store;
+
+const storeHandler = {
+  setStore: (storeData: Store) => {
+    store = storeData;
+  },
+  getStore: (): Store => store,
```
vojtechszocs, line 9: How is this conceptually different than `setUtilsConfig` ?

We could simply track reference to Redux store instance via ^^

packages/lib-core/src/types/k8s.ts
vojtechszocs: For now the general convention is to put types into `src/types` (within the given monorepo package) with a meaningful file name reflecting the purpose of these types.

vojtechszocs (changes requested):

packages/lib-core/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * Add this at app-level to make use of app's redux store and pass configurations prop needed to initialize the app, preferred to have it under Provider.
+ * It checks for store instance if present or not.
+ * If the store is there then the reference is persisted to be used in SDK else it creates a new store and passes it to the children with the provider
+ * @component AppInitSDK
+ * @example
+ * ```ts
+ * return (
+ *  <Provider store={store}>
+ *   <AppInitSDK configurations={{ appFetch, apiDiscovery }}>
+ *      <CustomApp />
```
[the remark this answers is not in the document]
vojtechszocs: My bad, `PluginStoreProvider` should be rendered via `AppInitSDK`.

florkbr:
@vojtechszocs @andrewballantyne I've implemented the requested changes. Let me know if I missed anything! Thank you.

packages/lib-core/src/types/window.ts
andrewballantyne: I actually like this idea... we are the middle ground package, we don't need to come with this support.

andrewballantyne: We may want to expose things later for testing scenarios, but for now I think we should avoid it.

packages/lib-core/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Provider } from 'react-redux';
+import type { Store } from 'redux';
+import type { InitApiDiscovery } from '../types/api-discovery';
+import type { UtilsConfig } from './configSetup';
+import { setUtilsConfig } from './configSetup';
+import { initApiDiscovery } from './k8s/api-discovery';
+import { useReduxStore } from './useReduxStore';
+
+type AppInitSDKProps = {
+  children: React.ReactNode;
```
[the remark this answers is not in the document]
andrewballantyne: Not in this case... child is needed. This needs to wrap your app.

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
+import { consoleLogger } from '@monorepo/common';
+import * as React from 'react';
+import { Provider } from 'react-redux';
+import type { UtilsConfig } from '../config';
+import { setUtilsConfig } from '../config';
+import type { InitAPIDiscovery } from '../types/api-discovery';
+import { initAPIDiscovery } from './api-discovery';
+import { useReduxStore } from './redux';
+
+type AppInitSDKProps = React.PropsWithChildren<{
```
andrewballantyne, line 10: I disagree with @vojtechszocs on this one. `children` is required -- but since `undefined` is part of `React.ReactNode` we'll need to use `ReactElement`. It's important to note this component is not intended to be used next to your app... it needs to be wrapping your app so we can provide all things important for hooks and usages inside your app.

Please change to:

```
type AppInitSDKProps = {
  children: React.ReactElement; // this is only a single JSX component (their app)
  configurations: {
    appFetch: UtilsConfig['appFetch'];
  };
};
```

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * ```ts
+ * return (
+ *  <Provider store={store}>
+ *   <AppInitSDK configurations={{ appFetch, apiDiscovery }}>
+ *      <CustomApp />
+ *      ...
+ *   </AppInitSDK>
+ *  </Provider>
+ * )
+ * ```
```
andrewballantyne, line 33: 🤔 
I wonder if we should add an example without Redux too since that's what the `useReduxStore` hook is trying to determine as well.

Maybe not needed for now. But we should delete the `...` as we only allow 1 child:

```suggestion
 * @example
 * ```ts
 * return (
 *  <Provider store={store}>
 *   <AppInitSDK configurations={{ appFetch, apiDiscovery }}>
 *      <App />
 *   </AppInitSDK>
 *  </Provider>
 * )
 * ```
```

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+ */
+const AppInitSDK: React.FC<AppInitSDKProps> = ({ children, configurations }) => {
+  const { store, storeContextPresent } = useReduxStore();
+  React.useEffect(() => {
+    const { appFetch, apiDiscovery = initAPIDiscovery } = configurations;
```
andrewballantyne, line 38: Destruct outside of `useEffect`. We want to key off the values of each, not the object itself.

`useEffect` uses referential equality and the object is not needed to be part of that check.

@vojtechszocs made [the request here](https://github.com/openshift/dynamic-plugin-sdk/pull/28#discussion_r809276108) too.

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      setUtilsConfig({ appFetch });
+      apiDiscovery(store);
+    } catch (e) {
+      consoleLogger.warn(e);
+    }
+  }, [configurations, store]);
+
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  return !storeContextPresent ? <Provider store={store}>{children}</Provider> : <>{children}</>;
```
andrewballantyne, line 48: Yeah, please remove the fragment. With the new typing Children will work fine.

packages/lib-utils/src/types/api-discovery.ts
```diff
@@ -0,0 +1,19 @@
+import type { Store, AnyAction } from 'redux';
+import type { ActionType as Action } from 'typesafe-actions';
+import type { K8sVerb } from './k8s';
+
+export type InitAPIDiscovery = (store: Store<unknown, Action<AnyAction>>) => void;
```
andrewballantyne, line 5: 🤔 
I wonder if we want to take a Generic approach to the store data... extending our data. Not sure it matters though as we'll probably move away from Redux to React Context or something

packages/lib-utils/src/types/api-discovery.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type APIResourceList = {
```
andrewballantyne, line 7:
```suggestion
export type APIResourceList = K8sModelCommon & {
```
I wonder if not basing this and dropping `groupVersion` is not better. That way we tie it back to the common model type.

packages/lib-utils/src/types/k8s.ts
```diff
@@ -58,3 +58,13 @@ export type Patch = {
...
+
+export type K8sVerb =
+  | 'create'
+  | 'get'
+  | 'list'
+  | 'update'
+  | 'patch'
+  | 'delete'
+  | 'deletecollection'
+  | 'watch';
```
andrewballantyne, line 70:
```suggestion
```
This is in [core.ts](https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/lib-core/src/types/core.ts).

andrewballantyne (changes requested):
Looking very close, just a few things left to cleanup for me.

packages/lib-core/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,59 @@
...
+  configurations: {
+    apiDiscovery?: InitApiDiscovery;
+    appFetch: UtilsConfig['appFetch'];
+    /** @deprecated - will be removed later when we have an interface for plugin */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    initPlugins?: (store: Store<any>) => void;
```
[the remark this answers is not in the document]
andrewballantyne: Created https://issues.redhat.com/browse/HAC-758 to cover the creation of the Store & adding the Provider.

packages/lib-utils/src/app/AppInitSDK.tsx
florkbr: Agreed on adding another example without redux. I would like to add that as a follow up to confirm everything behaves as intended in the SDK.

packages/lib-utils/src/app/AppInitSDK.tsx
florkbr: @andrewballantyne apologies - this got lost in the lot of comments

packages/lib-utils/src/types/api-discovery.ts
florkbr: I don't personally think generics are a great idea here, though I'm open to suggestions.

packages/lib-utils/src/types/api-discovery.ts
florkbr: This may change some of the logic in the api discovery reducers/actions, but I'm open to this change. I will change it.

packages/lib-utils/src/app/AppInitSDK.tsx
andrewballantyne: Ticket tracking the work: https://issues.redhat.com/browse/HAC-759

packages/lib-utils/src/types/api-discovery.ts
andrewballantyne: Let us not worry about it right now. The store is absolutely not `unknown` though.

packages/lib-utils/src/types/api-discovery.ts
andrewballantyne: hmmm... maybe this is not a model then? Did I misunderstand what it was?

packages/lib-utils/src/types/k8s.ts
florkbr: @andrewballantyne that type is not exported from the core package. I can update us to use it in the util package, but will also need to change the core package.

florkbr: @andrewballantyne I do not see that type being exported in the `dist` folder for the `core` package. I can pull it in directly from the src folder or export it from the entrypoint?

packages/lib-utils/src/types/api-discovery.ts
florkbr: It is a model, though I can't guarantee what we get back from API discovery will cleanly map into this structure. I'll try it out, but for now I'm fine changing it.

packages/lib-utils/src/types/k8s.ts
andrewballantyne: This should be easily fixed by exporting it normally out the index (or whatever we call the file we will create to handle the split) 🤔 

Not sure why it would be a problem. @vojtechszocs ?

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -25,27 +27,25 @@ type AppInitSDKProps = React.PropsWithChildren<{
...
   React.useEffect(() => {
-    const { appFetch, apiDiscovery = initAPIDiscovery } = configurations;
     try {
       setUtilsConfig({ appFetch });
       apiDiscovery(store);
     } catch (e) {
       consoleLogger.warn(e);
     }
-  }, [configurations, store]);
+  }, [apiDiscovery, appFetch, configurations, store]);
```
andrewballantyne, line 46: Don't make `configurations` a required part 🙂

andrewballantyne (changes requested):

packages/lib-utils/src/types/api-discovery.ts
andrewballantyne: Tracked here: https://issues.redhat.com/browse/HAC-765

andrewballantyne (approved):
Looks good to me.

/hold
Let's make sure Vojtech has his say though.

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { store, storeContextPresent } = useReduxStore();
+  const { appFetch, apiDiscovery = initAPIDiscovery } = configurations;
+  React.useEffect(() => {
+    try {
+      setUtilsConfig({ appFetch });
+      apiDiscovery(store);
+    } catch (e) {
+      consoleLogger.warn(e);
+    }
+  }, [apiDiscovery, appFetch, store]);
```
vojtechszocs, line 46:
```suggestion
  }, [utilsConfig, apiDiscovery, store]);
```

packages/lib-utils/src/app/redux/index.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @example
+ * ```ts
+ * function Component () {
+ *   const {store, storeContextPresent} = useReduxStore()
+ *   return ...
+ * }
+ * ```
+ */
+export const useReduxStore = (): { store: Store<unknown>; storeContextPresent: boolean } => {
```
vojtechszocs, line 22: Nit: I'd suggest extracting the hook's return type and use it here

```ts
type UseReduxStoreResult = {
  store: Store<unknown>;
  storeContextPresent: boolean;
};
```

```suggestion
export const useReduxStore = (): UseReduxStoreResult => {
```

We do a similar thing in `packages/lib-core/src/store/useResolvedExtensions.ts`

packages/lib-utils/src/app/redux/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  const storeContext = useStore();
```
vojtechszocs, line 23:
```suggestion
  const storeInUse = useStore();
```

packages/lib-utils/src/config.ts
```diff
@@ -39,3 +40,27 @@ export const getUtilsConfig = () => {
 
   return config;
 };
+
+let reduxStore: Store;
```
vojtechszocs, line 44: We could make `reduxStore` part of `UtilsConfig` so we won't need another setter function, but we can do this later on.

packages/lib-utils/src/types/api-discovery.ts
vojtechszocs: Generics make sense if we or the consumer can make a reliable assumption about the types involved. If not, using `unknown` is perfectly fine.

vojtechszocs (changes requested):

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+  /** Only child is your Application */
+  children: React.ReactElement;
+  configurations: {
+    apiDiscovery?: InitAPIDiscovery;
+    appFetch: UtilsConfig['appFetch'];
+  };
+};
+
+/**
+ * Component for providing store access to the SDK.
```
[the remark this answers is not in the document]
andrewballantyne: I have a ticket to clean this up - https://issues.redhat.com/browse/HAC-759

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
+import { consoleLogger } from '@monorepo/common';
+import * as React from 'react';
+import { Provider } from 'react-redux';
+import type { UtilsConfig } from '../config';
+import { setUtilsConfig } from '../config';
+import type { InitAPIDiscovery } from '../types/api-discovery';
+import { initAPIDiscovery } from './api-discovery';
+import { useReduxStore } from './redux';
+
+type AppInitSDKProps = {
```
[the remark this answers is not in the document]
andrewballantyne: Not needed for components... they can get the prop type of a component through React + TypeScript. I think we should not make every module export 100 things.

`React.ComponentProps<typeof MyComponent>`

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+      apiDiscovery(store);
+    } catch (e) {
+      consoleLogger.warn(e);
+    }
+  }, [apiDiscovery, appFetch, store]);
+
+  return !storeContextPresent ? <Provider store={store}>{children}</Provider> : children;
+};
+
+export default AppInitSDK;
```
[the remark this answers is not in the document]
andrewballantyne: Not a big fan of people who are not a big fan of default exports 😜

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+import type { InitAPIDiscovery } from '../types/api-discovery';
+import { initAPIDiscovery } from './api-discovery';
+import { useReduxStore } from './redux';
+
+type AppInitSDKProps = {
+  /** Only child is your Application */
+  children: React.ReactElement;
+  configurations: {
+    apiDiscovery?: InitAPIDiscovery;
+    appFetch: UtilsConfig['appFetch'];
```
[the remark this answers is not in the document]
andrewballantyne: Hard disagree. We need to keep this API. Let us not take everything we do internal and push it outwards... this is public API, we can store a ton of things in config if we want and it should not be exposed just because it's in our config.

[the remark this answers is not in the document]
andrewballantyne: So we can make two types now, and do like a `utilsConfig: PublicUtilsConfig` I guess?

packages/lib-utils/src/config.ts
andrewballantyne: This would break your previous comment on exposing all of the UtilsConfig to the API of AppInitSDK

packages/lib-utils/src/app/AppInitSDK.tsx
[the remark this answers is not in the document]
florkbr: For now opting with Andrews point. We can always export it later if we find the need.

packages/lib-utils/src/app/AppInitSDK.tsx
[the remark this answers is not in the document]
florkbr: For now opting to keep the default export if that's ok.

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * return (
+ *  <Provider store={store}>
+ *   <AppInitSDK configurations={{ appFetch, apiDiscovery }}>
+ *      <App />
+ *   </AppInitSDK>
+ *  </Provider>
+ * )
+ * ```
+ */
+const AppInitSDK: React.FC<AppInitSDKProps> = ({ children, configurations }) => {
```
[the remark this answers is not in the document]
florkbr: Not necessary with default export - though we can change this in the future if need be.

packages/lib-utils/src/config.ts
florkbr: Agreed. Will leave as is for now

packages/lib-utils/src/app/redux/index.ts
florkbr: @vojtechszocs proposing we keep `storeContext` for now - as the state variable `storeContextPresent` would also need to be renamed and `storeInUsePresent` sounds a bit odd to me.

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * Component for providing store access to the SDK.
+ * Add this at app-level to make use of app's redux store and pass configurations prop needed to initialize the app, preferred to have it under Provider.
+ * It checks for store instance if present or not.
+ * If the store is there then the reference is persisted to be used in SDK else it creates a new store and passes it to the children with the provider
+ * @component AppInitSDK
+ * @example
+ * ```ts
+ * return (
+ *  <Provider store={store}>
+ *   <AppInitSDK configurations={{ appFetch, apiDiscovery }}>
```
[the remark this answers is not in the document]
florkbr: See thread in slack regarding keeping the existing API: https://coreos.slack.com/archives/C02EQERSV3N/p1646154894536459

packages/lib-utils/src/app/AppInitSDK.tsx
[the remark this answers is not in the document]
andrewballantyne: We need to come to this conclusion here... because one or the other needs to happen...

florkbr:
@vojtechszocs @andrewballantyne can this be merged?

packages/lib-utils/src/app/AppInitSDK.tsx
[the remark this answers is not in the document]
vojtechszocs: @andrewballantyne My bad, I thought `UtilsConfig` is the public API but the actual public API is `AppInitSDK` component and its props.

packages/lib-utils/src/app/AppInitSDK.tsx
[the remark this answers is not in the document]
vojtechszocs: Nice, so we can use `React.ComponentProps` to infer the actual props type of a component.

packages/lib-utils/src/app/AppInitSDK.tsx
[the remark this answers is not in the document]
vojtechszocs: I'm fine with default exports if most people prefer to use them. Let's keep it here.

packages/lib-utils/src/app/AppInitSDK.tsx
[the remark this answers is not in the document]
vojtechszocs: I didn't mean to say we must not use default exports :smile: if most people are good with them, I'll accomodate.

packages/lib-utils/src/config.ts
vojtechszocs: Agreed, sorry for confusion.

PR #29, opened 2022-02-09
Move in extension types

karelhala:
JIRA: [HAC-424](https://issues.redhat.com/browse/HAC-424)

* [x] core.action/provider
* [x] core.action/resource-provider
* [x] core.action/group
* [x] core.action/filter
* [x] core.catalog/item-type
* [x] core.catalog/item-provider
* [x] core.catalog/item-filter
* [x] core.context-provider
* [x] core.flag
* [x] core.flag/model
* [x] core.navigation/href
* [x] core.navigation/resource-ns
* [x] core.navigation/separator
* [x] core.page/resource/list
* [x] core.page/resource/details
* [x] core.redux-reducer
* [x] core.resource/create
* [x] core.resource-metadata
 * this type doesn't exists in console, there's just `console.model-metadata` I've exported this extension type and also `core.resource-metadata`, they are both identical. We should consider deprecating one or the other.
* [x] core.telemetry/listener
* [x] core.yaml-template

Additional extensions:
* [x] core.page/route

### DIFF
--- a/packages/lib-core/package.json
+++ b/packages/lib-core/package.json
@@ -24,6 +24,7 @@
   "peerDependencies": {
     "lodash-es": "^4.17.21",
     "react": "^17.0.2",
+    "redux": "^4.1.2",
     "yup": "^0.32.11"
   },
   "devDependencies": {
@@ -32,8 +33,10 @@
     "@rollup/plugin-typescript": "^8.3.0",
     "@types/lodash-es": "^4.17.5",
     "@types/react": "^17.0.37",
+    "@types/react-router-dom": "^5.1.2",
     "lodash-es": "^4.17.21",
     "react": "^17.0.2",
+    "redux": "^4.1.2",
     "rollup": "^2.61.1",
     "rollup-plugin-analyzer": "^4.0.0",
     "rollup-plugin-dts": "^4.0.1",
--- a/packages/lib-core/src/extensions/actions.ts
+++ b/packages/lib-core/src/extensions/actions.ts
@@ -1,0 +1,128 @@
+import type { ReactNode } from 'react';
+import type { ExtensionK8sResourceIdentifier, ExtensionHook } from '../types/common';
+import type { CodeRef, Extension } from '../types/extension';
+
+/** ActionProvider contributes a hook that returns list of actions for specific context */
+export type ActionProvider = Extension<
+  'core.action/provider',
+  {
+    /** The context ID helps to narrow the scope of contributed actions to a particular area of the application. Ex - topology, helm */
+    contextId: string;
+    /**
+     * A react hook which returns actions for the given scope.
+     * If contextId = `resource` then the scope will always be a K8s resource object
+     */
+    provider: CodeRef<ExtensionHook<Action[]>>;
+  }
+>;
+
+/** ResourceActionProvider contributes a hook that returns list of actions for specific resource model */
+export type ResourceActionProvider = Extension<
+  'core.action/resource-provider',
+  {
+    /** The model for which this provider provides actions for. */
+    model: ExtensionK8sResourceIdentifier & {
+      version: string;
+      kind: string;
+    };
+    /** A react hook which returns actions for the given resource model */
+    provider: CodeRef<ExtensionHook<Action[]>>;
+  }
+>;
+
+/** ActionGroup contributes an action group that can also be a submenu */
+export type ActionGroup = Extension<'core.action/group', Group>;
+
+/** ActionFilter can be used to filter an action */
+export type ActionFilter = Extension<
+  'core.action/filter',
+  {
+    /** The context ID helps to narrow the scope of contributed actions to a particular area of the application. Ex - topology, helm */
+    contextId: string;
+    /**
+     * A function which will filter actions based on some conditions.
+     * scope: The scope in which actions should be provided for.
+     * Note: hook may be required if we want to remove the ModifyCount action from a deployment with HPA
+     */
+    filter: CodeRef<(scope: unknown, action: Action) => boolean>;
+  }
+>;
+
+// Type guards
+
+export const isActionProvider = (e: Extension): e is ActionProvider => {
+  return e.type === 'core.action/provider';
+};
+
+export const isResourceActionProvider = (e: Extension): e is ResourceActionProvider => {
+  return e.type === 'core.action/resource-provider';
+};
+
+export const isActionGroup = (e: Extension): e is ActionGroup => {
+  return e.type === 'core.action/group';
+};
+
+export const isActionFilter = (e: Extension): e is ActionFilter => {
+  return e.type === 'core.action/filter';
+};
+
+// Arbitrary types
+
+type Group = {
+  /** ID used to identify the action section. */
+  id: string;
+  /**
+   * The label to display in the UI.
+   * Required for submenus.
+   */
+  label?: string;
+  /** Whether this group should be displayed as submenu */
+  submenu?: boolean;
+  /**
+   * Insert this item before the item referenced here.
+   * For arrays, the first one found in order is used.
+   */
+  insertBefore?: string | string[];
+  /**
+   Insert this item after the item referenced here.
+   * For arrays, the first one found in order is used.
+   * insertBefore takes precedence.
+   */
+  insertAfter?: string | string[];
+};
+
+type Action = {
+  /** A unique identifier for this action. */
+  id: string;
+  /** The label to display in the UI. */
+  label: ReactNode;
+  /** Subtext for the menu item */
+  description?: string;
+  /**
+   * Executable callback or href.
+   * External links should automatically provide an external link icon on action.
+   */
+  cta: (() => void) | { href: string; external?: boolean };
+  /** Whether the action is disabled. */
+  disabled?: boolean;
+  /** The tooltip for this action. */
+  tooltip?: string | ReactNode;
+  /** The icon for this action. */
+  icon?: string | ReactNode;
+  /**
+   * A `/` separated string where each segment denotes
+   * Eg. `add-to-project`, `menu-1/menu-2`
+   */
+  path?: string;
+  /**
+   * Insert this item before the item referenced here.
+   * For arrays, the first one found in order is used.
+   */
+  insertBefore?: string | string[];
+  /**
+   * Insert this item after the item referenced here.
+   * For arrays, the first one found in order is used.
+   * insertBefore takes precedence.
+   */
+  insertAfter?: string | string[];
+};
--- a/packages/lib-core/src/extensions/catalog.ts
+++ b/packages/lib-core/src/extensions/catalog.ts
@@ -1,0 +1,145 @@
+import type { AnyObject } from '@monorepo/common';
+import type { ReactNode } from 'react';
+import type { ExtensionHook } from '../types/common';
+import type { Extension, CodeRef } from '../types/extension';
+
+export type CatalogItemType = Extension<
+  'core.catalog/item-type',
+  TypeAndTitle & {
+    /** Description for the type specific catalog. */
+    catalogDescription?: string;
+    /** Description for the catalog item type. */
+    typeDescription?: string;
+    /** Custom filters specific to the catalog item.  */
+    filters?: CatalogItemAttribute[];
+    /** Custom groupings specific to the catalog item. */
+    groupings?: CatalogItemAttribute[];
+  }
+>;
+
+export type CatalogItemProvider = Extension<
+  'core.catalog/item-provider',
+  TypeAndTitle & {
+    /** The unique identifier for the catalog this provider contributes to. */
+    catalogId: string | string[];
+    /** Fetch items and normalize it for the catalog. Value is a react effect hook. */
+    provider: CodeRef<ExtensionHook<CatalogItem[], CatalogExtensionHookOptions>>;
+    /** Priority for this provider. Defaults to 0. Higher priority providers may override catalog
+        items provided by other providers. */
+    priority?: number;
+  }
+>;
+
+export type CatalogItemFilter = Extension<
+  'core.catalog/item-filter',
+  {
+    /** The unique identifier for the catalog this provider contributes to. */
+    catalogId: string | string[];
+    /** Type ID for the catalog item type. */
+    type: string;
+    /** Filters items of a specific type. Value is a function that takes CatalogItem[] and returns a subset based on the filter criteria. */
+    filter: CodeRef<(item: CatalogItem) => boolean>;
+  }
+>;
+
+// Type guards
+
+export const isCatalogItemType = (e: Extension): e is CatalogItemType => {
+  return e.type === 'core.catalog/item-type';
+};
+
+export const isCatalogItemProvider = (e: Extension): e is CatalogItemProvider => {
+  return e.type === 'core.catalog/item-provider';
+};
+
+export const isCatalogItemFilter = (e: Extension): e is CatalogItemFilter => {
+  return e.type === 'core.catalog/item-filter';
+};
+
+export type CatalogItem<TData extends AnyObject = AnyObject> = {
+  uid: string;
+  type: string;
+  name: string;
+  /** Optional title to render a custom title using ReactNode.
+   * Rendered in catalog tile and side panel
+   *  */
+  title?: string | ReactNode;
+  // Used as the second label next to the provider label in the list result.
+  secondaryLabel?: string | ReactNode;
+  provider?: string;
+  // Used as the tile description. If provided as a string, the description is truncated to 3 lines.
+  // If provided as a ReactNode, the contents will not be truncated.
+  // This description will also be shown in the side panel if there are no `details.descriptions`.
+  description?: string | ReactNode;
+  tags?: string[];
+  creationTimestamp?: string;
+  supportUrl?: string;
+  documentationUrl?: string;
+  attributes?: AnyObject;
+  cta?: {
+    label: string;
+    href?: string;
+    callback?: (props?: AnyObject) => void;
+  };
+  icon?: {
+    url?: string;
+    class?: string;
+    node?: string | ReactNode;
+  };
+  details?: CatalogItemDetails;
+  // Optional text only badges for the catalog item which will be rendered on the tile and details panel.
+  badges?: CatalogItemBadge[];
+  // Optional data attached by the provider.
+  // May be consumed by filters.
+  // `data` for each `type` of CatalogItem should implement the same interface.
+  data?: TData;
+};
+
+export type CatalogExtensionHookOptions = {
+  namespace: string;
+};
+
+export type ItemType = TypeAndTitle & {
+  /** Description for the type specific catalog. */
+  catalogDescription?: string;
+  /** Description for the catalog item type. */
+  typeDescription?: string;
+  /** Custom filters specific to the catalog item.  */
+  filters?: CatalogItemAttribute[];
+  /** Custom groupings specific to the catalog item. */
+  groupings?: CatalogItemAttribute[];
+};
+
+export type CatalogItemAttribute = {
+  label: string;
+  attribute: string;
+};
+
+export type CatalogItemBadge = {
+  text: string;
+  color?: 'blue' | 'cyan' | 'green' | 'orange' | 'purple' | 'red' | 'grey';
+  icon?: ReactNode;
+  variant?: 'outline' | 'filled';
+};
+
+export type CatalogItemDetails = {
+  properties?: CatalogItemDetailsProperty[];
+  descriptions?: CatalogItemDetailsDescription[];
+};
+
+export type CatalogItemDetailsProperty = {
+  label: string;
+  value: string | ReactNode;
+};
+
+export type CatalogItemDetailsDescription = {
+  label?: string;
+  value: string | ReactNode;
+};
+
+type TypeAndTitle = {
+  /** Type ID for the item type. */
+  type: string;
+  /** Title for the item */
+  title: string | ReactNode;
+};
--- a/packages/lib-core/src/extensions/context-providers.ts
+++ b/packages/lib-core/src/extensions/context-providers.ts
@@ -1,0 +1,18 @@
+import type { Provider } from 'react';
+import type { CodeRef, Extension } from '../types/extension';
+
+/** Adds new React context provider to Console application root. */
+export type ContextProvider<TValue = unknown> = Extension<
+  'core.context-provider',
+  {
+    /** Context Provider component. */
+    provider: CodeRef<Provider<TValue>>;
+    /** Hook for the Context value. */
+    useValueHook: CodeRef<() => TValue>;
+  }
+>;
+
+// Type guards
+
+export const isContextProvider = (e: Extension): e is ContextProvider =>
+  e.type === 'core.context-provider';
--- a/packages/lib-core/src/extensions/model-metadata.ts
+++ b/packages/lib-core/src/extensions/model-metadata.ts
@@ -1,0 +1,29 @@
+import type { ExtensionK8sResourceIdentifier } from '../types/common';
+import type { Extension } from '../types/extension';
+
+/** Customize the display of models by overriding values retrieved and generated through API discovery. */
+export type ModelMetadata = Extension<'core.model-metadata', Metatdata>;
+
+// Type guards
+
+export const isModelMetadata = (e: Extension): e is ModelMetadata =>
+  e.type === 'core.model-metadata';
+
+// Arbitrary types
+
+export type Metatdata = {
+  /** The model to customize. May specify only a group, or optional version and kind. */
+  model: ExtensionK8sResourceIdentifier & {
+    group: string;
+  };
+  /** Whether to consider this model reference as tech preview or dev preview. */
+  badge?: 'tech' | 'dev';
+  /** The color to associate to this model. */
+  color?: string;
+  /** Override the label. Requires `kind` be provided. */
+  label?: string;
+  /** Override the plural label. Requires `kind` be provided. */
+  labelPlural?: string;
+  /** Customize the abbreviation. Defaults to All uppercase chars in the kind up to 4 characters long. Requires `kind` be provided. */
+  abbr?: string;
+};
--- a/packages/lib-core/src/extensions/pages.ts
+++ b/packages/lib-core/src/extensions/pages.ts
@@ -1,0 +1,56 @@
+import type { ComponentType } from 'react';
+import type { RouteComponentProps } from 'react-router';
+import type { ExtensionK8sResourceIdentifier } from '../types/common';
+import type { CodeRef, Extension } from '../types/extension';
+
+/** Adds new page to host application's React router. */
+export type RoutePage = Extension<'core.page/route', RoutePageProperties>;
+
+/** Adds new resource list page to host application's React router. */
+export type ResourceListPage = Extension<'core.page/resource/list', ResourcePageProperties>;
+
+/** Adds new resource details page to host application's React router. */
+export type ResourceDetailsPage = Extension<'core.page/resource/details', ResourcePageProperties>;
+
+// Type guards
+
+export const isRoutePage = (e: Extension): e is RoutePage => e.type === 'core.page/route';
+export const isResourceListPage = (e: Extension): e is ResourceListPage =>
+  e.type === 'core.page/resource/list';
+export const isResourceDetailsPage = (e: Extension): e is ResourceDetailsPage =>
+  e.type === 'core/Resource/details';
+
+// Arbitrary types
+
+export type RoutePageProperties = {
+  /** The perspective to which this page belongs to. If not specified, contributes to all perspectives. */
+  perspective?: string;
+  /** The component to be rendered when the route matches. */
+  component: CodeRef<ComponentType<RouteComponentProps>>;
+  /** Valid URL path or array of paths that `path-to-regexp@^1.7.0` understands. */
+  path: string | string[];
+  /** When true, will only match if the path matches the `location.pathname` exactly. */
+  exact?: boolean;
+};
+
+export type ResourcePageProperties = {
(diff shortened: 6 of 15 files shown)


karelhala:
ping @vojtechszocs

packages/lib-core/src/api/common-types.ts
```diff
@@ -0,0 +1,40 @@
+import type { AnyObject } from '@monorepo/common';
+
+// Type for extension hook
+export type ExtensionHook<TType, TOptions extends AnyObject = AnyObject> = (
+  options: TOptions,
+) => ExtensionHookResult<TType>;
+
+// Type for extension hook result that returns [data, resolved, error]
+export type ExtensionHookResult<TType> = [TType, boolean, unknown];
```
andrewballantyne, line 9: `ReturnType<ExtensionHook<...>>` will return this value. I think we can avoid exporting this.

andrewballantyne (changes requested):
Partial review... ran out of time. Will continue tomorrow.

packages/lib-core/src/api/common-types.ts
andrewballantyne: It should see the [TypeScript Playground example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggdiA8gIwFYQMbCgXigbwF8BuKAejKgDMB7AJymAgGdgBLOAcwBpGALNsygYAhnCjJoYkMAFcAsACglFKABVw0WgwgAPJnGZsa4vjRoBrJXrD1soSFACi+iIeNwAEuYsAeNRqQvGqIYOwmQnoGACZC8EhomNh48SjoWAB8uFAAFEpQUDRhHswAXOqh4YbcSgCUuFkuBkYm3pYASiwArgA2wP6BEBnESiqUg9T0UFFuLaY+UHTdffwi2EvAXXSGUADa0WsivEvMND0AbhDRvBB0dPQAuta6tnT2ms6u7q0+ncy9-QCmiyeF2QKCEnMPQgYl4XTgFjgNAA7nAHiNlJiHNAAMo0AC2EAAwl1WASJnh8FA4CJCeVWHQOJwoCRRopVABZGik6A0S4MAAGahYwAFjBoUGYEGgshlHzY2D4ImYSmx6hF2U6m22g18TVmHjafjxhJJZPxgwyGSAA) I coded up to test it.

packages/lib-core/src/api/common-types.ts
```diff
@@ -0,0 +1,40 @@
...
+  group: string;
+  version?: string;
+  kind: string;
+};
+
+export type ExtensionK8sGroupModel = {
+  group: string;
+  version?: string;
+  kind?: string;
+};
```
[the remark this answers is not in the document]
vojtechszocs: @andrewballantyne AFAIK this was discussed with @spadgett and the consensus was to have a single way to identify k8s resources - check out `K8sResourceIdentifier` type in #30 - group is optional but version and kind are mandatory:

```ts
type K8sResourceIdentifier = {
  apiGroup?: string;
  apiVersion: string;
  kind: string;
};
```

Notice that k8s APIs use `apiGroup` instead of `group` and `apiVersion` instead of `version` - I wouldn't change that in `K8sResourceIdentifier` type, as they reflect the k8s API response data.

TL;DR - I'd suggest introducing a single type via e.g. `packages/lib-core/src/types/extension-k8s.ts` which is conceptually equivalent to above.

```ts
type ExtensionK8sResourceIdentifier = {
  group?: string;
  version: string;
  kind: string;
};
```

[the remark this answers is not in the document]
vojtechszocs: It's technically possible to reuse `K8sResourceIdentifier` from `lib-utils` and rename the group/version property names (like [this](https://stackoverflow.com/a/52702528)) but that would introduce `lib-core` => `lib-utils` (or the opposite) dependency, which we generally want to avoid.

Core k8s types should be kept in `lib-utils` since this package deals primarily with k8s API calls, among other things.

packages/lib-core/src/api/common-types.ts
```diff
@@ -0,0 +1,40 @@
...
+  group?: string;
+  version: string;
+  kind: string;
+};
+
+export type TypeAndTitle = {
+  /** Type ID for the item type. */
+  type: string;
+  /** Title for the item */
+  title: string | React.ReactNode;
```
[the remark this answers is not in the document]
vojtechszocs: The `React.ReactNode` part should definitely fail without importing React as `React`, this is weird.

packages/lib-core/src/api/common-types.ts
vojtechszocs: We already have `UseResolvedExtensionsResult<TExtension extends Extension>` type declared in `packages/lib-core/src/store/useResolvedExtensions.ts` so we should reuse that one.

The copied-over `ExtensionHook` type actually refers to the result of `useResolvedExtensions` hook.

andrewballantyne: I'm fine with it either way.

My thoughts are - I think we could just use the `ReturnType` for types that have no name-benefit. It's a 3-param array and is just a collection of responses. A name will not help us use it anymore than if it was a `ReturnType<...>` result.

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
+import type { ReactNode } from 'react';
+import type { ExtensionHook, ExtensionK8sKindVersionModel } from '../api/common-types';
+import type { CodeRef, Extension } from '../types/extension';
+import type { AccessReviewResourceAttributes } from './core-types';
+
+/** ActionProvider contributes a hook that returns list of actions for specific context */
+export type ActionProvider = Extension<
```
vojtechszocs, line 7: FYI @andrewballantyne our `Extension` type has two generic type parameters, as opposed to `Extension` type in Console (can't change the latter easily since it's already used by a lot of code).

This eliminates the need for `ExtensionDeclaration` type in this SDK.

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
...
+  'core.action/provider',
```
vojtechszocs, line 8: @andrewballantyne We should consider writing a custom ESLint rule that checks all specific extension types and ensures the first generic type parameter is a string literal that starts with `core.` prefix.

https://typescript-eslint.io/docs/development/custom-rules/

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
...
+  {
+    /** The context ID helps to narrow the scope of contributed actions to a particular area of the application. Ex - topology, helm */
+    contextId: string | 'resource';
```
vojtechszocs, line 11: TS coerces this to just `string` so I'm not sure why we need the `| 'resource'` part.

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
...
+    /** A react hook which returns actions for the given scope.
+     * If contextId = `resource` then the scope will always be a K8s resource object
+     * */
```
vojtechszocs, line 14: Nit: please sanitize the doc comments, ideally it should be either
```ts
/** one-line comment */
```
or
```ts
/**
 * multi-line comment
 * multi-line comment
 */
```

Note that JSDoc comments should always start with `/**` and not with `/*`.

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
...
+    provider: CodeRef<ExtensionHook<Action[]>>;
```
vojtechszocs, line 15: I'll look into all `ExtensionHook<T>` type references as part of a second round of code review.

I'm not sure why we need to reference the signature of the extension consuming hook itself, I'd assume this is primarily due to type/signature commonality in which case we should introduce a type with a more suitable name.

@andrewballantyne FYI this needs looking into.

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
...
+export type ActionFilter = Extension<
+  'core.action/filter',
+  {
+    /** The context ID helps to narrow the scope of contributed actions to a particular area of the application. Ex - topology, helm */
+    contextId: string | 'resource';
+    /** A function which will filter actions based on some conditions.
+     * scope: The scope in which actions should be provided for.
+     * Note: hook may be required if we want to remove the ModifyCount action from a deployment with HPA
+     * */
+    filter: CodeRef<(scope: unknown, action: Action) => boolean>;
```
vojtechszocs, line 43: I guess `scope` was originally typed as `any` and replaced with `unknown` type.

Ideally we would provide a better type here, it should be more apparent from how the extension is used.

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+>;
+
+// Type guards
+
+export const isActionProvider = (e: Extension): e is ActionProvider => {
```
vojtechszocs, line 49: These type guards are technically boilerplate code, using the following template:

```
export const is${ExtensionTypeName} = (e: Extension): e is ${ExtensionTypeName} => {
  return e.type === ${ExtensionTypeLiteralValue};
};
```

I'd be happier if we could generate them to avoid having to maintain them manually.

@andrewballantyne We should explore using TypeScript custom transformers to generate those type guards during TS compilation. The Rollup plugin we're using `@rollup/plugin-typescript` already supports this TS compiler feature.

https://github.com/madou/typescript-transformer-handbook

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
...
+  return e.type === 'core.action/group';
+};
+
+export const isActionFilter = (e: Extension): e is ActionFilter => {
+  return e.type === 'core.action/filter';
+};
+
+// Arbitrary types
+
+export type Group = {
```
vojtechszocs, line 67: By convention, these are helper types which are relevant for / used in this file only.

packages/lib-core/src/extensions/core-types.ts
```diff
@@ -0,0 +1,18 @@
+export type K8sVerb =
```
vojtechszocs, line 1: Similar to my above comment, let's have a consistent way to represent all types instead of spreading them over the source tree.

@andrewballantyne If we identify an overlap / commonality between k8s `lib-core` types vs. k8s `lib-utils` types, we can move them into the `common` package, whose code & types is bundled into all libraries that use it.

packages/lib-core/src/extensions/pages.ts
```diff
@@ -0,0 +1,56 @@
...
+import type { ExtensionK8sGroupKindModel, ExtensionK8sModel } from '../api/common-types';
+import type { CodeRef, Extension } from '../types/extension';
+
+/** Adds new page to Console router. */
+export type RoutePage = Extension<'core.page/route', RoutePageProperties>;
+
+/** Adds new resource list page to Console router. */
+export type ResourceListPage = Extension<
+  'core.page/resource/list',
+  ResourcePageProperties & AnyObject
```
vojtechszocs, line 13: @andrewballantyne FYI the `& AnyObject` was a workaround related to JSON schema generation via [ts-json-schema-generator](https://github.com/vega/ts-json-schema-generator) tool.

I think we should remove this workaround here and on other places.

vojtechszocs (changes requested):

packages/lib-core/package.json
```diff
@@ -40,5 +40,9 @@
     "tslib": "^2.3.1",
     "typescript": "~4.4.4",
     "yup": "^0.32.11"
+  },
+  "dependencies": {
+    "@types/react-router-dom": "5.1.2",
+    "redux": "^4.1.2"
```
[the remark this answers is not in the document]
karelhala: Good point! I'll move the redux to devDeps and peerDeps as for the yup and lodash-es I'd rather not make this PR bigger than it needs to be.

packages/lib-core/src/api/common-types.ts
```diff
@@ -0,0 +1,38 @@
...
+  options: TOptions,
+) => [TType, boolean, unknown];
+
+export type ExtensionK8sKindVersionModel = {
+  group?: string;
+  version: string;
+  kind: string;
+};
+
+export type TypeAndTitle = {
```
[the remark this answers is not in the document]
karelhala: Usually the `title` is optional. I was considering making it optional in here as well and for those usecases where the title was required to extend it. But that was less often, than keeping the `title` required (only one other place had it optional).

This type was introduced solely in this PR, as I said it felt a bit better to have it as common type than to write 5 times the same type signature.

packages/lib-core/src/api/common-types.ts
```diff
@@ -0,0 +1,38 @@
...
+  kind: string;
+};
+
+export type ExtensionK8sGroupKindModel = {
+  group: string;
+  version?: string;
+  kind: string;
+};
+
+export type ExtensionK8sGroupModel = {
```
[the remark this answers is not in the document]
karelhala: I'll mark all as optional and where some will be required I'll mark them as optional. What about backwards compatibility? Should we export the type under multiple values?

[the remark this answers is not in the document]
karelhala: A bit better approach, I've added `ExtensionK8sResourceIdentifier` which has all values optional and kept the previous types just extending this one.

packages/lib-core/src/extensions/actions.ts
```diff
@@ -0,0 +1,119 @@
+import type { ReactNode } from 'react';
```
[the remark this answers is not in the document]
karelhala: Mhh, that won't really work for types. With the above import `import * as React from 'react'` it is throwing an error of `All imports in the declaration are only used as types. Use 'import type'.` I can turn that rule off, but it feels a bit clunky to import every React type just to use one of them.

packages/lib-core/src/extensions/actions.ts
karelhala: Yeah, this felt a bit funky when I moved it over. But I thought it had some purpose when type suggestion. Removing the `| 'resource'` part!

packages/lib-core/src/extensions/actions.ts
karelhala: Yeah, good point!

packages/lib-core/src/extensions/actions.ts
karelhala: I'm not sure I can provide much information what the `scope` can be here :disappointed:

packages/lib-core/src/extensions/pages.ts
karelhala: We're using the `AnyObject` in frontend-components repository to consume additional props. Could be usefull in these extension definitionas (if someone implements them they can add custom props as well). But if we want to prevent this let's nuke the usage of `AnyObject` here.

packages/lib-core/src/extensions/actions.ts
christianvogt: The `scope` is different for every context. For example in topology the `scope` would be a scene graph element, while for a resource page, the scope would be the resource object itself.
The intent was for reusing the same extension for supplying actions in different contexts. But maybe it makes more sense to use completely different extension types that inherit the same structure ?

packages/lib-core/src/types/core.ts
```diff
@@ -0,0 +1,18 @@
...
+  | 'watch';
+
+export type AccessReviewResourceAttributes = {
+  group?: string;
+  resource?: string;
+  subresource?: string;
+  verb?: K8sVerb;
+  name?: string;
+  namespace?: string;
+};
```
andrewballantyne, line 18: I feel this is an "OpenShift Console"-esk of a type and not an accurate type. We had a serious problem (less so now) about types in OpenShift Console being fully optional... making our code a mess of if-statements.

May need a bit of investigation 🤔

andrewballantyne (changes requested):

packages/lib-core/src/extensions/actions.ts
christianvogt: `scope` should follow the same typing as `ExtensionHook` which accepts a typed parameter in the function call.
Unsure if `unknown` or `any` is more appropriate here but the type can be cast to a specific type based on the `contextId`.

christianvogt: resource actions will be important to bring over for supporting details and list pages which present resource actions to the user.

karelhala: So the function type can be written as:
```TS
type FilterType<TType> = (scope: TType, action: Action) => boolean
```

packages/lib-core/src/types/extension.ts
```diff
@@ -1,6 +1,9 @@
 import type { AnyObject, ReplaceProperties } from '@monorepo/common';
 
-export type Extension<TType extends string = string, TProperties extends AnyObject = AnyObject> = {
+export type Extension<
+  TType extends string | AnyObject[] = string,
```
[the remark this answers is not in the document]
karelhala: Ah, yeah. This was my blind attempt to change the type for `Extension` as suggested in here https://github.com/openshift/dynamic-plugin-sdk/pull/29#discussion_r804916683 will rol back to the previous one.

packages/lib-core/src/types/common.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export type TypeAndTitle = {
+  /** Type ID for the item type. */
+  type: string;
+  /** Title for the item */
+  title: string | ReactNode;
+};
+
+// Backwards compatible types
```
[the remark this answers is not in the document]
karelhala: Since this was copied over from console I thought it would be nice to keep them so we don't break the apps when migrating to the new SDK.

packages/lib-core/src/types/common.ts
```diff
@@ -0,0 +1,42 @@
...
+  options: TOptions,
+) => [TType, boolean, unknown];
+
+export type ExtensionK8sResourceIdentifier = {
+  group?: string;
+  version?: string;
+  kind?: string;
+};
+
+export type TypeAndTitle = {
```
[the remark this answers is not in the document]
karelhala: There were 3 places where the extension had `type` and `title` I didn't wanted to blindly copy them over just out of spite. We could in theory keep them in `catalog` extension (and not export it) as that's the place where this type is used.

packages/lib-core/src/extensions/actions.ts
andrewballantyne: Not sure the type necessarily helps in this case. The user is a plugin. They define their functionality -- the caller is the App (in this case) and they will provide what they provide based on where they call it.

The Generic type won't translate well between the two I don't think. It's a weird concept. Not sure why we have such a loose extension. I think we can leave it as `unknown` for now imo. The plugin can type it when they know the use-case they built it for for their own internal function typing.

These action types are weird 😕

packages/lib-core/src/extensions/actions.ts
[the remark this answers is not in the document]
vojtechszocs: Oh right, @karelhala we should use `import type` whenever all imported variables refer to types.

When we need to import both some types and values from some module, there would be two import statements, one `import` and one `import type`.

packages/lib-core/src/extensions/actions.ts
vojtechszocs: Right, we don't need two types because we have the ability to change the base `Extension` one :smiley: in other words, `Extension` with `type: string` in Console is a legacy bit that we can mitigate here.

packages/lib-core/src/extensions/actions.ts
vojtechszocs: > as part of a second round of code review

Can be done after merging this PR, I don't want to hold it on this.

@andrewballantyne My concern was why do we even need to know the signature of `useResolvedExtension` hook (function) when dealing with extension-related types :smiley: is it because of structural similarity, or other reason? As I wrote, we can improve this later on.

vojtechszocs: To elaborate, extension-related types (i.e. ones referenced by specific `Extension` declarations) are used to _describe_ specific extension points in the application; basically using data and/or callback functions.

I'm missing the logical link to `useResolvedExtension` which is a function used to consume/interpret extensions.

packages/lib-core/src/extensions/actions.ts
vojtechszocs: > The Generic type won't translate well between the two I don't think.

Right, if it's generic, both app and plugin need to make an assumption about it.

packages/lib-core/src/extensions/actions.ts
vojtechszocs: This can also be handled as part of [HAC-376](https://issues.redhat.com/browse/HAC-376).

packages/lib-core/src/extensions/pages.ts
vojtechszocs: I think the default should be to explicitly list all possible properties (mandatory and/or optional) and disallow unknown/custom ones.

There may be use cases where we can allow that, but I think it shouldn't be the default pattern.

packages/lib-core/src/types/extension.ts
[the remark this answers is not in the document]
vojtechszocs: Yup, extension's `type` property should always be `string` at runtime, but here we allow it to be narrowed-down to a specific literal type. (In TypeScript, `'foo'` type is assignable to `string` type, but not the other way around.)

packages/lib-core/src/types/common.ts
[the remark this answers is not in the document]
vojtechszocs: Right, let's not bring in unnecessary complexity if possible.

.gitignore
```diff
@@ -12,3 +12,5 @@
 
 # Generated project artifacts
 **/dist
+
+__coverage__
```
vojtechszocs, line 16: I think this is out of scope for this PR :smiley:

packages/lib-core/package.json
```diff
@@ -32,8 +33,10 @@
     "@rollup/plugin-typescript": "^8.3.0",
     "@types/lodash-es": "^4.17.5",
     "@types/react": "^17.0.37",
+    "@types/react-router-dom": "5.1.2",
```
vojtechszocs, line 36: Any reason not to use a semver pattern, but a specific version? (Just asking, I'm OK with this change.)

packages/lib-core/src/extensions/actions.ts
vojtechszocs: OK just realized it's because we want to pass React hook via `provider` property.

packages/lib-core/src/extensions/actions.ts
vojtechszocs: I feel it's still better typed with a generic type param (falling back to `unknown`) but we can add it later.

packages/lib-core/src/extensions/pages.ts
vojtechszocs: Let's nuke `& AnyObject` part for now, its original purpose was to make https://github.com/vega/ts-json-schema-generator happy in context of Console plugin SDK.

packages/lib-core/src/extensions/pages.ts
```diff
@@ -0,0 +1,63 @@
...
+>;
+
+/** Adds new resource details page to Console router. */
+export type ResourceDetailsPage = Extension<
+  'core.page/resource/details',
+  ResourcePageProperties & AnyObject
```
vojtechszocs, line 19: Let's nuke `& AnyObject` part for now, its original purpose was to make https://github.com/vega/ts-json-schema-generator happy in context of Console plugin SDK.

packages/lib-core/src/types/common.ts
```diff
@@ -0,0 +1,42 @@
+import type { AnyObject } from '@monorepo/common';
+import type { ReactNode } from 'react';
+
+// Type for extension hook
+export type ExtensionHook<TType, TOptions extends AnyObject = AnyObject> = (
+  options: TOptions,
+) => [TType, boolean, unknown];
```
vojtechszocs, line 7: TypeScript 4+ allows tuple elements to be named, we already do that in tuple type `UseResolvedExtensionsResult`. We should improve this in a follow-up change.

packages/lib-core/src/types/common.ts
```diff
@@ -0,0 +1,42 @@
...
+export type TypeAndTitle = {
+  /** Type ID for the item type. */
+  type: string;
+  /** Title for the item */
+  title: string | ReactNode;
+};
+
+// Backwards compatible types
+
+export type ExtensionK8sKindVersionModel = ExtensionK8sResourceIdentifier & {
```
vojtechszocs, line 24: We can create mapped type helpers, for example:

```ts
type Bar = RequiredProperties<Foo, 'a' | 'b'>; // remove optional modifier for given properties of Foo
type Bar = OptionalProperties<Foo, 'a' | 'b'>; // add optional modifier for given properties of Foo
```

Can be done later as follow-up improvement.

packages/lib-core/src/types/core.ts
vojtechszocs: Ping, if we don't need `AccessReviewResourceAttributes` let's remove it for now.

vojtechszocs (changes requested):

.gitignore
karelhala: yeah :smile:

packages/lib-core/package.json
karelhala: The router v6 is not backwards compatible. But good point about adding carret in this one.

packages/lib-core/src/extensions/pages.ts
karelhala: Sure thing!

packages/lib-core/src/extensions/pages.ts
karelhala: Removing it.

packages/lib-core/src/types/common.ts
karelhala: Yeah, that makes sense. I'll keep as is, and we can iterate on it.

packages/lib-core/src/types/core.ts
karelhala: It's used in `actions.ts` I'll remove the optional prop that is typed by this.

andrewballantyne (approved):
Looks good to me. Waiting on @vojtechszocs final review & the test passes 🙂

packages/lib-core/src/extensions/catalog.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export const isCatalogItemProvider = (e: Extension): e is CatalogItemProvider => {
+  return e.type === 'core.catalog/item-provider';
+};
+
+export const isCatalogItemFilter = (e: Extension): e is CatalogItemFilter => {
+  return e.type === 'core.catalog/item-filter';
+};
+
+export type CatalogItem<TType extends AnyObject = AnyObject> = {
```
vojtechszocs, line 59:
```suggestion
export type CatalogItem<TData extends AnyObject = AnyObject> = {
```

packages/lib-core/src/extensions/context-providers.ts
```diff
@@ -0,0 +1,18 @@
+import type { Provider } from 'react';
+import type { CodeRef, Extension } from '../types/extension';
+
+/** Adds new React context provider to Console application root. */
+export type ContextProvider<TType = unknown> = Extension<
```
vojtechszocs, line 5:
```suggestion
export type ContextProvider<TValue = unknown> = Extension<
```

packages/lib-core/src/types/common.ts
```diff
@@ -0,0 +1,12 @@
+import type { AnyObject } from '@monorepo/common';
+
+// Type for extension hook
+export type ExtensionHook<TType, TOptions extends AnyObject = AnyObject> = (
```
vojtechszocs, line 4:
```suggestion
export type ExtensionHook<TResult, TOptions extends AnyObject = AnyObject> = (
```

packages/lib-core/src/types/common.ts
```diff
@@ -0,0 +1,12 @@
...
+  options: TOptions,
+) => [TType, boolean, unknown];
+
+export type ExtensionK8sResourceIdentifier = {
```
vojtechszocs, line 8: Please add the following comment above:
```
// TODO(vojtech): use types like RequiredProperties<T, K> and OptionalProperties<T, K>
// to change optionality of specific properties, instead of redefining them via "&" intersection
```

packages/lib-core/src/extensions/model-metadata.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/** Customize the display of models by overriding values retrieved and generated through API discovery. */
+export type ModelMetadata = Extension<'core.model-metadata', Metatdata>;
+
+// Type guards
+
+export const isModelMetadata = (e: Extension): e is ModelMetadata =>
+  e.type === 'core.model-metadata';
+
+// Arbitraty types
```
vojtechszocs, line 12:
```suggestion
// Arbitrary types
```

vojtechszocs (changes requested):

packages/lib-core/src/store/useResolvedExtensions.ts
```diff
@@ -12,7 +12,7 @@ import { isExtensionCodeRefsResolutionError, resolveCodeRefValues } from './code
 import { usePluginManager } from './PluginStoreContext';
 import { useExtensions } from './useExtensions';
 
-type UseResolvedExtensionsResult<TExtension extends Extension> = [
+export type UseResolvedExtensionsResult<TExtension extends Extension> = [
```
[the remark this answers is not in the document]
karelhala: Yup, good point!

PR #30, opened 2022-02-09
Add utilities for working with k8s resources

vojtechszocs:
Fixes [HAC-336](https://issues.redhat.com/browse/HAC-336)

- add `k8sBaseURL` option to `UtilsConfig`
- move `consoleLogger` to `common` package for better reusability
- `commonFetchJSON` & related functions now have a mandatory `TResult` type parameter
- add Kubernetes utilities that delegate to the corresponding `commonFetchJSON` function
  - `k8sGet`
  - `k8sCreate`
  - `k8sUpdate`
  - `k8sPatch`
  - `k8sDelete`
  - `k8sList`
  - `k8sListItems` (calls `k8sList` and returns `result.items`)


packages/lib-utils/src/config.ts
```diff
@@ -10,6 +10,11 @@ export type UtilsConfig = {
    * with an appropriate error.
    */
   appFetch: ResourceFetch;
+
+  /**
+   * Kubernetes API server base URL (without the trailing slash).
+   */
+  k8sBaseURL: string;
```
andrewballantyne, line 17: Interesting... I forgot this was jammed in at building of script time. I think we need to rethink this. 

I want to propose we actually do not do this in the SDK. I think this is on the App and thus should be something they prepend in their `appFetch` impl. Thus making our call in [commonFetch](https://github.com/openshift/dynamic-plugin-sdk/blob/6028af97c376111189edafcf59aea7a74d6881e0/packages/lib-utils/src/utils/common-fetch.ts#L14) being a protocol-less, host-less URL part. The host app can do:

```
const appFetch = (url: string, ...) => {
  //...
  fetch(`${baseURL}${url}`, ...);
  //...
}
```
This gives the power to the App to redirect as needed per-call. Some may go to KCP some can go to DevSandbox, etc etc etc... This gives more flexibility for us to configure out way through a multiple backend - one API situation.

Thoughts?

packages/lib-utils/src/index.ts
```diff
@@ -1,2 +1,11 @@
 export { UtilsConfig, setUtilsConfig, getUtilsConfig } from './config';
 export { commonFetch, commonFetchText, commonFetchJSON } from './utils/common-fetch';
+export {
+  k8sGet,
+  k8sCreate,
+  k8sUpdate,
+  k8sPatch,
+  k8sDelete,
+  k8sList,
+  k8sListItems,
```
andrewballantyne, line 10: We went with `k8s{crud}Resource` for these calls... I wonder if we want to keep that name 🤔 

I'm indifferent, not sure it matters... but it'll likely need to be aliased in OpenShift Console. Also HAC-Dev believes the calls will have a suffixed `...Resource` so the adoption will take an extra hit when they have the NPM package included.

packages/lib-utils/src/k8s/k8s-resource.ts
```diff
@@ -0,0 +1,126 @@
...
+import { commonFetchJSON } from '../utils/common-fetch';
+import { getK8sResourceURL } from './k8s-utils';
+
+type K8sResourceReadArgs = [
+  model: K8sModelCommon,
+  options?: QueryOptions,
+  ...fetchOptions: FetchOptionArgs
+];
+
+type K8sResourceUpdateArgs<TResource extends K8sResourceCommon> = [
```
andrewballantyne, line 13: Q: I assumed T stood for "Targeted" -- did you have another thought?

packages/lib-utils/src/types/k8s.ts
```diff
@@ -0,0 +1,60 @@
...
+  fieldSelector: string;
+  resourceVersion: string;
+  [key: string]: string;
+}>;
+
+export type Patch = {
+  op: string;
+  path: string;
+  value?: unknown;
+};
```
andrewballantyne, line 60: Are these so much k8s types as API types?

Maybe the `Patch` is K8s... but I feel we should be careful about calling everything K8s types... as the SDK is K8s focused, so everything will quickly end up here.

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  if (path) {
+    resourcePath += `/${path}`;
+  }
+
+  if (queryParams && !_.isEmpty(queryParams)) {
+    resourcePath += `?${getQueryString(queryParams)}`;
+  }
+
+  return `${getUtilsConfig().k8sBaseURL}/${resourcePath}`;
```
andrewballantyne, line 52: To my other comment, I am suggesting this:

```suggestion
  return `/${resourcePath}`;
```

andrewballantyne (changes requested):
Also I think we need the object call implementation for the CRUD utils... It's more preferred than the multiple arg approach.

packages/lib-utils/src/config.ts
vojtechszocs: I'm OK with this, but if we do this, we should rename `url` to `path` because it'd be just a part of the whole URL.

packages/lib-utils/src/index.ts
vojtechszocs: I'm also indifferent here, I can add the `Resource` suffix if it's more meaningful.

packages/lib-utils/src/k8s/k8s-resource.ts
vojtechszocs: `T` prefix for type parameters stands for "type" :smiley: another example would be `TValue` which means "value type"

```ts
export type CodeRef<TValue = unknown> = () => Promise<TValue>;
```

packages/lib-utils/src/types/k8s.ts
vojtechszocs: Yup, we can break it into smaller chunks if needed.

For now, I've put all types related to `k8s-resource.ts` here.

packages/lib-utils/src/index.ts
andrewballantyne: Let's go with the `Resource` suffix, everything is easier that way.

packages/lib-utils/src/config.ts
andrewballantyne: Agreed, `path` rename as well please

packages/lib-utils/src/k8s/k8s-resource.ts
andrewballantyne: Ah, dur lol -- thanks!

andrewballantyne (approved):

PR #31, opened 2022-02-11
Setup global tests and add simple test

karelhala:


config/globalSetup.ts
```diff
@@ -0,0 +1,6 @@
+require('ts-node/register');
```
andrewballantyne, line 1: There is a lot of vague references to what this does in the `ts-node` README... can I get a quick one liner on what this is helping us with?

My guess is it is to setup the commonJS stuff for node, but this can be added to the `tsconfig` if I understand the setup correctly.

config/setupTests.ts
```diff
@@ -0,0 +1,12 @@
+import '@testing-library/jest-dom';
+
+jest.mock('react', () => {
+  const originalModule = jest.requireActual('react');
+
+  return {
+    __esModule: true,
+    ...originalModule,
+    useLayoutEffect: jest.requireActual('react').useEffect,
```
andrewballantyne, line 9: I'm confused why you added this... can you give me some context? This mock feels unnecessary.

package.json
```diff
@@ -7,6 +7,20 @@
   "scripts": {
     "build": "yarn workspaces foreach run build",
     "lint": "yarn workspaces foreach run lint",
-    "test": "yarn workspaces foreach run test"
+    "test": "jest"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.17.2",
+    "@babel/preset-env": "^7.16.11",
+    "@babel/preset-typescript": "^7.16.7",
```
andrewballantyne, line 15: We have TS, Babel shouldn't be needed.

package.json
```diff
@@ -7,6 +7,20 @@
...
+    "@testing-library/dom": "^8.11.3",
+    "@testing-library/jest-dom": "^5.16.2",
+    "@testing-library/react": "^12.1.2",
+    "@testing-library/react-hooks": "^7.0.2",
+    "@testing-library/user-event": "^13.5.0",
+    "@types/jest": "^27.4.0",
+    "jest": "^27.5.1",
+    "ts-jest": "^27.1.3",
+    "ts-node": "^10.5.0"
   }
```
andrewballantyne, line 25: @vojtechszocs did we want to put our jest stuff in another package? I know you've been trying to keep the root package as clean as possible.

packages/lib-core/src/utils/__tests__/logger.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { consoleLogger } from '../logger';
+
+describe('consoleLogger', () => {
```
andrewballantyne, line 3: Anyway we can test for env variables? Because that'd be a good test too to make sure `info` is not doing anything in production.

andrewballantyne (changes requested):
Looking pretty good... few questions and concerns. Please let me know what you think @karelhala

andrewballantyne:
I'm not sure why `ci/prow/all-packages` failed... only thing I can see is unmet peer dependencies -- but I don't think that fails the CI.

config/setupTests.ts
karelhala: There's a warning when using `useLayoutEffect` that is has no effect on server https://github.com/mui/material-ui/issues/15798 we are not using it anywhere (yet), but it helps to add it now so we don't have to deal with it later down the line.

packages/common/package.json
```diff
@@ -5,8 +5,7 @@
   "private": true,
   "main": "src/index.ts",
   "scripts": {
-    "lint": "yarn lint:run --ext .ts,.tsx src rollup-configs.js",
-    "test": "echo 'Write some tests!'"
```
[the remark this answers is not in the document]
karelhala: Mhh, interesting idea to add test script for each package. I'll add them so we don't have to run all tests, but we can choose which one to run. Usually what I do is `yarn test packages/lib-utils` but we can do that for the developer in the test script.

packages/lib-core/src/utils/__tests__/logger.spec.ts
karelhala: Yeah, we can do that! I'll ad it there.

config/setupTests.ts
andrewballantyne: Interesting. Useful for sure... we might have utility components that could use this hook 🤔 

cc @vojtechszocs -- thoughts?

config/globalSetup.ts
karelhala: I was not aware of the `ts-node` config option, let me give that a try.

package.json
karelhala: Ah yeah, I was having a hard time setting up JS files with TS.

config/globalSetup.ts
andrewballantyne: For posterity, [ts-node docs](https://github.com/TypeStrong/ts-node#via-tsconfigjson-recommended)

packages/lib-core/src/utils/__tests__/logger.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    jest.resetModules();
+  });
+
+  test('should have all console loggers', () => {
+    expect(consoleLogger.warn).toBeDefined();
+    expect(consoleLogger.error).toBeDefined();
+    expect(consoleLogger.info).toBeDefined();
+  });
+
+  test('should have all console loggers except info', () => {
```
andrewballantyne, line 14:
```suggestion
  test('should have all console loggers except info on prod', () => {
```

packages/lib-core/src/utils/__tests__/logger.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    const consoleSpy = jest.spyOn(console, 'info');
+    process.env.NODE_ENV = 'production';
+    expect(consoleLogger.warn).toBeDefined();
+    expect(consoleLogger.error).toBeDefined();
+    expect(consoleLogger.info).toBeDefined();
+    consoleLogger.info('foo');
+    expect(consoleSpy).not.toBeCalled();
```
andrewballantyne, line 21: This is fair, I guess we don't care what the impl is... as long as it doesn't blow up and doesn't call `console.info`

andrewballantyne:
I wonder if we are failing because we ran out of memory... everything passes locally, but this is in the CI logs:

```
ERRO[2022-02-15T13:04:42Z] Some steps failed:                           
ERRO[2022-02-15T13:04:42Z] 
  * could not run steps: step all-packages failed: test "all-packages" failed: the pod ci-op-tpl674jq/all-packages failed after 1m47s (failed containers: test): ContainerFailed one or more containers exited
Container test exited with code 129, reason OOMKilled
```

andrewballantyne:
@karelhala @vojtechszocs  thoughts on the test failure?

/hold

We might need to increase the PROW memory usage?

karelhala:
@andrewballantyne huh, after adding your suggested change, the test passed. We might still want to increase the mem size though.

andrewballantyne (approved):
Well, let us not make any changes until it's needed... if we don't have an OOM issue, we shouldn't try to fix it.

andrewballantyne:
@vojtechszocs I'll unhold this tomorrow, please review before it merges (if you want)

config/globalSetup.ts
vojtechszocs: @andrewballantyne This allows `ts-node` to hook into the Node `require` machinery, so when a module is loaded, it can transform it via TS compiler.

config/setupTests.ts
vojtechszocs: @karelhala This addresses the issue when testing a component that calls the `React.useLayoutEffect` hook, right?

package.json
vojtechszocs: No Babel please, we already transform via TypeScript :smiley:

package.json
vojtechszocs: Currently we're doing that kind of dependency & config isolation for ESLint stuff via `eslint-plugin-internal` monorepo package.

We should try to have root `package.json` without any dependencies; when a specific monorepo package needs e.g. Jest it should have an explicit dependency in its own package manifest. This aligns with general Yarn 2+ best practices (`yarn run <tool>` generally expects `tool` to be a direct dependency of the given package, for example).

In future, we can create e.g. `testing-internal` monorepo package and have `lib-core` and any other packages to depend on that, while keeping the root `package.json` clean of any dependencies.

.gitignore
```diff
@@ -12,3 +12,4 @@
 
 # Generated project artifacts
 **/dist
+__coverage__
```
vojtechszocs, line 15: Do we actually generate code coverage? Also, can we not use the ugly Python-like `__something__` convention?

config/setupTests.ts
andrewballantyne: That is correct @vojtechszocs

config/globalSetup.ts
```diff
@@ -0,0 +1,4 @@
+export default async () => {
+  process.env.TZ = 'UTC';
```
vojtechszocs, line 2: Just wondering, why is this required?

vojtechszocs: Also, in our `.editorconfig` we have `insert_final_newline = true` but extra trailing whitespace shouldn't be allowed.

If you're using VS Code, I'd suggest using the following extensions:
```
christian-kohler.npm-intellisense
EditorConfig.EditorConfig
esbenp.prettier-vscode
```

Paired with `.vscode/settings.json` this should auto-format common source file types on save.

config/setupTests.ts
vojtechszocs: @andrewballantyne I'm fine keeping this bit, just want to make sure the provided override (`React.useEffect`) is fully compatible with what we're replacing (`React.useLayoutEffect`) in terms of both signature and behavior.

config/setupTests.ts
```diff
@@ -0,0 +1,12 @@
...
+  };
+});
+Element.prototype.scrollTo = () => {};
```
vojtechszocs, line 12: I feel like this kind of code should be part of the React testing library itself, but if we needed, let's keep it.

jest.config.ts
```diff
@@ -0,0 +1,42 @@
+import type {Config} from '@jest/types';
```
vojtechszocs, line 1: Nit: there should be spaces around `Config` e.g. `{ Config }` - I think this is due to your VS Code not auto-formatting on save. Please check my above comment.

jest.config.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const config: Config.InitialOptions = {
+  verbose: true,
+  preset: 'ts-jest',
+  coverageDirectory: '__coverage__',
```
vojtechszocs, line 6: Aha! Can we use a better name for the coverage directory?

vojtechszocs: Also, do we actually need code coverage report at this point? (We can always add it later. Note that all "lib" packages are still under initial development.)

jest.config.ts
```diff
@@ -0,0 +1,42 @@
...
+  collectCoverage: true,
+  transformIgnorePatterns: [
+    'node_modules/(?!@patternfly|lodash-es|@console|@novnc|@spice-project|@popperjs|i18next)'
```
vojtechszocs, line 9: I guess this was copied from Console, we can remove some stuff from here.

jest.config.ts
```diff
@@ -0,0 +1,42 @@
...
+    'text',
+    'text-summary'
+  ],
+  collectCoverageFrom: [
+    'packages/*/src/**/*.{js,jsx,ts,tsx}',
+    '!**/node_modules/**',
+    '!**/dist/**'
+  ],
+  testURL: 'http://localhost',
+  testRegex: '.*\\.(spec|test)\\.(ts|tsx|js|jsx)$',
```
vojtechszocs, line 23: I'd suggest using a single way to name tests, i.e. just `spec` or `test` but not `spec|test`.

jest.config.ts
```diff
@@ -0,0 +1,42 @@
...
+  testEnvironment: 'jsdom',
+  testPathIgnorePatterns: [
+    '<rootDir>/node_modules/'
+  ],
+  roots: ['<rootDir>/packages/'],
+  globalSetup: '<rootDir>/config/globalSetup.ts',
+  setupFilesAfterEnv: ['<rootDir>/config/setupTests.ts'],
+  moduleNameMapper: {
+    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga|css|less|scss)$': 'identity-obj-proxy',
```
vojtechszocs, line 32: I guess this can also be shortened a bit.. `mp3` or `wav` etc. shouldn't be needed :smile:

jest.config.ts
```diff
@@ -0,0 +1,42 @@
...
+    'lodash-es': 'lodash'
```
vojtechszocs, line 33: I don't think we need this. We aim to use `lodash-es` (only) for all of our packages.

package.json
```diff
@@ -7,6 +7,17 @@
   "scripts": {
     "build": "yarn workspaces foreach run build",
     "lint": "yarn workspaces foreach run lint",
-    "test": "yarn workspaces foreach run test"
+    "test": "jest"
```
vojtechszocs, line 10: I'd keep this and let every specific package to provide its own `test` script, which would execute `jest` CLI scoped to that package's directory.

packages/common/package.json
```diff
@@ -6,7 +6,7 @@
   "main": "src/index.ts",
   "scripts": {
     "lint": "yarn lint:run --ext .ts,.tsx src rollup-configs.js",
-    "test": "echo 'Write some tests!'"
+    "test": "yarn --cwd ../.. test packages/common"
```
vojtechszocs, line 9: This is the Console way of running tests, with all common dependencies crammed into root `package.json` file. This is not what Yarn 2+ recommends.

I'd change this to `jest .` or similar.

Check the `lint` script above, it follows the same pattern. Root package has `"lint": "yarn workspaces foreach run lint"` and specific packages run ESLint on files/directories within that package.

packages/common/tsconfig-bases/all-targets.json
```diff
@@ -1,5 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
+  "exclude": ["**/*.test.*", "**/__mocks__/*", "**/__tests__/*"],
```
vojtechszocs, line 3: For now I'd exclude what actually exists and needs to be excluded.

Is `__mocks__` directory configurable in Jest? (I remember they had an issue about it.)

Also, I'd personally prefer avoiding `__tests__` directories and instead putting test files right next to actual code. (Test code is as important as production code, not just an after-thought.)

Wondering what others think :thinking:

packages/lib-core/src/utils/__tests__/logger.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  });
+
+  test('should have all console loggers', () => {
+    expect(consoleLogger.warn).toBeDefined();
+    expect(consoleLogger.error).toBeDefined();
+    expect(consoleLogger.info).toBeDefined();
+  });
+
+  test('should have all console loggers except info on prod', () => {
+    const consoleSpy = jest.spyOn(console, 'info');
```
vojtechszocs, line 15: This can be done via `beforeEach` so `consoleSpy` variable is accessible to all test cases (each `test` or `it` instance).

vojtechszocs: Generally, code that creates a Jest spy/mock should be close to code that resets/teardowns it.

vojtechszocs (changes requested):

.gitignore
andrewballantyne: This is the default configuration for Jest... not sure if you can change it 🤔

config/setupTests.ts
andrewballantyne: It's noted as such on the React Docs

> useLayoutEffect... The signature is identical to useEffect...
https://reactjs.org/docs/hooks-reference.html#uselayouteffect

jest.config.ts
andrewballantyne: Again, think we need prettier if we want to do this. Or a linter that errors out on the PR (ideally both)

jest.config.ts
andrewballantyne: Agreed, we should align... what does everyone think? 

I say `test` is always straight forward. `spec` always felt like "oh yeah, and that's the test"

package.json
andrewballantyne: `yarn workspaces foreach run test` does make sense now that he set it all up again in each.

packages/common/package.json
karelhala: Unfortunately that won't work `jest` would have to be installed in each package and the config would have to be shared (that's OK-ish), however collecting codecov would be nightmare (each package would have its own coverage report). I've tried this in a couple of monorepos and unless there's been a big change in jest this is the way.

There's an option to have package just for the jest config and run it similarly how `lint:run` is done. I could do that, but that brings extra complexity to the play. I'd rather go with the standard "One jest in root and run it from there". 

Should we want to have more granular config (one for react and the other one for eslint for instance) we can use the `projects` field https://github.com/facebook/jest/issues/5420#issuecomment-369714348.

packages/common/tsconfig-bases/all-targets.json
karelhala: I am fine by adding tests directly next to the files, we should however list all ignore patterns here, otherwise when we'll run the build command these tests could be accidentally added to the build folder.

packages/lib-core/src/utils/__tests__/logger.spec.ts
karelhala: Not sure I follow, the spy is watched just in this test so I wouldn't abuse the `beforeEach` for that. However good point of checking the spy in the test before this one.

package.json
karelhala: Unfortunately that won't work, see https://github.com/openshift/dynamic-plugin-sdk/pull/31#discussion_r808949063

jest.config.ts
karelhala: In all of built packages the `lodash-es` is used, however since jest doesn't understand new ES modules we have to provide a fallback to `lodash` we would either have to pull in babel config, or use this moduleMapper.

.gitignore
karelhala: We are generating code coverage, and we can change it. What would be the prefered way?

config/globalSetup.ts
karelhala: Yeah, I am all down for prettier!

jest.config.ts
karelhala: I'd say let's collect coverage from the start. It doesn't add anything to tests and we can see the number go up.

jest.config.ts
karelhala: Yup, I think we can remove the `@console`, but I'd rather keep the rest in here. It's always a struggle for new devs to find why some new tests are failing because they are testing for instance a component that imports `@patternfly`.

jest.config.ts
karelhala: I think the `spec` comes from the history, honestly I'd rather keep both. Otherwise someone who is used to wrigin `.spec.ts` would fail to run the new tests.

jest.config.ts
karelhala: Yeah, many of these are not required, but I'd rather keep them here otherwise it could bite us in future.

config/globalSetup.ts
karelhala: I've also noticed that `.ts` files are not properly linted (no vs code error), it just throws an error when running `yarn lint`

packages/common/tsconfig-bases/all-targets.json
andrewballantyne: Agreed `__tests__` is not my desired pattern either 🙂

packages/lib-core/src/utils/__tests__/logger.spec.ts
andrewballantyne: Hmmm... I think I was thinking of a mock impl not a spy 🤔 

I can't find anything that backs up needing to reset a spy when it's in a scoped situation like inside a test.

The scenario I was thinking of was:
```
      // Suppress the console log for a cleaner test
      jest.spyOn(console, 'log').mockImplementation(jest.fn());

     . . . // some test stuff that calls console.log

      // re-enable logs for future tests
      // eslint-disable-next-line no-console
      (console.log as any).mockRestore();
```

packages/common/package.json
karelhala: If you find something let me know, I've been looking for the non-root approach of jest testig for quite some time and couldn't find anything. There's the `packages` config in jest we could use, but that still didn't allow running jest from within each package. After all the jest is installed only in root, we'd have to install it to every package and that is hell to maintain.

config/globalSetup.ts
andrewballantyne: Interesting 🤔 

My IDE (WebStorm) shows a linter error when I removed the ignore in the logger file:
image

It's only a warning in our linter... so that's why it's highlighted in this yellow shade. Otherwise it would have a red underline for an error. 

My guess is it is a configuration for VSCode? Maybe @vojtechszocs can share his configuration?

vojtechszocs: I think the issue here is we generally run ESLint on each package's sources, and this PR introduces a source directory outside the typical monorepo structure.

I'm actually in favor of moving Jest infra related code into a dedicated monorepo package, but it can be done as a second step.

jest.config.ts
vojtechszocs: +1 for just `test`, AFAIK `spec` term originates from Ruby world (but I can be wrong).

jest.config.ts
vojtechszocs: We're already un-ignoring `lodash-es` from `ts-node` processing, so Jest should be able to process `lodash-es` imports just fine (?)

packages/common/package.json
vojtechszocs: > `jest` would have to be installed in each package

This is the standard approach for monorepos using Yarn 2+

vojtechszocs: Each package is a standalone module, what's conceptually wrong with each one having its own code coverage report?

packages/common/tsconfig-bases/all-targets.json
vojtechszocs: If you mean Rollup build for "lib" packages, that always starts at the configured index file, specified in `rollup.config.js` files. Unless that code directly or indirectly references test code, there's no chance test code would be picked up by Rollup bundler.

vojtechszocs: Also, only "lib" packages have `build` script defined; the rest are private/internal helper packages that don't have a build output.

packages/lib-core/src/utils/__tests__/logger.spec.ts
vojtechszocs: What I meant is you can move `const consoleSpy = jest.spyOn(console, 'info');` into the outer scope.

It would be created just once, and reset/intitialized before or after each test case.

packages/common/package.json
karelhala: If we'd were to add jest as dev dependency of each package that would be nightmare to keep up to date (been there, done that). Codecov monitors cpde coverage per repository and not per package. After all it tracks a coverage ot code.

We would still have to add config for each package, again extra work on our side to keep everything in order.

And the last thing is that jest takes a couple seconds to start for each instance.

As I said, I've tried having jest configs per package and it didn't work. If you want we can give it another try, but I would mark that as Spike and not stall this PR anymore

karelhala:
@vojtechszocs I had to remove `type: module` from both core and lib packages, because jest runs in Node and ES modules are not supported. There's another way how to tell NPM "this package uses ES modules" and that's by using `module` field. It's similiar to `main`. In fact `main` points to `CJS` module `module` points to `ES` module [1]. It's a proposal, but webpack already works with it [2].

We should consider building two modules out of our SDK, I've added a topic for out next Arch sync and I can showcase how we've enabled it in other projects.

[1] https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md
[2] https://webpack.js.org/guides/author-libraries/#final-steps

PR #33, opened 2022-02-25
Add lib-webpack package

vojtechszocs:
Fixes [HAC-491](https://issues.redhat.com/browse/HAC-491)

- add `lib-webpack`
  - to be distributed as `@openshift/dynamic-plugin-sdk-webpack`
  - has peer dependency on core package `@openshift/dynamic-plugin-sdk`

- update `lib-core`
  - name of the global callback function is fixed to `__load_plugin_entry__`
  - expose support APIs (types and code) for use in `lib-webpack`

- update `lib-*`
  - move utility-like dependencies from `peerDependencies` & `devDependencies` to `dependencies`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,3 +11,64 @@
 ```
 npm install -g yarn
 ```
+
+## Distributable Packages
+
+Following packages are distributed via [npm registry](https://www.npmjs.com/):
+
+| Package Name | Description |
+| ------------ | ----------- |
+| `@openshift/dynamic-plugin-sdk` | Allows loading, managing and interpreting plugins. |
+| `@openshift/dynamic-plugin-sdk-utils` | Provides Kubernetes and React utilities. |
+| `@openshift/dynamic-plugin-sdk-webpack` | Allows building plugin assets with webpack. |
+
+Each package is versioned and published independently from other packages.
+
+## Publishing Packages
+
+### Update npm
+
+Make sure the `npm` package manager that comes with Node.js is up-to-date.
+
+To see the current installed version:
+
+```sh
+npm ls -g npm
+```
+
+To check if the current installed version is outdated:
+
+```sh
+npm outdated -g npm
+```
+
+To install the latest version:
+
+```sh
+npm install -g npm@<version>
+```
+
+### Log into npmjs account
+
+Only members of npmjs [openshift organization](https://www.npmjs.com/org/openshift) can publish
+packages maintained in this repo.
+
+```sh
+npm login
+```
+
+### Publish a package
+
+To publish the given package using the current version as specified in its `package.json`:
+
+```sh
+npm publish packages/lib-<pkg> --no-git-tag-version
+```
+
+To publish a package under a different version and update its `package.json` accordingly:
+
+```sh
+npm publish packages/lib-<pkg> --no-git-tag-version --new-version <version>
+```
+
+It's recommended to add `--dry-run` to `npm publish` command prior to actual publish.
--- a/packages/lib-core/src/index.ts
+++ b/packages/lib-core/src/index.ts
@@ -12,10 +12,10 @@ export { usePluginInfo } from './store/usePluginInfo';
 export { useResolvedExtensions } from './store/useResolvedExtensions';
 
 // Support APIs, to be consumed by other plugin SDK packages
-export { PluginRuntimeMetadata, PluginRuntimeManifest } from './types/plugin';
+export { PLUGIN_MANIFEST, REMOTE_ENTRY_SCRIPT, REMOTE_ENTRY_CALLBACK } from './constants';
+export { PluginRuntimeMetadata, PluginManifest } from './types/plugin';
 export {
   extensionArraySchema,
   pluginRuntimeMetadataSchema,
-  pluginRuntimeManifestSchema,
+  pluginManifestSchema,
 } from './yup-schemas';
-export { pluginManifestFile, remoteEntryScript, remoteEntryCallback } from './store/PluginLoader';
--- a/packages/lib-core/src/index.ts
+++ b/packages/lib-core/src/index.ts
@@ -1,3 +1,5 @@
+// Main APIs, to be consumed by plugins and host applications
+export { Extension } from './types/extension';
 export { PluginLoader, PluginLoaderOptions } from './store/PluginLoader';
 export { PluginStore, PluginStoreOptions } from './store/PluginStore';
 export {
@@ -8,3 +10,12 @@ export {
 export { useExtensions } from './store/useExtensions';
 export { usePluginInfo } from './store/usePluginInfo';
 export { useResolvedExtensions } from './store/useResolvedExtensions';
+
+// Support APIs, to be consumed by other plugin SDK packages
+export { PluginRuntimeMetadata, PluginRuntimeManifest } from './types/plugin';
+export {
+  extensionArraySchema,
+  pluginRuntimeMetadataSchema,
+  pluginRuntimeManifestSchema,
+} from './yup-schemas';
+export { pluginManifestFile, remoteEntryScript, remoteEntryCallback } from './store/PluginLoader';
--- a/packages/lib-core/src/store/PluginLoader.ts
+++ b/packages/lib-core/src/store/PluginLoader.ts
@@ -1,22 +1,23 @@
 import type { AnyObject, ResourceFetch } from '@monorepo/common';
 import { consoleLogger } from '@monorepo/common';
 import * as _ from 'lodash-es';
-import type { PluginRuntimeManifest } from '../types/plugin';
+import { PLUGIN_MANIFEST, REMOTE_ENTRY_SCRIPT, REMOTE_ENTRY_CALLBACK } from '../constants';
+import type { PluginManifest } from '../types/plugin';
 import type { PluginEntryModule, PluginEntryCallback } from '../types/runtime';
 import { basicFetch } from '../utils/basic-fetch';
 import { resolveURL } from '../utils/url';
-import { pluginRuntimeManifestSchema } from '../yup-schemas';
+import { pluginManifestSchema } from '../yup-schemas';
 
 type PluginLoadData = {
   entryCallbackFired: boolean;
   status: 'pending' | 'loaded' | 'failed';
-  manifest: PluginRuntimeManifest;
+  manifest: PluginManifest;
 };
 
 type PluginLoadResult =
   | {
       success: true;
-      manifest: PluginRuntimeManifest;
+      manifest: PluginManifest;
       entryModule: PluginEntryModule;
     }
   | {
@@ -24,10 +25,6 @@ type PluginLoadResult =
     };
 
 type PluginLoadListener = (pluginName: string, result: PluginLoadResult) => void;
-
-export const pluginManifestFile = 'plugin-manifest.json';
-export const remoteEntryScript = 'plugin-entry.js';
-export const remoteEntryCallback = '__load_plugin_entry__';
 
 export type PluginLoaderOptions = Partial<{
   /** Control which plugins can be loaded. */
@@ -37,7 +34,7 @@ export type PluginLoaderOptions = Partial<{
   /** Get shared scope object for initializing `PluginEntryModule` containers. */
   getSharedScope: () => AnyObject;
   /** Post-process the plugin manifest. Can be used as a custom validation hook. */
-  postProcessManifest: (manifest: PluginRuntimeManifest) => Promise<PluginRuntimeManifest>;
+  postProcessManifest: (manifest: PluginManifest) => Promise<PluginManifest>;
 }>;
 
 /**
@@ -93,14 +90,14 @@ export class PluginLoader {
    * Fetch the manifest from a plugin's `baseURL` and validate it.
    */
   async getPluginManifest(baseURL: string) {
-    const manifestURL = resolveURL(baseURL, pluginManifestFile);
+    const manifestURL = resolveURL(baseURL, PLUGIN_MANIFEST);
 
     consoleLogger.info(`Loading plugin manifest from ${manifestURL}`);
 
     const response = await this.options.fetchImpl(manifestURL);
     const responseText = await response.text();
 
-    let manifest: PluginRuntimeManifest = await pluginRuntimeManifestSchema
+    let manifest: PluginManifest = await pluginManifestSchema
       .strict(true)
       .validate(JSON.parse(responseText));
 
@@ -114,11 +111,11 @@ export class PluginLoader {
    */
   loadPluginEntryScript(
     baseURL: string,
-    manifest: PluginRuntimeManifest,
+    manifest: PluginManifest,
     getDocument: () => typeof document = _.constant(document),
   ) {
     const pluginName = manifest.name;
-    const scriptURL = resolveURL(baseURL, remoteEntryScript);
+    const scriptURL = resolveURL(baseURL, REMOTE_ENTRY_SCRIPT);
 
     if (this.entryCallback === undefined) {
       throw new Error(`Attempt to load plugin ${pluginName} before registering entry callback`);
@@ -205,7 +202,7 @@ export class PluginLoader {
    */
   registerPluginEntryCallback(getWindow: () => typeof window = _.constant(window)) {
     const windowGlobal = getWindow() as unknown as AnyObject;
-    const callbackName = remoteEntryCallback;
+    const callbackName = REMOTE_ENTRY_CALLBACK;
 
     if (this.entryCallback !== undefined) {
       throw new Error(`Global function ${callbackName} is already registered by this loader`);
--- a/packages/lib-core/src/store/PluginLoader.ts
+++ b/packages/lib-core/src/store/PluginLoader.ts
@@ -1,22 +1,22 @@
 import type { AnyObject, ResourceFetch } from '@monorepo/common';
 import { consoleLogger } from '@monorepo/common';
 import * as _ from 'lodash-es';
-import { pluginManifestSchema } from '../schema/plugin-manifest';
-import type { PluginManifest } from '../types/plugin';
+import type { PluginRuntimeManifest } from '../types/plugin';
 import type { PluginEntryModule, PluginEntryCallback } from '../types/runtime';
 import { basicFetch } from '../utils/basic-fetch';
 import { resolveURL } from '../utils/url';
+import { pluginRuntimeManifestSchema } from '../yup-schemas';
 
 type PluginLoadData = {
   entryCallbackFired: boolean;
   status: 'pending' | 'loaded' | 'failed';
-  manifest: PluginManifest;
+  manifest: PluginRuntimeManifest;
 };
 
 type PluginLoadResult =
   | {
       success: true;
-      manifest: PluginManifest;
+      manifest: PluginRuntimeManifest;
       entryModule: PluginEntryModule;
     }
   | {
@@ -25,17 +25,19 @@ type PluginLoadResult =
 
 type PluginLoadListener = (pluginName: string, result: PluginLoadResult) => void;
 
+export const pluginManifestFile = 'plugin-manifest.json';
+export const remoteEntryScript = 'plugin-entry.js';
+export const remoteEntryCallback = '__load_plugin_entry__';
+
 export type PluginLoaderOptions = Partial<{
   /** Control which plugins can be loaded. */
   canLoadPlugin: (pluginName: string) => boolean;
-  /** Name of the global function used by plugin entry scripts. */
-  entryCallbackName: string;
   /** Custom resource fetch implementation. */
   fetchImpl: ResourceFetch;
   /** Get shared scope object for initializing `PluginEntryModule` containers. */
   getSharedScope: () => AnyObject;
   /** Post-process the plugin manifest. Can be used as a custom validation hook. */
-  postProcessManifest: (manifest: PluginManifest) => Promise<PluginManifest>;
+  postProcessManifest: (manifest: PluginRuntimeManifest) => Promise<PluginRuntimeManifest>;
 }>;
 
 /**
@@ -56,7 +58,6 @@ export class PluginLoader {
   constructor(options: PluginLoaderOptions = {}) {
     this.options = {
       canLoadPlugin: options.canLoadPlugin ?? (() => true),
-      entryCallbackName: options.entryCallbackName ?? '__plugin_entry_callback__',
       fetchImpl: options.fetchImpl ?? basicFetch,
       getSharedScope: options.getSharedScope ?? _.constant({}),
       postProcessManifest: options.postProcessManifest ?? (async (manifest) => manifest),
@@ -92,15 +93,15 @@ export class PluginLoader {
    * Fetch the manifest from a plugin's `baseURL` and validate it.
    */
   async getPluginManifest(baseURL: string) {
-    const manifestURL = resolveURL(baseURL, 'plugin-manifest.json');
+    const manifestURL = resolveURL(baseURL, pluginManifestFile);
 
     consoleLogger.info(`Loading plugin manifest from ${manifestURL}`);
 
     const response = await this.options.fetchImpl(manifestURL);
     const responseText = await response.text();
 
-    let manifest: PluginManifest = await pluginManifestSchema
-      .strict()
+    let manifest: PluginRuntimeManifest = await pluginRuntimeManifestSchema
+      .strict(true)
       .validate(JSON.parse(responseText));
 
     manifest = await this.options.postProcessManifest(manifest);
@@ -113,11 +114,11 @@ export class PluginLoader {
    */
   loadPluginEntryScript(
     baseURL: string,
-    manifest: PluginManifest,
+    manifest: PluginRuntimeManifest,
     getDocument: () => typeof document = _.constant(document),
   ) {
     const pluginName = manifest.name;
-    const scriptURL = resolveURL(baseURL, 'plugin-entry.js');
+    const scriptURL = resolveURL(baseURL, remoteEntryScript);
 
     if (this.entryCallback === undefined) {
       throw new Error(`Attempt to load plugin ${pluginName} before registering entry callback`);
@@ -204,7 +205,7 @@ export class PluginLoader {
    */
   registerPluginEntryCallback(getWindow: () => typeof window = _.constant(window)) {
     const windowGlobal = getWindow() as unknown as AnyObject;
-    const callbackName = this.options.entryCallbackName;
+    const callbackName = remoteEntryCallback;
 
     if (this.entryCallback !== undefined) {
       throw new Error(`Global function ${callbackName} is already registered by this loader`);
--- a/packages/lib-core/src/types/plugin.ts
+++ b/packages/lib-core/src/types/plugin.ts
@@ -5,7 +5,7 @@ export type PluginRuntimeMetadata = {
   version: string;
 };
 
-export type PluginRuntimeManifest = PluginRuntimeMetadata & {
+export type PluginManifest = PluginRuntimeMetadata & {
   extensions: Extension[];
 };
 
--- a/packages/lib-core/src/types/plugin.ts
+++ b/packages/lib-core/src/types/plugin.ts
@@ -1,16 +1,16 @@
 import type { Extension, LoadedExtension } from './extension';
 
-export type PluginMetadata = {
+export type PluginRuntimeMetadata = {
   name: string;
   version: string;
 };
 
-export type PluginManifest = PluginMetadata & {
+export type PluginRuntimeManifest = PluginRuntimeMetadata & {
   extensions: Extension[];
 };
 
 export type LoadedPlugin = {
-  metadata: Readonly<PluginMetadata>;
+  metadata: Readonly<PluginRuntimeMetadata>;
   extensions: Readonly<LoadedExtension[]>;
   enabled: boolean;
 };
--- a/packages/lib-webpack/.eslintrc.js
+++ b/packages/lib-webpack/.eslintrc.js
@@ -1,0 +1,4 @@
+module.exports = {
+  root: true,
+  extends: ['plugin:@monorepo/eslint-plugin-internal/node-typescript-prettier'],
+};
--- a/packages/lib-webpack/package.json
+++ b/packages/lib-webpack/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@openshift/dynamic-plugin-sdk-webpack",
   "version": "1.0.0-alpha1",
-  "description": "Allows building plugins with webpack",
+  "description": "Allows building plugin assets with webpack",
   "license": "Apache-2.0",
   "repository": {
     "type": "git",
--- a/packages/lib-webpack/package.json
+++ b/packages/lib-webpack/package.json
@@ -1,12 +1,12 @@
 {
-  "name": "@openshift/dynamic-plugin-sdk",
+  "name": "@openshift/dynamic-plugin-sdk-webpack",
   "version": "1.0.0-alpha1",
-  "description": "Allows loading, managing and interpreting plugins",
+  "description": "Allows building plugins with webpack",
   "license": "Apache-2.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/openshift/dynamic-plugin-sdk.git",
-    "directory": "packages/lib-core"
+    "directory": "packages/lib-webpack"
   },
   "files": [
     "dist/index.js",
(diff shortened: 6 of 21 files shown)


packages/lib-core/src/index.ts
```diff
@@ -8,3 +10,12 @@ export {
 export { useExtensions } from './store/useExtensions';
 export { usePluginInfo } from './store/usePluginInfo';
 export { useResolvedExtensions } from './store/useResolvedExtensions';
+
+// Support APIs, to be consumed by other plugin SDK packages
```
vojtechszocs, line 14: @andrewballantyne Proposed to have a separate file to export these support APIs (to keep the main API surface as small as possible), we can do it via follow-up change.

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -25,17 +25,19 @@ type PluginLoadResult =
...
 
+export const pluginManifestFile = 'plugin-manifest.json';
+export const remoteEntryScript = 'plugin-entry.js';
+export const remoteEntryCallback = '__load_plugin_entry__';
+
 export type PluginLoaderOptions = Partial<{
   /** Control which plugins can be loaded. */
   canLoadPlugin: (pluginName: string) => boolean;
-  /** Name of the global function used by plugin entry scripts. */
-  entryCallbackName: string;
```
vojtechszocs, line 32: I don't think we need this to be configurable (at least for now), which is why I've removed it.

packages/lib-core/src/types/plugin.ts
```diff
@@ -1,16 +1,16 @@
 import type { Extension, LoadedExtension } from './extension';
 
-export type PluginMetadata = {
+export type PluginRuntimeMetadata = {
   name: string;
   version: string;
 };
 
-export type PluginManifest = PluginMetadata & {
+export type PluginRuntimeManifest = PluginRuntimeMetadata & {
```
vojtechszocs, line 8: Note to self, maybe we can keep this as `PluginManifest`, can be changed later.

packages/lib-webpack/.eslintrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  root: true,
+  extends: ['plugin:@monorepo/eslint-plugin-internal/node-typescript-prettier'],
```
vojtechszocs, line 3: FYI - we have some issues with our `node-typescript-prettier` preset, which is why the `lint` script in `lib-webpack` is currently disabled.

I'll look into this and fix it via follow-up change.

packages/lib-webpack/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "webpack": "^5.69.1"
+  },
+  "dependencies": {
+    "glob": "^7.2.0",
+    "lodash-es": "^4.17.21",
+    "yup": "^0.32.11"
+  },
+  "devDependencies": {
+    "@monorepo/common": "0.0.0-fixed",
+    "@openshift/dynamic-plugin-sdk": "link:../lib-core",
```
vojtechszocs, line 35: Once we publish the core SDK package to npmjs, we switch to a semver pattern here.

In general, each `lib-*` package is meant to be versioned and published separately.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+   * Modules shared between the host application and its plugins at runtime.
+   *
+   * Essential modules like `react` and `redux` will be added automatically.
+   *
+   * Default value: empty object.
+   */
+  sharedModules: WebpackSharedModules;
+}>;
+
+export class DynamicRemotePlugin implements webpack.WebpackPluginInstance {
```
vojtechszocs, line 89: I'm planning to add READMEs (for all packages) and document this webpack plugin via follow-up change.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+    const logger = compiler.getInfrastructureLogger(DynamicRemotePlugin.name);
+    const containerName = this.pluginMetadata.name;
+
+    if (!compiler.options.output.publicPath) {
+      logger.warn(
+        'output.publicPath is not defined, which may cause plugin assets to not load properly in the browser',
+      );
+    }
+
+    if (!compiler.options.output.uniqueName) {
```
vojtechszocs, line 134: @andrewballantyne If a plugin explicitly declares `output.uniqueName` then we assume it has the right value.

What we could do here is compare it to `pluginMetadata.name` and fail if the values aren't equal.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+ */
+type WebpackSharedModules = { [moduleRequest: string]: WebpackSharedModuleConfig };
+
+export type DynamicRemotePluginOptions = Partial<{
+  /**
+   * Plugin metadata JSON file name, or the parsed plugin metadata.
+   *
+   * Default value: `plugin.json`.
+   */
+  pluginMetadata: string | PluginBuildMetadata;
```
vojtechszocs, line 67: Based on @psav's suggestion, we can have two (optional but mutually exclusive) options:
```
pluginMetadata: PluginBuildMetadata;
pluginMetadataFile: string;
```

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * List of extensions contributed by the plugin.
+   *
+   * The value is either a `minimatch` compatible JSON file glob pattern,
+   * or the parsed extensions array.
+   *
+   * Default value: `extensions.json`.
+   */
+  extensions: string | Extension[];
```
vojtechszocs, line 77: Similar to above, this could be expanded into two options:
```
extensions: Extension[];
extensionsFiles: string;
```

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    // Generate webpack federated module container assets
+    new webpack.container.ModuleFederationPlugin({
+      name: containerName,
+      library: {
+        type: 'jsonp',
+        name: remoteEntryCallback,
+      },
+      filename: remoteEntryScript,
```
karelhala, line 146: I would allow passing custom `filename` as well. For instance what if we want to pass `hash` or something else from the app side of things?

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -25,17 +25,19 @@ type PluginLoadResult =
 
 type PluginLoadListener = (pluginName: string, result: PluginLoadResult) => void;
 
+export const pluginManifestFile = 'plugin-manifest.json';
+export const remoteEntryScript = 'plugin-entry.js';
```
karelhala, line 29: We found that serving static assets from CDN is often unreliable and it serves them constantly from cache. For such a fast paced release that is really unwanted as you might depend on the newest version. We've found out that the name is better to be generated with date and `fullhash`
```JS
const remoteEntryScript = (useFileHash) => `plugin-entry${useFileHash ? `.${Date.now()}.[fullhash]` : ''}.js`
```

However this feels out of place, so it might be better to change this to `remoteEntryScript = 'plugin-entry'` and later add the `.js` or fullhash.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
karelhala: As mentioned above, the `filename` should include `hash` as well.
```suggestion
      filename: options.filename ?? `${remoteEntryScript}${options.useFileHash ? `.${Date.now()}.[fullhash]` : ''}.js`,
```

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+      exposes: _.mapValues(
+        this.pluginMetadata.exposedModules || {},
+        (moduleRequest, moduleName) => ({
+          import: moduleRequest,
+          name: `exposed-${moduleName}`,
```
karelhala, line 151: I would allow ignoring this `name`, webpack by default in production mode exposes these modules as hash numbers, that's much better for CDN to dump the cache if such file is stored and requested.

packages/lib-webpack/src/webpack/GenerateManifestPlugin.ts
```diff
@@ -0,0 +1,24 @@
+import { pluginManifestFile } from '@openshift/dynamic-plugin-sdk';
+import type { PluginRuntimeManifest } from '@openshift/dynamic-plugin-sdk';
+import * as webpack from 'webpack';
+
+export class GenerateManifestPlugin implements webpack.WebpackPluginInstance {
+  constructor(private readonly manifest: PluginRuntimeManifest) {}
```
karelhala, line 6: I would allow passing custom `pluginManifestFile` name.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    // Generate plugin manifest
+    new GenerateManifestPlugin({
+      name: this.pluginMetadata.name,
+      version: this.pluginMetadata.version,
+      extensions: this.extensions,
+    }).apply(compiler);
+
+    // Post-process container entry generated by ModuleFederationPlugin
+    new PatchContainerEntryPlugin(this.pluginMetadata.name).apply(compiler);
```
karelhala, line 165: I would add option to bypass this patch, the extension could eventually write their own webpack plugin and sew them together in different way.

packages/lib-core/package.json
```diff
@@ -22,9 +22,11 @@
     "test": "echo 'Write some tests!'"
   },
   "peerDependencies": {
-    "lodash-es": "^4.17.21",
     "react": "^17.0.2",
-    "redux": "^4.1.2",
+    "redux": "^4.1.2"
+  },
+  "dependencies": {
+    "lodash-es": "^4.17.21",
```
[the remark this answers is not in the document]
andrewballantyne: I'm in favour of this being a `dependency` -- the users of lib-core package shouldn't need to bring lodash into their dependencies.

packages/lib-core/src/types/plugin.ts
andrewballantyne: Is there any other type of Manifest than the one we build for us to use at runtime? 🤔

packages/lib-webpack/package.json
andrewballantyne: Is this something we can automate or in some way not miss (a checklist) when publishing? Because I'm pretty sure the npm publish won't resolve this for us, right?

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: We should create a story for this if you have not done so already.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: Not sure why we would allow them to specify if the only value they can give us is equal to the `name`? Wouldn't we just set it to that anyways?

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: In a different story / PR?

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -25,17 +25,19 @@ type PluginLoadResult =
...
+export const remoteEntryCallback = '__load_plugin_entry__';
```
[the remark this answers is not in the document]
andrewballantyne: Agree with @karelhala here -- we should definitely look to make these upper case as they are immutable constants. And placing them in a generic file is my preference as it pollutes the exports of this file with general constants if we leave it this way.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
+import {
+  extensionArraySchema,
+  remoteEntryScript,
+  remoteEntryCallback,
+} from '@openshift/dynamic-plugin-sdk';
+import type { Extension } from '@openshift/dynamic-plugin-sdk';
+import * as _ from 'lodash-es';
```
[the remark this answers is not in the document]
andrewballantyne: Yeah, does rollup have tree-shaking? 🤔

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+const getSharedModulesWithStrictSingletonConfig = (modules: string[]) =>
+  modules.reduce((acc, moduleRequest) => {
+    acc[moduleRequest] = {
+      // Enforce a single version of the shared module to be used at runtime
+      singleton: true,
+      // Prevent plugins from using a fallback version of the shared module
+      import: false,
+    };
+    return acc;
+  }, {} as WebpackSharedModules);
```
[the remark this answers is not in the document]
andrewballantyne: I do like this in some fashion -- it's a bit more knowledge needed to read this cleanly but I think this is a good functional way to do this. 👍

packages/lib-core/src/index.ts
andrewballantyne: We should create a ticket for this.

packages/lib-webpack/.eslintrc.js
andrewballantyne: This needs a ticket

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+import type { PluginBuildMetadata } from '../types/plugin';
+import { pluginBuildMetadataSchema } from '../yup-schemas';
+import { GenerateManifestPlugin } from './GenerateManifestPlugin';
+import { PatchContainerEntryPlugin } from './PatchContainerEntryPlugin';
+
+const parseJSONFile = (filePath: string) => JSON.parse(fs.readFileSync(filePath, 'utf-8'));
+
+const loadPluginMetadata = (fileName: string, baseDir = process.cwd()) => {
+  const filePath = path.resolve(baseDir, fileName);
+  return validatePluginMetadata(parseJSONFile(filePath));
```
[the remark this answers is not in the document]
andrewballantyne: I agree... we want to validate it in both cases, doing a load then validate in one is not ideal imo.

```
// calling code
if ('string')
  // load file and validate contents
else
  // validate contents
```
Why not just combine them... if the contents are invalid we can handle that error together rather than separate (if later desired). Loading should just load.
```
// calling code
if ('string') {
  // load file
}
// validate contents
```
It's the same content, just two different delivery mechanisms... for maintenance I think we should handle those two together.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+};
+
+const loadExtensions = (globPattern: string, baseDir = process.cwd()) => {
+  const filePaths = glob.sync(globPattern, { cwd: baseDir, absolute: true, nodir: true });
+  return _.flatMap(filePaths.map((filePath) => validateExtensions(parseJSONFile(filePath))));
```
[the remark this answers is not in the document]
andrewballantyne: Same stance on file loading vs validation here for me too.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+};
+
+const validatePluginMetadata = (obj: unknown): PluginBuildMetadata =>
+  pluginBuildMetadataSchema.strict(true).validateSync(obj);
+
+const validateExtensions = (obj: unknown): Extension[] =>
+  extensionArraySchema.strict(true).validateSync(obj);
```
andrewballantyne, line 33: I do not believe I'd expect a validate function to return the data. I'm asking "is this valid" not "and give me back the data if it is". 🤔 

I get you're also using this as a typeguard, but it seems odd to me.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+}>;
+
+export class DynamicRemotePlugin implements webpack.WebpackPluginInstance {
+  private readonly pluginMetadata: PluginBuildMetadata;
+
+  private readonly extensions: Extension[];
+
+  private readonly sharedModules: WebpackSharedModules;
+
+  constructor(options: DynamicRemotePluginOptions = {}) {
```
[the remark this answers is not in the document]
andrewballantyne: Hmm, this sounds like a good feature for extending options but not needed first pass imo. We should create a story for this.

packages/lib-webpack/src/webpack/GenerateManifestPlugin.ts
andrewballantyne, line 6: Do you not need to call `super()`?

andrewballantyne: I think that's in the `manifest` 🤔 

```
declare type PluginRuntimeMetadata = {
    name: string;
    version: string;
};
declare type PluginRuntimeManifest = PluginRuntimeMetadata & {
    extensions: Extension[];
};
```

packages/lib-webpack/src/webpack/PatchContainerEntryPlugin.ts
```diff
@@ -0,0 +1,38 @@
+import { remoteEntryScript, remoteEntryCallback } from '@openshift/dynamic-plugin-sdk';
+import * as webpack from 'webpack';
+
+export class PatchContainerEntryPlugin implements webpack.WebpackPluginInstance {
+  constructor(private readonly pluginName: string) {}
```
andrewballantyne, line 5: No need to call `super()` here either?

I'm getting a lint error on both of these for "useless constructor" but that's not true because of the property creation by TS, maybe our linter rules need to be updated?

andrewballantyne (changes requested):

packages/lib-core/src/index.ts
andrewballantyne: Created: https://issues.redhat.com/browse/HAC-760

packages/lib-webpack/.eslintrc.js
andrewballantyne: Ticket: https://issues.redhat.com/browse/HAC-761

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: Added one for the export items of this package: https://issues.redhat.com/browse/HAC-762
Added one for the READMEs on each package: https://issues.redhat.com/browse/HAC-763 (I feel this is important to do separate as it has a lot of text to it likely)

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: Ticket: https://issues.redhat.com/browse/HAC-764

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: Ticket: https://issues.redhat.com/browse/HAC-764

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
andrewballantyne: Created a story here: https://issues.redhat.com/browse/HAC-766 -- @karelhala feel free to expand if you want.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: Future feature ticket: https://issues.redhat.com/browse/HAC-767

packages/lib-core/src/types/plugin.ts
vojtechszocs: No, there shouldn't be any other type of manifest.

The build-generated vs. runtime manifest should have the same structure.

I'll revert this back to `PluginManifest`.

packages/lib-webpack/package.json
vojtechszocs: @andrewballantyne `devDependencies` part of a package manifest is irrelevant when publishing that package to the npm registry. Even if your app or plugin's dependency has `devDependencies` in its package manifest, it's not taken into account.

Check my other comment https://github.com/openshift/dynamic-plugin-sdk/pull/28#discussion_r816931623

vojtechszocs: This is basically us saying "when building `lib-webpack` we want to compile using the local `lib-core` code present in the monorepo".

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
vojtechszocs: @andrewballantyne Technically, `output.uniqueName` doesn't have to equal `pluginMetadata.name` for the plugin to work at runtime. `output.uniqueName` is used for scoping module chunks containing the given plugin's exposed modules, which are loaded by host application at runtime.

We cannot prevent plugins to not specify webpack `output.uniqueName` options :smile: if we want to have it strictly equal to `pluginMetadata.name` then we can fail the build early.

packages/lib-core/package.json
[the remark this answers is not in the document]
vojtechszocs: Spoke with @karelhala about this earlier, for utility-like dependencies which meet both of these conditions

- does not use module-level state (which would cause two different instances to use two different states)
- does not have to be exactly the same version vs. what the actual plugin or host application uses at runtime

it's fine to put them into `dependencies` and have them pulled into the consuming project's `node_modules` structure.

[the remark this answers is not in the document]
vojtechszocs: > the users of lib-core package shouldn't need to bring lodash into their dependencies

Right. Also, it's completely fine if users would use a different version of Lodash if they wanted to.

packages/lib-core/src/store/PluginLoader.ts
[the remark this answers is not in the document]
vojtechszocs: OK, will move into a separate file and export with uppercase naming.

packages/lib-webpack/package.json
andrewballantyne: Doh, yes of course. Sorry about that. I didn't see devDependencies lol.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: That's fair haha... I was thinking more why would we allow them to ever use it... flat out be annoying and throw an error if we can only have one value and it's already set somewhere else 😛

packages/lib-webpack/rollup.config.js
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line import/no-relative-packages
+import { tsLibConfig, dtsLibConfig } from '../common/rollup-configs';
+import pkg from './package.json';
+
+// eslint-disable-next-line no-console
+console.log(`Rollup ${pkg.name}`);
```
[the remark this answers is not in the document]
vojtechszocs: Yes, I'd keep it :smiley: Rollup itself doesn't print the context directory.

I think this is important in case of a monorepo, where you execute Rollup multiple times on different packages.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
vojtechszocs: `lodash-es` is already code that is ESmodules syntax friendly.

All external module imports are reflected into `dist/index.js` as-is. It's the module bundler employed by the consuming plugin or host application that will have to process our SDK code (which is currently expressed as ESmodules).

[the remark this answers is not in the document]
vojtechszocs: Also, sometimes a namespace import (`import * as foo`) is useful for scoping the imported stuff into a single object.

I agree that it's generally best to pick specific imports from the given library, instead of using a namespace import.

For now, we've been sticking with a namespace import style for vendor modules, let's improve it as a follow-up.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
vojtechszocs: If something fails, the error thrown by `JSON.parse` or Yup schema validator should be informative enough.

This code is executed in Node.js runtime as part of webpack compilation.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
vojtechszocs: > it's a bit more knowledge needed to read this cleanly

Code readability is very important.

`Object.fromEntries` takes an iterable and expects each element to be a tuple `[key, value]`, returning an object with properties constructed from these elements. It's basically a reverse operation to `Object.entries`.

Honestly, I don't think this is much better, compared to plain and simple `modules.reduce` array-to-object transformation.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
vojtechszocs: Agreed, it's better to separate load vs. validation logic, will do the change.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
vojtechszocs: Sounds good, will try to use the new `flatMap` API here.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -0,0 +1,167 @@
...
+    const adaptedOptions: Required<DynamicRemotePluginOptions> = {
+      pluginMetadata: options.pluginMetadata ?? 'plugin.json',
```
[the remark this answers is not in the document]
vojtechszocs: This relates to string vs. object load / validation concern mentioned above, will do the changes according to our earlier discussion.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
andrewballantyne: Pure functional programming has always taken a leap, mentally, for me. Reducer is a mixture of imperative and functional. Personally I think the functional programming way is more elegant and straight to the point. But I don't think this is needed for this PR / ever if you don't want. Preference.

I spent a couple years in `clojurescript` so pure functional things look cleaner to me than otherwise... but I'm definitely okay with the `reducer` way of things.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
vojtechszocs: Adding hash is a good idea, will try to do this change.

packages/lib-core/src/store/PluginLoader.ts
vojtechszocs: Yup, dynamically generated entry script name brings some more complexity but I see the argument for it being a `.js` file and potentially cached forever by a CDN.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
vojtechszocs: Yeah, webpack's default chunk naming strategy for production is deterministic IDs :arrow_right: numbers which are unique to the given compilation.

Adding `exposed-` prefix doesn't add or subtract uniqueness of the chunk IDs, while it makes the file names more meaningful for humans.

packages/lib-webpack/src/webpack/GenerateManifestPlugin.ts
vojtechszocs: @andrewballantyne This is about customizing the name of plugin manifest asset, which is currently hard-coded to `plugin-manifest.json`.

I don't see a use case for that right now, but I'm not against it :smiley:

vojtechszocs: No need for `super()` since we are not extending a base class.

We're just implementing an interface `WebpackPluginInstance` here.

packages/lib-webpack/src/webpack/PatchContainerEntryPlugin.ts
vojtechszocs: The ESLint "useless constructor" means constructor with empty body.

This is **not** adequate for TypeScript constructors which use short-hand syntax to initialize class properties.

Above code is short-hand equivalent to
```ts
export class PatchContainerEntryPlugin implements webpack.WebpackPluginInstance {
  private readonly pluginName;

  constructor(pluginName: string) {
    this.pluginName = pluginName;
  }

  // .. rest of code
}
```

vojtechszocs: Reference: https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

packages/lib-webpack/src/webpack/GenerateManifestPlugin.ts
andrewballantyne: DOH, `implements` not `extends` .. gosh I gotta pay more attention

packages/lib-webpack/src/webpack/PatchContainerEntryPlugin.ts
andrewballantyne: No I understand that... but my IDE is showing me the lint errors you have just not allowed to be run by the npm script. We'll need to fix it in the lint issue https://issues.redhat.com/browse/HAC-761

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
vojtechszocs: Agreed, will update code to fail if `output.uniqueName` is specified explicitly, since this is something our webpack plugin will set automatically.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
vojtechszocs: @karelhala Well, there are two concerns here, logical and performance.

For each of these concerns, the question is, does it make sense to return a new accumulator object for each processed array element?

I agree that the object spread based form is shorter and nicer, but can be less performant and I don't see any real benefit from not reusing the same accumulator object from logical perspective :smiley:

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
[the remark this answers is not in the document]
vojtechszocs: @karelhala FYI, the standard `Array.prototype.flatMap` API expects an iteratee callback as its first argument, while `_.flatMap` doesn't have this expectation.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

For now, I'll keep the `_.flatMap` and we can improve it later.

TL;DR `arr.flatMap(_.identity)` should be equal to `_.flatMap(arr)` where `_.identity` is `(x) => x`

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
vojtechszocs: @andrewballantyne I'd suggest to do this in a follow-up to get this PR merged soon.

Basically webpack `ModuleFederationPlugin` should support placeholders like `[hash]` for the `filename` option, but I haven't tested this yet and don't want to hold merging this PR because of this.

Overall, it's a good idea to add hashes to any `.js` assets associated with the plugin. If we do this, we'll also need to pass the exact file name (with the hash) via `plugin-manifest.json` so that the host application knows its name.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
vojtechszocs: I've improved the code to separate the parse vs. validate concerns better.

README.md
```diff
@@ -11,3 +11,64 @@
...
+
+| Package Name | Description |
+| ------------ | ----------- |
+| `@openshift/dynamic-plugin-sdk` | Allows loading, managing and interpreting plugins. |
+| `@openshift/dynamic-plugin-sdk-utils` | Provides Kubernetes and React utilities. |
+| `@openshift/dynamic-plugin-sdk-webpack` | Allows building plugin assets with webpack. |
+
+Each package is versioned and published independently from other packages.
+
+## Publishing Packages
```
andrewballantyne, line 27: That's this ticket: https://issues.redhat.com/browse/HAC-768 -- I guess you can merge it with this stuff...

andrewballantyne (approved):
LGTM

README.md
vojtechszocs: Thanks! Updated PR description to mention this ticket.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
andrewballantyne: Ticket to track the work... https://issues.redhat.com/browse/HAC-771 please add any additional details or rework if I misrepresented it.

PR #34, opened 2022-02-28
Create a table component

fhlavac:
[HAC-753](https://issues.redhat.com/browse/HAC-753)

Table:
table1

Loading:
loading

Default error:
error

Data:
```
const columns = [
    {
      title: 'Name',
      id: 'name',
      sort: 'metadata.name',
      transforms: [sortable], 
      props: {
        className: '',
      },
    },
    {
      title: 'Namespace',
      id: 'namespace',
      props: {
        className: '',
      },
    },
    {
      title: 'Status',
      id: 'status',
      transforms: [null],
      props: {
        className: '',
      },
    },
    {
      title: 'Detail',
      id: 'detail',
      transforms: [sortable],
      props: {
        className: '',
      },
    },
    {
      title: 'Note',
      id: 'note',
      transforms: [null],
      props: {
        className: '',
      },
    },
  ];

const data = [
  { name: 'one', namespace: 'two', status: 'three', detail: 'four', note: 'five' },
  { name: 'one - 2', namespace: null, status: null, detail: 'four - 2', note: 'five - 2' },
  {
    name: 'one - 3',
    namespace: 'two - 3',
    status: 'three - 3',
    detail: 'four - 3',
    note: 'five - 3',
  },
];

export const Row: React.FC<RowProps<RowObject, Record<string, unknown>>> = ({ obj }) => (
  <>
    <Td dataLabel={obj.name}>
      {obj.name}
    </Td>
    <Td dataLabel={obj.namespace}>
      {obj.namespace}
    </Td>
    <Td dataLabel={obj.status}>
      {obj.status}
    </Td>
    <Td dataLabel={obj.detail}>
      {obj.detail}
    </Td>
    <Td dataLabel={obj.note}>
      {obj.note}
    </Td>
  </>
);
```

Component usage:
```
<SdkTable
  columns={columns}
  data={data}
  loaded={false}
  // emptyLabel={'Something went wrong...'}
  // loadError={'An error occured when fetching the data'}
  Row={Row}
/>
```

### DIFF
--- a/packages/lib-utils/package.json
+++ b/packages/lib-utils/package.json
@@ -22,6 +22,8 @@
   },
   "peerDependencies": {
     "@openshift/dynamic-plugin-sdk": "^1.0.0",
+    "@patternfly/react-core": "^4.198.19",
+    "@patternfly/react-table": "^4.67.19",
     "react": "^17.0.2",
     "react-redux": "^7.2.2",
     "redux": "^4.1.2",
--- a/packages/lib-utils/src/components/status/StatusBox.tsx
+++ b/packages/lib-utils/src/components/status/StatusBox.tsx
@@ -1,0 +1,99 @@
+import { Alert, EmptyState, EmptyStateBody, Spinner, Text, Title } from '@patternfly/react-core';
+import * as _ from 'lodash-es';
+import * as React from 'react';
+
+export type LoadError = {
+  message: string;
+  status: number;
+};
+
+type AccessDeniedProps = {
+  message: string | React.ComponentType;
+};
+
+export const AccessDenied: React.FC<AccessDeniedProps> = ({
+  message = 'Insufficient access rights.',
+}) => (
+  <EmptyState>
+    <Title headingLevel="h2">Restricted Access</Title>
+    <EmptyStateBody>
+      <Text>You don&apos;t have access to this section due to cluster policy.</Text>
+      {message && (
+        <Alert isInline variant="danger" title="Error details">
+          {message}
+        </Alert>
+      )}
+    </EmptyStateBody>
+  </EmptyState>
+);
+
+export type StatusBoxProps = {
+  areFiltersApplied?: boolean;
+  noData: boolean;
+  children?: React.ReactElement;
+  emptyLabel?: string;
+  loadError?: LoadError;
+  loaded?: boolean;
+  LoadErrorDefaultMsg?: React.ComponentType;
+  NoDataEmptyMsg?: React.ComponentType;
+  EmptyMsg?: React.ComponentType;
+};
+
+export const StatusBox: React.FC<StatusBoxProps> = ({
+  areFiltersApplied,
+  noData,
+  children = null,
+  loadError,
+  emptyLabel,
+  loaded,
+  EmptyMsg,
+  LoadErrorDefaultMsg,
+  NoDataEmptyMsg,
+}): React.ReactElement | null => {
+  if (loadError) {
+    const status = _.get(loadError, 'response.status');
+    const loadErrorMsg = loadError.message || LoadErrorDefaultMsg || 'Data loading failed.';
+    switch (status) {
+      case 404:
+        return (
+          <EmptyState>
+            <Title headingLevel="h1">404: Not Found</Title>
+          </EmptyState>
+        );
+      case 403:
+        return <AccessDenied message={loadErrorMsg} />;
+      default:
+        break;
+    }
+
+    return (
+      <EmptyState>
+        <Title headingLevel="h1">{emptyLabel}</Title>
+        <EmptyStateBody>{loadErrorMsg}</EmptyStateBody>
+      </EmptyState>
+    );
+  }
+
+  if (!loaded) {
+    return (
+      <EmptyState>
+        <Spinner />
+      </EmptyState>
+    );
+  }
+
+  if (noData) {
+    if (!areFiltersApplied && NoDataEmptyMsg) {
+      return <NoDataEmptyMsg />;
+    }
+    return EmptyMsg ? (
+      <EmptyMsg />
+    ) : (
+      <EmptyState>
+        <Text>{emptyLabel}</Text>
+      </EmptyState>
+    );
+  }
+
+  return children;
+};
--- a/packages/lib-utils/src/components/table/Table.tsx
+++ b/packages/lib-utils/src/components/table/Table.tsx
@@ -1,0 +1,123 @@
+import type { AnyObject } from '@monorepo/common';
+import type { ICell, ThProps } from '@patternfly/react-table';
+import { TableComposable, Tbody, Th, Thead, Tr } from '@patternfly/react-table';
+import * as _ from 'lodash-es';
+import * as React from 'react';
+import type { LoadError } from '../status/StatusBox';
+import { StatusBox } from '../status/StatusBox';
+
+export type RowProps<D> = {
+  obj: D;
+};
+
+export type TableColumn = ICell & {
+  title: string;
+  id: string;
+  sort?: ThProps['sort'] | string;
+};
+
+export type TableProps<D> = {
+  areFiltersApplied?: boolean;
+  data: D[];
+  loaded: boolean;
+  loadError?: LoadError;
+  columns: TableColumn[];
+  Row: React.FC<RowProps<D>>;
+  LoadErrorDefaultMsg?: React.ComponentType;
+  NoDataEmptyMsg?: React.ComponentType;
+  EmptyMsg?: React.ComponentType;
+  emptyLabel?: string;
+  'aria-label'?: string;
+};
+
+const Table: React.FC<TableProps<AnyObject>> = ({
+  areFiltersApplied,
+  data: initialData,
+  loaded,
+  loadError,
+  columns,
+  Row,
+  LoadErrorDefaultMsg,
+  NoDataEmptyMsg,
+  EmptyMsg,
+  emptyLabel,
+  'aria-label': ariaLabel,
+}) => {
+  const [data, setData] = React.useState(initialData);
+  const [activeSortIndex, setActiveSortIndex] = React.useState(-1);
+  const [activeSortDirection, setActiveSortDirection] = React.useState('none');
+
+  React.useEffect(() => {
+    setData(initialData);
+  }, [initialData]);
+
+  const onSort = (event: React.FormEvent, index: number, direction: string) => {
+    setActiveSortIndex(index);
+    setActiveSortDirection(direction);
+    // back compatibility with sort column attribute defined as a string + transforms: [sortable]
+    const columnSort = _.isString(columns[index].sort) ? columns[index].sort : undefined;
+    const updatedRows = data.sort((objA, objB) => {
+      const a = columnSort ? _.get(objA, String(columnSort)) : Object.values(objA)[index];
+      const b = columnSort ? _.get(objB, String(columnSort)) : Object.values(objB)[index];
+      if (typeof a === 'number' && typeof b === 'number') {
+        return direction === 'asc' ? a - b : b - a;
+      }
+      return direction === 'asc'
+        ? String(a).localeCompare(String(b))
+        : String(b).localeCompare(String(a));
+    });
+    setData(updatedRows);
+  };
+
+  return (
+    <StatusBox
+      noData={!data || _.isEmpty(data)}
+      loaded={loaded}
+      loadError={loadError}
+      areFiltersApplied={areFiltersApplied}
+      emptyLabel={emptyLabel}
+      LoadErrorDefaultMsg={LoadErrorDefaultMsg}
+      NoDataEmptyMsg={NoDataEmptyMsg}
+      EmptyMsg={EmptyMsg}
+    >
+      <TableComposable aria-label={ariaLabel}>
+        <Thead>
+          <Tr>
+            {columns.map(({ title, props: properties, sort, transforms }, columnIndex) => {
+              const isSortable = !!transforms?.find((item) => item?.name === 'sortable');
+              const defaultSort = {
+                sortBy: {
+                  index: activeSortIndex,
+                  direction: activeSortDirection,
+                },
+                onSort,
+                columnIndex,
+              };
+              return (
+                <Th
+                  // eslint-disable-next-line react/no-array-index-key
+                  key={`column-${columnIndex}`}
+                  sort={isSortable ? defaultSort : sort}
+                  // eslint-disable-next-line react/jsx-props-no-spreading
+                  {...properties}
+                >
+                  {title}
+                </Th>
+              );
+            })}
+          </Tr>
+        </Thead>
+        <Tbody>
+          {data.map((item, idx) => (
+            // eslint-disable-next-line react/no-array-index-key
+            <Tr key={`row-${idx}`}>
+              <Row obj={item} />
+            </Tr>
+          ))}
+        </Tbody>
+      </TableComposable>
+    </StatusBox>
+  );
+};
+
+export default Table;
--- a/package.json
+++ b/package.json
@@ -13,6 +13,8 @@
     "eslint-print-config": "yarn eslint --print-config"
   },
   "devDependencies": {
+    "@patternfly/react-core": "^4.198.19",
+    "@patternfly/react-table": "^4.67.19",
     "@rollup/plugin-commonjs": "^21.0.2",
     "@rollup/plugin-node-resolve": "^13.1.1",
     "@rollup/plugin-typescript": "^8.3.0",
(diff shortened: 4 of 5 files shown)


packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,72 @@
+// @ts-nocheck
```
florkbr, line 1: What issues did we run into with typescript here? I don't think we could merge this without TS checking.

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,72 @@
...
+import * as React from 'react';
+import { Alert, Bullseye, Spinner, Text, Title } from '@patternfly/react-core';
+import * as _ from 'lodash';
+
+type AccessDeniedProps = {
+  message?: string;
```
florkbr, line 7: Wondering if we should have a fallback message if none are provided? Or we could hide the error details section if no message is provided. I'm thinking it would look odd to see `Error details` in the UI and underneath is blank :)

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,72 @@
...
+  EmptyMsg?: React.ComponentType;
+  children?: React.ReactElement;
+};
+
+export const StatusBox: React.FC<StatusBoxProps> = (props) => {
+  const { data, EmptyMsg, loadError, loaded, children, NoDataEmptyMsg, unfilteredData } = props;
+
+  if (loadError) {
+    const status = _.get(loadError, 'response.status');
+    if (status === 404) {
```
florkbr, line 39: we could probably do a `switch` statement on the `status` so it's easier to add/catch new cases in the future.

packages/lib-utils/src/components/table/Table.tsx
```diff
@@ -0,0 +1,95 @@
+// @ts-nocheck
```
florkbr, line 1: Same here, should re-enable TS when we can.

packages/lib-utils/src/components/table/Table.tsx
```diff
@@ -0,0 +1,95 @@
...
+  OnSelect,
+  SortByDirection,
+  TableComposable,
+  TableGridBreakpoint,
+  Tbody,
+  // Td,
+  Th,
+  Thead,
+  Tr,
+} from '@patternfly/react-table';
```
florkbr, line 14: Did we try starting with https://www.patternfly.org/v4/components/table/ first and tweak it to what we need? I'm worried we're rebuilding another table implementation using the patternfly building blocks. That might be necessary in the end, but we should start simple.

 I think the idea is we want a `Table` component as a middle man between SDK consumers and the Patternfly table that can sprinkle some K8s handling magic inbetween.

packages/lib-utils/src/components/status/StatusBox.tsx
fhlavac: This code is so far just a draft. I'll remove this. It's just for dev purposes to prevent it from causing build errors.

packages/lib-utils/src/components/status/StatusBox.tsx
fhlavac: Absolutely, I'll add it.

packages/lib-utils/src/components/table/Table.tsx
fhlavac: Not sure I'm getting your point. Do you mean using `<Table>` PF component instead of `<TableComposable>` with the build blocks? I wanted to avoid using `<Table>` since it's marked as legacy - see https://www.patternfly.org/v4/components/table/react-legacy

fhlavac:
@florkbr thanks for the review. Sorry for the small errors and TS ignores. I should have warned you :laughing: 
I wanted more of a review of the concept and features we need urgently than code details. I'll go through it a lot of times again and clean it. Just wanted to be clear about what is a priority for us and if this is ok from the concept point of view.

packages/lib-utils/src/components/table/Table.tsx
florkbr: Ah, I completely missed that! Good point. I didn't see the `TableComposable` import

packages/lib-utils/src/components/table/Table.tsx
```diff
@@ -0,0 +1,95 @@
...
+  additional?: boolean;
+  // ??! sorting TO DO
+  sort?: ((data: D[], sortDirection: SortByDirection) => D[]) | string; // map to the new sort
+};
+
+type TableProps<D, R extends any = {}> = {
+  data: D[];
+  unfilteredData: D[];
+  loaded: boolean;
+  loadError: any;
```
florkbr, line 35: Since we have a chance to revisit our table component API, we should try avoiding `any` if possible

fhlavac:
Some more questions:


**1) Ad StatusBox** - status component wrapping the table
- I added just very simple statuses, that comply with the virtualized table status types. How much do we want to bring from the console's version in this implementation from the UX point of view? 
- Do we want to use some kind of table skeleton or the empty state is currently enough?
- `IncompleteDataError`, `TimeoutError` do we want to migrate them from console/dynamic-plugin-sdk? For now commented the code sections that should use them.


**2) Ad types**

@vojtechszocs disabled `react/jsx-props-no-spreading`, `'react/require-default-props`. Is it ok or do we want to have any of these enabled?


**3) Ad table**
- do we want to use `rowData` attribute?
- do we want to use `activeColumnIds` attribute now?
- not sure what `additional` is in `TableColumn` type
- for table props - do we want to deal with `scrollNode`, `gridBreakPoint`
- do we want to utilize `onSelect` - selecting items?

fhlavac:
@florkbr pushed updates according to our sync on Tuesday. Cleaned the code, modified component properties, and added a possibility to use custom sorting.

**In follow up PRs, I think we should:**
- add filtering
- add row selecting
- handle translations
- use rowRenderer - `rowData` property
- manage columns being displayed
- handle scrolling and pagination

packages/eslint-plugin-internal/src/rules/react-typescript-prettier.js
```diff
@@ -10,4 +10,10 @@ module.exports = {
       unnamedComponents: 'arrow-function',
     },
   ],
+
+  // Allow props spreading
+  'react/jsx-props-no-spreading': 'off',
+
+  // Disable default props require
+  'react/require-default-props': 'off',
```
florkbr, line 18: @fhlavac I really think we should keep these rules enabled if possible. 

`jsx-props-no-spreading` - while I like this for easy component development, we need to think of our components as an API, and tend to agree with the eslint doc:
> This enhances readability of code by being more explicit about what props are received by the component. It is also good for maintainability by avoiding passing unintentional extra props and allowing react to emit warnings when invalid HTML props are passed to HTML elements

`require-default-props` - I think this forces us to consider which props are required and which are not WRT our component API.

> One advantage of defaultProps over custom default logic in your code is that defaultProps are resolved by React before the PropTypes typechecking happens, so typechecking will also apply to your defaultProps.

packages/lib-utils/package.json
```diff
@@ -24,6 +24,8 @@
   },
   "peerDependencies": {
     "@openshift/dynamic-plugin-sdk": "^1.0.0",
+    "@patternfly/react-core": "^4.198.19",
+    "@patternfly/react-table": "^4.67.19",
```
florkbr, line 28: @fhlavac let's leave this as is for now. I'll need to think on this a bit more. Technically I believe we can have multiple different versions of a patternfly component running on a page - therefore we may not need to enforce the table component as a peer dependency.

packages/lib-utils/src/components/table/Table.tsx
```diff
@@ -0,0 +1,126 @@
...
+  obj: D;
+};
+
+export type TableColumn<D> = ICell & {
+  title: string;
+  id: string;
+  sort?: ThProps['sort'] | ((data: D[], sortDirection: SortByDirection) => D[]) | string;
+};
+
+type TableProps<D> = {
```
florkbr, line 18: @fhlavac we should also export the table props object to be used when creating an instance of the table.

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+      {message && (
+        <Alert isInline variant="danger" title="Error details">
+          {message}
+        </Alert>
+      )}
+    </EmptyStateBody>
+  </EmptyState>
+);
+
+type StatusBoxProps = {
```
florkbr, line 30: We should export the type object for status box props too

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+  areFiltersApplied?: boolean;
+  data?: unknown;
```
florkbr, line 32: @fhlavac FWIW `unknown` should be avoided. Can we use a shape?

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+type AccessDeniedProps = {
+  message: string | React.ComponentType;
+};
+
+export const AccessDenied: React.FC<AccessDeniedProps> = ({
+  message = 'Insufficient access rights.',
+}) => (
+  <EmptyState>
+    <Title headingLevel="h2">Restricted Access</Title>
```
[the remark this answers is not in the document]
fhlavac: @florkbr created one https://issues.redhat.com/browse/HAC-958

packages/lib-utils/src/components/status/StatusBox.tsx
fhlavac: @florkbr I will probably use `Record<string, unknown>`. Is that ok?

andrewballantyne: @fhlavac @florkbr I like to make a callback to my point around `unknown` I [made on slack](https://coreos.slack.com/archives/C02EQERSV3N/p1646162496757559). I still believe this is true... types need to be Generic long before we consider `unknown` valid. `unknown` is a step up from `any` and as an SDK we will definitely have `unknown`s... but we need to make sure this is not something that can be properly typed. Otherwise the code that receives `data` will be effectively `any`.

fhlavac: @andrewballantyne @florkbr added a generic type to `StatusBoxProps`. In this case, the `data` property is used just to verify if it is empty or not so it's probably enough to expect that it is an array.

packages/eslint-plugin-internal/src/rules/react-typescript-prettier.js
fhlavac: Completely agree with the `require-default-props` just not sure about `jsx-props-no-spreading ` since in https://github.com/openshift/dynamic-plugin-sdk/pull/34/files#diff-b56b27c61ebe92f1fe8626d77d6dcae22ddfbccfd4c1d71f7da9d3c629e615c5R101 we probably want to spread any properties passed. Disabled the rule for that line for now.

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+    <EmptyStateBody>
+      <Text>You don&apos;t have access to this section due to cluster policy.</Text>
```
vojtechszocs, line 20: If it makes sense, I'd suggest a follow-up change where we use props instead of hard-coded strings like the one above.

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+);
+
+export type StatusBoxProps<D> = {
+  areFiltersApplied?: boolean;
+  data?: D[];
+  children?: React.ReactElement;
+  emptyLabel?: string;
+  loadError?: LoadError;
+  loaded?: boolean;
+  LoadErrorDefaultMsg?: React.ComponentType<Record<string, never>>;
```
vojtechszocs, line 37: `Record<string, never>` infers into an empty object, see [here](https://github.com/typescript-eslint/typescript-eslint/issues/2063#issuecomment-675156492) for details.

I'd suggest adding the following into `packages/common/src/types/common.ts` right after `AnyObject` type declaration:
```ts
/**
 * Use the `EmptyObject` type for objects that have no properties.
 * 
 * @see https://github.com/typescript-eslint/typescript-eslint/issues/2063#issuecomment-675156492
 */
export type EmptyObject = Record<string, never>;
```

and using `EmptyObject` wherever it's needed.

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+  NoDataEmptyMsg?: React.ComponentType<Record<string, never>>;
+  EmptyMsg?: React.ComponentType<Record<string, never>>;
+};
+
+export const StatusBox: React.FC<StatusBoxProps<Record<string, unknown>>> = ({
```
vojtechszocs, line 42: You can use the existing `AnyObject` type which is aliased to `Record<string, unknown>`.

```ts
import type { AnyObject } from '@monorepo/common';
```

I'd suggest doing a search & replace where
- `Record<string, unknown>` :arrow_right: `AnyObject`
- `Record<string, never>` :arrow_right: `EmptyObject`

vojtechszocs (changes requested):

packages/lib-utils/src/components/status/StatusBox.tsx
fhlavac: I'll add this to a follow-up list. We probably have to discuss it a bit. In the original console's implementation, there is a translated string being rendered. Since we don't have translations yet, I put just this string for now.

packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  if (!loaded) {
+    return (
+      <EmptyState>
+        <Spinner />
+      </EmptyState>
+    );
+  }
+
+  if (!data || _.isEmpty(data)) {
```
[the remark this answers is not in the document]
fhlavac: then we return `children` - in this case the table. Is that right?

packages/lib-utils/src/components/table/Table.tsx
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export type TableProps<D> = {
+  areFiltersApplied?: boolean;
+  data: D[];
+  loaded: boolean;
+  loadError?: LoadError;
+  columns: TableColumn[];
+  Row: React.FC<RowProps<D>>;
+  LoadErrorDefaultMsg?: React.ComponentType;
```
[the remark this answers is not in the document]
fhlavac: makes sense, thank you :slightly_smiling_face:

packages/lib-utils/src/components/status/StatusBox.tsx
vojtechszocs: @fhlavac @florkbr Well, there are two aspects to this, first one is localization (for which we currently don't have support in the SDK repo yet), while the second one is reusability (hard-coded strings, regardless if localized or not, make a component less reusable).

Ideally, for every component we move from Console to `lib-utils` in the SDK repo, we should re-evaluate its API and code. I agree that initially we want to have some components ready and good-enough to be used by HAC Core, this is more for a long-term component & API related discussion.

packages/lib-utils/src/components/status/StatusBox.tsx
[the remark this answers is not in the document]
vojtechszocs: Well, I was just curious as to why `data` is not used at all once we have it non-empty :smiley: just returning `children` is typically used as fallback behavior, normally we'd use `data` to render something different than `children` pass-through.

So if this is intentional and we want to have it working this way, we can keep it. I'd definitely put it on a follow-up TODO list to understand why this code is written this way, and maybe refactor to make it simpler.

vojtechszocs:
@fhlavac One more change, please add

```json
"@patternfly/react-core": "^4.198.19",
"@patternfly/react-table": "^4.67.19",
```

also to `packages/lib-utils/package.json` into `peerDependencies` to ensure they are treated as proper external dependencies.

packages/lib-utils/src/components/status/StatusBox.tsx
florkbr: @andrewballantyne agreed; I pinned your comment in the forum as well so we can find it again :)

florkbr: @fhlavac for now that's ok. One thing of note - if we're only checking `data` to see if it's empty, we could just add a new prop `empty: boolean;` and do the check in the consumer (that way we don't need to know anything about the data structure here).

florkbr (approved):
Looks good! Thanks @fhlavac. We have more work to do on the table in the future, but this is a good start.

PR #35, opened 2022-02-28
Bump lib package versions

vojtechszocs:
[SemVer spec](https://semver.org/) says

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time.

this is sometimes referred to as "the escape clause" for `0.y.z` versions. (See [this article](https://nodesource.com/blog/semver-tilde-and-caret/) for details.)

This PR bumps all lib packages to `1.0.0-alpha1` since we're close to publishing their first versions.


andrewballantyne (approved):
I agree with the major version not being zero for first release.

PR #36, opened 2022-03-02
Add isUtilsConfigSet

andrewballantyne:
Fixes: https://issues.redhat.com/browse/HAC-770

We found that there is no clean way to handle utils in a React App without the need to check if the config is set. Adding that function.

### DIFF
--- a/packages/lib-utils/src/config.ts
+++ b/packages/lib-utils/src/config.ts
@@ -15,6 +15,13 @@ export type UtilsConfig = {
 let config: Readonly<UtilsConfig> | undefined;
 
 /**
+ * Checks if the {@link UtilsConfig} is set.
+ */
+export const isUtilsConfigSet = (): boolean => {
+  return config !== undefined;
+};
+
+/**
  * Set the {@link UtilsConfig} reference.
  *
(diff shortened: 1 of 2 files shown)


packages/lib-utils/src/config.ts
```diff
@@ -32,7 +39,7 @@ export const setUtilsConfig = (c: UtilsConfig) => {
  *
  * Throws an error if the reference isn't already set.
  */
-export const getUtilsConfig = () => {
+export const getUtilsConfig = (): UtilsConfig => {
   if (config === undefined) {
```
andrewballantyne, line 43: Couldn't update this reference due to the way TypeScript guarantees that undefined was handled earlier in the flow. So I left this check in here not calling the utility I just created. 

If anyone can help me figure out that "type guard" (for the lack of a better way of saying it), let me know. 

[TS Docs on Narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html)

karelhala:
@vojtechszocs can we get a new release out of this?

andrewballantyne:
> @vojtechszocs can we get a new release out of this?

We should do this when he gets to releasing lib-webpack atm (likely tomorrow)

PR #37, opened 2022-03-03
HAC-488: Migrate K8s reducers to SDK

florkbr:
Migrates the K8s reducer to the SDK.

As of now, these are not migrated in this PR:
- [ ] ~tests~
- [x] core reducer
- [ ] ~static models~


packages/lib-utils/src/k8s/ws-factory.ts
```diff
@@ -0,0 +1,332 @@
+import { consoleLogger } from '@monorepo/common';
```
florkbr, line 1: Andrew is working on migrating this code. I just moved it over as a placeholder API.

packages/lib-utils/src/app/redux/actions/k8s/index.ts
```diff
@@ -0,0 +1,335 @@
...
+    if (!_.get(k8sType, 'verbs', ['watch']).includes('watch')) {
+      consoleLogger.warn('Resource does not support watching', k8sType);
+      return _.noop;
+    }
+
+    const { subprotocols } = getImpersonate(getState()) || {};
+
+    // TODO use websocket handler
+    WS[id] = k8sWatch(k8sType, queryWithCluster, { subprotocols });
+    // WS[id] = k8sWatch(k8sType, queryWithCluster, { subprotocols }).onbulkmessage((events: { object: K8sResourceCommon }[]) =>
```
florkbr, line 311: This is where I'll need to tie into @andrewballantyne's websocket work.

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -50,3 +62,189 @@ export const getK8sResourceURL = (
...
+ * @param model k8s model
+ * @return The reference for model i.e `group~version~kind`.
+ * * */
+export const getReferenceForModel = (model: K8sModelCommon): K8sResourceKindReference =>
+  getReference({ apiGroup: model.apiGroup, apiVersion: model.apiVersion, kind: model.kind });
+
+// TODO Migrate implementation or refactor K8s reducer to avoid usage
+let k8sModels: ImmutableMap<K8sResourceKindReference, K8sModelCommon>;
+
+export const allModels = () => {
```
florkbr, line 236: @andrewballantyne @vojtechszocs this logic (not yet migrated) ultimately pulls in the console list of static models. See https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/src/utils/k8s/k8s-utils.ts#L215 and https://github.com/openshift/console/blob/master/frontend/public/models/index.ts

This PR is already pretty large, so I opted to stub it out for now until we review what I have already.

packages/lib-utils/src/types/k8s.ts
```diff
@@ -50,11 +64,64 @@ export type QueryParams = Partial<{
   labelSelector: string;
   fieldSelector: string;
   resourceVersion: string;
-  [key: string]: string;
+  [key: string]: string | number;
```
florkbr, line 67: Number is required for specifiying `limit: paginationLimit,` in the queryparams. See https://github.com/openshift/dynamic-plugin-sdk/pull/37/files#diff-052548bf85cbe16601d8136f8af94f09dfa8c038d68d2affd1d9eb1a8b06e4ebR126

packages/lib-utils/src/k8s/k8s-utils.ts
florkbr: Additionally, the missing logic requires changes to our `Extension` types in the core package.

packages/lib-utils/src/k8s/k8s-resource.ts
```diff
@@ -33,6 +33,7 @@ type K8sResourceListResult<TResource extends K8sResourceCommon> = {
   items: TResource[];
   metadata: {
     resourceVersion: string;
+    continue: string;
```
florkbr, line 36: Required for action logic here: https://github.com/openshift/dynamic-plugin-sdk/pull/37/files#diff-052548bf85cbe16601d8136f8af94f09dfa8c038d68d2affd1d9eb1a8b06e4ebR128

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -50,3 +62,189 @@ export const getK8sResourceURL = (
...
+  return '';
+};
+
+const createEquals = (key: string, value: string): MatchExpression => ({
+  key,
+  operator: 'Equals',
+  values: [value],
+});
+
+const isOldFormat = (selector: Selector | MatchLabels) =>
```
florkbr, line 108: @andrewballantyne @vojtechszocs I can remove this logic, but until we test how this behaves thoroughly I figured I'd leave it.

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -50,3 +62,189 @@ export const getK8sResourceURL = (
...
+ * @deprecated - This will become obsolete when we move away from K8sResourceKindReference to K8sGroupVersionKind
+ * Provides a reference string that uniquely identifies the group, version, and kind of a k8s resource.
+ * @param K8sGroupVersionKind Pass K8sGroupVersionKind which will have group, version, and kind of a k8s resource.
+ * @param K8sGroupVersionKind.apiGroup Pass group of k8s resource or model.
+ * @param K8sGroupVersionKind.apiVersion Pass version of k8s resource or model.
+ * @param K8sGroupVersionKind.kind Pass kind of k8s resource or model.
+ * @return The reference for any k8s resource i.e `group~version~kind`.
+ * If the group will not be present then "core" will be returned as part of the group in reference.
+ * * */
+export const getReference = ({
```
florkbr, line 216: Debating if it's worth stripping the `~` logic out right now... It's adding complexity and I've had to waltz around it. One question I have is how will pushing and retrieving from the `Set<>` here: https://github.com/openshift/dynamic-plugin-sdk/pull/37/files#diff-8e891455963ea25676398e13f0e91cf63350df91315a520dd470da3039fd3c23R121 will handle the `{ apiGroup, apiVersion, kind }` object

packages/lib-utils/src/types/k8s.ts
```diff
@@ -1,4 +1,4 @@
-type K8sResourceIdentifier = {
+export type K8sResourceIdentifier = {
```
florkbr, line 1: I'm using this in the K8s `~` reference logic... Ideally we wouldn't expose this

packages/lib-utils/src/k8s/k8s-utils.ts
andrewballantyne: We should _need_ the models... they are additional details of custom nature (colour for instance)... this is fine to be omitted on first pass. We'll no doubt have to iterate on everything.

packages/lib-utils/src/types/k8s.ts
andrewballantyne: Hmmm... yeah, we may need tech debt story to clean this up

packages/lib-utils/src/app/redux/actions/k8s/index.ts
```diff
@@ -0,0 +1,335 @@
...
+  (
+    id: string,
+    name: string,
+    namespace: string,
+    query: { [key: string]: string },
+    k8sType: K8sModelCommon,
+    partialMetadata = false,
+  ) =>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (dispatch: any, getState: any) => {
```
andrewballantyne, line 255:
```suggestion
  (dispatch: Dispatch, getState: GetState): void => {
```
Maybe we make a `GetState` common type too; we'll definitely have other uses.
```
type GetState = () => SDKStoreState;
```

packages/lib-utils/src/app/redux/reducers/index.ts
```diff
@@ -1,5 +1,5 @@
 import { coreReducer } from './core';
-import { sdkK8sReducer } from './k8s';
+import sdkK8sReducers from './k8s';
```
andrewballantyne, line 2:
```suggestion
import sdkK8sReducer from './k8s';
```

It's only 1 reducer... right? You're assigning it as such on line 12.

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
```diff
@@ -1,15 +1,271 @@
...
 import type { K8sState } from '../../../../types/redux';
+import type { K8sAction } from '../../actions/k8s';
+import { ActionType } from '../../actions/k8s';
+import { getK8sDataById } from './selector';
 
-// TODO remove when actions/reducers are migrated
-const emptyK8sState: K8sState = fromJS({
-  RESOURCES: { inFlight: false, models: ImmutableMap<string, K8sModelCommon>() },
-});
+const getQN: (obj: K8sResourceCommon) => string = (obj: K8sResourceCommon) => {
```
andrewballantyne, line 11:
```suggestion
const getQN: (obj: K8sResourceCommon) => string = (obj) => {
```

Don't need to double type.

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
```diff
@@ -1,15 +1,271 @@
...
+  const { name, namespace } = obj?.metadata || {};
+  // Name + namespace is not unique for PackageManifest resources, so include the catalog source.
+  // TODO: We should be able to remove this when the upstream OLM bug is fixed:
+  //       https://bugzilla.redhat.com/show_bug.cgi?id=1814822
+  if (obj.apiVersion === 'packages.operators.coreos.com/v1' && obj.kind === 'PackageManifest') {
+    return `(${obj.status?.catalogSource})-${name}`;
+  }
```
andrewballantyne, line 18: Delete this use-case? We do not have OLM.

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
```diff
@@ -1,15 +1,271 @@
...
+import type { K8sAction } from '../../actions/k8s';
+import { ActionType } from '../../actions/k8s';
+import { getK8sDataById } from './selector';
 
-// TODO remove when actions/reducers are migrated
-const emptyK8sState: K8sState = fromJS({
-  RESOURCES: { inFlight: false, models: ImmutableMap<string, K8sModelCommon>() },
-});
+const getQN: (obj: K8sResourceCommon) => string = (obj: K8sResourceCommon) => {
+  const { name, namespace } = obj?.metadata || {};
```
andrewballantyne, line 12: `obj` is not optional in your signature... so either `?` isn't needed or we need to worry about `name` being undefined.

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
```diff
@@ -1,15 +1,271 @@
...
+  // TODO: We should be able to remove this when the upstream OLM bug is fixed:
+  //       https://bugzilla.redhat.com/show_bug.cgi?id=1814822
+  if (obj.apiVersion === 'packages.operators.coreos.com/v1' && obj.kind === 'PackageManifest') {
+    return `(${obj.status?.catalogSource})-${name}`;
+  }
+  return (namespace ? `(${namespace})-` : '') + name;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const moreRecent = (a: any, b: any) => {
```
andrewballantyne, line 23: I assume these are some sorta `ImmutableMaps` wrapping maybe `K8sResourceCommon` (because of `metadata` references)?

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
```diff
@@ -1,15 +1,271 @@
...
+  const metaA = a.get('metadata').toJSON();
+  const metaB = b.get('metadata').toJSON();
+  if (metaA.uid !== metaB.uid) {
+    return new Date(metaA.creationTimestamp) > new Date(metaB.creationTimestamp);
+  }
+  return parseInt(metaA.resourceVersion, 10) > parseInt(metaB.resourceVersion, 10);
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const removeFromList = (list: ImmutableMap<string, any>, resource: K8sResourceCommon) => {
```
andrewballantyne, line 33:
```suggestion
const removeFromList = (list: ImmutableMap<string, unknown>, resource: K8sResourceCommon) => {
```

Don't think we care what the value is... unless we get a centralized type for this item, maybe `unknown` is all we should do.

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
```diff
@@ -1,15 +1,271 @@
...
+  const qualifiedName = getQN(resource);
+  consoleLogger.info(`deleting ${qualifiedName}`);
+  return list.delete(qualifiedName);
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const updateList = (list: ImmutableMap<string, any>, nextJS: K8sResourceCommon) => {
```
andrewballantyne, line 40: Boy, I do not understand this signature... `nextJS` is a K8s object but comes back as a ImmutableMap? Which `current` is already? So `list` is a map of maps?

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
```diff
@@ -1,15 +1,271 @@
...
+  ) {
+    // If the only thing that differs is resource version, don't fire an update.
+    return list;
+  }
+
+  return list.set(qualifiedName, next);
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const loadList = (oldList: any, resources: any) => {
```
andrewballantyne, line 68: oldList is probably the ImmutableMap of ImmutableMaps?

Resources is typed on line 72, so I think we can safely do an array here.

```suggestion
const loadList = (oldList: any, resources: K8sResourceCommon[]) => {
```

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
```diff
@@ -1,15 +1,271 @@
...
+        consoleLogger.warn(
+          `${metadata.namespace}-${metadata.name} is gone with no deletion timestamp!`,
+        );
+      }
+      list.delete(k);
+    });
+  });
+};
+
+const sdkK8sReducers = (state: K8sState, action: K8sAction): K8sState => {
```
andrewballantyne, line 95:
```suggestion
const sdkK8sReducer = (state: K8sState, action: K8sAction): K8sState => {
```

andrewballantyne, line 95: This reducer is a mess, and I opt for us not to try and fix that now.

I assume the logic is fairly close to that of OpenShift Console one we migrated?

packages/lib-utils/src/app/redux/reducers/k8s/selector.ts
```diff
@@ -0,0 +1,6 @@
+import type { K8sState, SDKStoreState } from '../../../../types/redux';
+
+export const getReduxIdPayload = (state: SDKStoreState, reduxId: string) =>
+  state?.k8s?.get(reduxId);
```
andrewballantyne, line 4:
```suggestion
  state.k8s.get(reduxId);
```
They are not optional in the call signature, they can't be chained as such.

packages/lib-utils/src/app/redux/reducers/k8s/selector.ts
```diff
@@ -0,0 +1,6 @@
...
+
+export const getK8sDataById = (state: K8sState, id: string) => state?.getIn([id, 'data']);
```
andrewballantyne, line 6:
```suggestion
export const getK8sDataById = (state: K8sState, id: string) => state.getIn([id, 'data']);
```

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -50,3 +62,189 @@ export const getK8sResourceURL = (
...
+  kind: K8sModelCommon,
+  query: {
+    labelSelector?: Selector;
+    resourceVersion?: string;
+    ns?: string;
+    fieldSelector?: string;
+  } = {},
+  wsOptions: {
+    [key: string]: unknown;
+  } = {},
```
andrewballantyne, line 145: This is type-able soon... `Partial<WebSocketOptions>` kinda deal

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -50,3 +62,189 @@ export const getK8sResourceURL = (
...
+) => {
+  const queryParams: QueryParams = { watch: 'true' };
+  const opts: {
+    queryParams: QueryParams;
+    ns?: string;
+  } = { queryParams };
+  const wsOptionsUpdated = {
+    path: '',
+    host: 'auto',
```
andrewballantyne, line 154: won't need this.
```suggestion
```

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -50,3 +62,189 @@ export const getK8sResourceURL = (
...
+  wsOptions: {
+    [key: string]: unknown;
+  } = {},
+) => {
+  const queryParams: QueryParams = { watch: 'true' };
+  const opts: {
+    queryParams: QueryParams;
+    ns?: string;
+  } = { queryParams };
+  const wsOptionsUpdated = {
```
andrewballantyne, line 152: This probably should be typed.
```suggestion
  const wsOptionsUpdated: WebSocketOptions = {
```

packages/lib-utils/src/k8s/k8s-utils.ts
andrewballantyne: Yeah, this is a painful migration for us... we should try to make a stance on it though.

packages/lib-utils/src/types/k8s.ts
```diff
@@ -7,27 +7,41 @@ type K8sResourceIdentifier = {
 export type K8sModelCommon = K8sResourceIdentifier & {
   plural: string;
   propagationPolicy?: 'Foreground' | 'Background';
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  verbs?: any;
```
andrewballantyne, line 11:
```suggestion
  verbs?: K8sVerb[];
```

packages/lib-utils/src/types/k8s.ts
```diff
@@ -7,27 +7,41 @@ type K8sResourceIdentifier = {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  shortNames?: any;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  crd?: any;
```
andrewballantyne, line 15:
```suggestion
  crd?: boolean;
```

packages/lib-utils/src/types/k8s.ts
```diff
@@ -7,27 +7,41 @@ type K8sResourceIdentifier = {
 export type K8sModelCommon = K8sResourceIdentifier & {
   plural: string;
   propagationPolicy?: 'Foreground' | 'Background';
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  verbs?: any;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  shortNames?: any;
```
andrewballantyne, line 13:
```suggestion
  shortNames?: string[];
```

packages/lib-utils/src/types/k8s.ts
```diff
@@ -7,27 +7,41 @@ type K8sResourceIdentifier = {
...
+      ownerReferences: OwnerReference[];
+      resourceVersion: string;
+      uid: string;
+    }>;
+    spec: {
+      selector?: Selector | MatchLabels;
+      [key: string]: unknown;
+    };
+    status: { [key: string]: unknown };
+    data: { [key: string]: unknown };
```
andrewballantyne, line 43: Hmmm... I wonder if we want to split these out to other types if you need to invoke generic actions. If you need SPECIFIC actions on structure, we should look at why those are not more strictly typed.

This is probably not a good change overall though.

packages/lib-utils/src/types/k8s.ts
andrewballantyne: Hmmm, no way to cast eh? I guess we probably want it to be a number for purity reasons 🤔

packages/lib-utils/src/types/k8s.ts
```diff
@@ -50,11 +64,64 @@ export type QueryParams = Partial<{
...
+ */
+export type GroupVersionKind = string;
+
+/**
+ * @deprecated Use GetGroupVersionKindForModel type instead. Support for type K8sResourceKindReference will be removed in a future release.
+ * @see GetGroupVersionKindForModel
+ * The canonical, unique identifier for a Kubernetes resource type.
+ * Maintains backwards-compatibility with references using the `kind` string field.
+ */
+export type K8sResourceKindReference = GroupVersionKind | string;
```
andrewballantyne, line 127: I wish this would not be exported if we can avoid it... our goal is to get rid of it and I don't want to make a release with stuff we want to move away from.

packages/lib-utils/src/types/redux.ts
```diff
@@ -1,7 +1,7 @@
 import type { Map as ImmutableMap } from 'immutable';
 import type { K8sResourceCommon } from './k8s';
 
-export type K8sState = ImmutableMap<string, unknown>;
+export type K8sState = ImmutableMap<string, unknown> | undefined;
```
andrewballantyne, line 4: Reducer state can never been undefined save from the first invocation... I would not type this with `| undefined`... what's the use-case you're trying to solve?

andrewballantyne (changes requested):
Lots of comments, most are to help with the `any` types.

Looking pretty good. We should definitely be careful about functionality changes to the `k8sReducer` though... that's a mess I think we need follow up work for.

packages/lib-utils/src/app/redux/actions/k8s/index.ts
```diff
@@ -0,0 +1,335 @@
...
+    };
+    POLLs[id] = setInterval(poller, 30 * 1000);
+    poller();
+
+    if (!_.get(k8sType, 'verbs', ['watch']).includes('watch')) {
+      consoleLogger.warn('Resource does not support watching', k8sType);
+      return _.noop;
+    }
+
+    const { subprotocols } = getImpersonate(getState()) || {};
```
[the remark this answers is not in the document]
andrewballantyne: A backlog ticket: https://issues.redhat.com/browse/HAC-783

packages/lib-utils/src/app/redux/reducers/k8s/selector.ts
andrewballantyne: This looks to be a problem -- maybe we leave it as-is 🤔

packages/lib-utils/src/app/redux/reducers/k8s/selector.ts
andrewballantyne: Same here, maybe we leave this

packages/lib-utils/src/types/redux.ts
andrewballantyne: Maybe we leave this too -- this map needs to be fixed but maybe not for this PR for us to move forward.

packages/lib-utils/src/app/redux/actions/k8s/index.ts
```diff
@@ -0,0 +1,335 @@
...
+  const ws = WS[id];
+  if (ws) {
+    ws.destroy();
+    delete WS[id];
+  }
+  const poller = POLLs[id];
+  clearInterval(poller);
+  delete POLLs[id];
+  delete REF_COUNTS[id];
+  return dispatch(stopWatchK8s(id));
```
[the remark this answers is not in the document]
florkbr: We have a linter rule that prevents this: `consistent-return`
image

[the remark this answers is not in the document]
florkbr: Also some conflicting calls below to `stopK8sWatch` - we need to look at better typing for thunks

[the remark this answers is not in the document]
andrewballantyne: Thunk problem is here: https://github.com/openshift/dynamic-plugin-sdk/pull/37#discussion_r819868447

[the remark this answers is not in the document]
andrewballantyne: `Consistent-return` is because you have another return statement earlier that is returning a value -- at least I'm fairly sure... I think it's just saying "return always"

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
florkbr: The `Loaded` action case below was not happy when I tried this :( 
image

packages/lib-utils/src/app/redux/reducers/k8s/selector.ts
florkbr: I think since `SDKStoreState` can yield `undefined` (see https://github.com/openshift/dynamic-plugin-sdk/pull/37/files#r819902820), this is required.

packages/lib-utils/src/app/redux/reducers/k8s/index.ts
andrewballantyne: Yeah, I think LoadList has to be probably wrong that it can take a single item 🤔

PR #38, opened 2022-03-03
Add WebSocket factory

andrewballantyne:
Addresses: https://issues.redhat.com/browse/HAC-717

What's new:

- `WebSocketFactory` is a new class to handle the operations around
    - Creating one of these will rely on another global config (like our k8s utilities)
- Added a global config to the `AppInitSDK` and the underlying `UtilsConfig` type
- App should own the sub protocols and host, whereas the the Plugin can provide the path to the resource
- Exposed select types, an enum and the WSFactory out of the `lib-utils` package

Current Issues:

- [x] Fixing the type issues with setTimeout/setInternal in WSFactory

Working example of non-cleaned up code can be found in HAC-Core (https://github.com/openshift/hac-core/pull/34)


packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -13,6 +13,7 @@ type AppInitSDKProps = {
   configurations: {
     apiDiscovery?: InitAPIDiscovery;
     appFetch: UtilsConfig['appFetch'];
+    wsConfigs: UtilsConfig['wsConfigs'];
```
florkbr, line 16: FWIW - I didn't immediately see `wsConfigs` and think `websockets`. Not sure if we could name this a bit better? Just a nit.

packages/lib-utils/src/ws/WSFactory.ts
```diff
@@ -0,0 +1,407 @@
...
+  DestroyHandler,
+  OpenHandler,
+  ErrorHandler,
+} from './types';
+import { applyConfigSubProtocols, applyConfigHost, createURL } from './ws-utils';
+
+/**
+ * The WS configurations.
+ */
+export type WSOptions = {
```
florkbr, line 18: Wondering if we should create a new file in `lib-utils/src/types/ws.ts` for the enum/types in this file? Not blocking.

packages/lib-utils/src/app/AppInitSDK.tsx
andrewballantyne: 🤔 
`ws:` is a web socket protocol... `wss:` is the secure version... just like `http:` and `https:`... not sure I can agree with you on this one Bryan... but I'll think on it. If you have a better name I can go with -- I don't mind changing.

I'd like to avoid `webSocketConfig` though...

packages/lib-utils/src/ws/WSFactory.ts
andrewballantyne: I have a types file... I could put it in there...

packages/lib-utils/src/ws/types.ts
```diff
@@ -0,0 +1,27 @@
...
+export type OpenHandler = GenericHandler<never>;
+export type CloseHandler = GenericHandler<CloseEvent>;
+export type ErrorHandler = GenericHandler<Event>;
+/**
+ * The WebSocket data can be returned in an object state or in the raw string response passed.
+ *
+ * This is configured through `jsonParse` options flag.
+ * @see WSOptions
+ */
+export type MessageDataType = AnyObject | string;
```
florkbr, line 13: See other comment here: https://github.com/openshift/dynamic-plugin-sdk/pull/38/files#r819000771. 

Not sure if we want to push these types to the root types folder?

andrewballantyne: Not sure we need to push all types into the same folder. That seems like a bad way to stay organized as we grow 🤔 

Not sure why we have a `types` folder if not for common stuff.

packages/lib-utils/src/app/AppInitSDK.tsx
florkbr: Docs might be our answer here. Maybe what's bothering me more is we have `Configs` nested in a `configurations` object.

andrewballantyne: An interesting point. Configurations for the AppInitSDK can have configurations for web sockets though -- maybe I'm bias, but I'm not seeing a real issue here. 

I'll call it settings if you want. It's a configuration for the WSFactory... but Settings is more-or-less swappable.

packages/lib-utils/src/ws/types.ts
```diff
@@ -0,0 +1,27 @@
+import type { AnyObject } from '@monorepo/common';
+
+type GenericHandler<T = unknown> = (data: T) => void;
+export type OpenHandler = GenericHandler<never>;
+export type CloseHandler = GenericHandler<CloseEvent>;
+export type ErrorHandler = GenericHandler<Event>;
+/**
```
vojtechszocs, line 7: Nit: newline before export with JSDoc comment.

packages/lib-utils/src/ws/types.ts
```diff
@@ -0,0 +1,27 @@
...
+ * The WebSocket data can be returned in an object state or in the raw string response passed.
+ *
+ * This is configured through `jsonParse` options flag.
+ * @see WSOptions
```
vojtechszocs, line 11: FYI, if `WSOptions` type was imported, you could reference it like so
```
@see {@link WSOptions}
```
but there are times when it shouldn't be imported, so above is OK.

packages/lib-utils/src/ws/types.ts
vojtechszocs: It was a convention we started with.

For example, types for code in `packages/lib-utils/src/k8s` would live in `packages/lib-utils/src/types/k8s.ts` or `packages/lib-utils/src/types/k8s-whatever.ts` etc.

It's OK to put types close to relevant code. Also, we can always change the structure as necessary.

packages/lib-utils/src/ws/ws-utils.ts
```diff
@@ -0,0 +1,30 @@
+import { getUtilsConfig } from '../config';
+
+export const createURL = (host: string, path: string): string => {
```
vojtechszocs, line 3: Not sure if we're tracking it, but this reminds me that we should have an ESLint rule that consistently enforces an explicit return type for all functions and class methods, as we discussed earlier.

vojtechszocs (changes requested):

packages/lib-utils/src/ws/WSFactory.ts
andrewballantyne: > Vojtech Szocs
> I understand your point, let's do what is best and easiest. Putting types close to implementation makes sense.
> Having a separate types.ts next to specific code sounds reasonable.

This was agreed upon in slack to go the other direction and keep my local types file. Not going to do this request as a result.

packages/common/src/types/fetch.ts
```diff
@@ -6,3 +6,11 @@ export type ResourceFetch = (
   requestInit?: RequestInit,
   isK8sAPIRequest?: boolean,
 ) => Promise<Response>;
+
+/**
+ * Configure to use WebSockets.
+ */
+export type WebSocketConfigs = {
```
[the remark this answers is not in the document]
andrewballantyne: Ah good point, I was co-locating. I'll create a types file for the Config stuff.

packages/lib-utils/src/config.ts
```diff
@@ -11,6 +11,11 @@ export type UtilsConfig = {
    * with an appropriate error.
    */
   appFetch: ResourceFetch;
+
+  /**
+   * Configure the WS settings for your application.
```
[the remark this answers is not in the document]
andrewballantyne: Fair, I'll expand the name

packages/lib-utils/src/ws/WSFactory.ts
```diff
@@ -0,0 +1,407 @@
+import { consoleLogger } from '@monorepo/common';
```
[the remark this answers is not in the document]
andrewballantyne: Yeah there isn't a way to do "folder" or "file" comments sadly... will do this as we discussed it on Slack as well.

packages/lib-utils/src/ws/WSFactory.ts
```diff
@@ -0,0 +1,407 @@
...
+
+  private connectionAttempt: ReturnType<typeof setTimeout>;
+
+  private ws: WebSocket | null = null;
+
+  /**
+   * @param id - unique id for the WebSocket.
+   * @param options - websocket options to initiate the WebSocket with.
+   */
+  constructor(private readonly id: string, private readonly options: WSOptions) {
```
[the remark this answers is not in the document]
andrewballantyne: Hmm... if that gets properly pulled by our doc tool when we add it, that actually seems like a much better structure. Will do and hopefully we remember to test this use-case 🙂

packages/lib-utils/src/ws/ws-utils.ts
andrewballantyne: I'll make sure one is on the backlog... good recall.

andrewballantyne: Added here: https://issues.redhat.com/browse/HAC-782

PR #39, opened 2022-03-03
Update publish docs and fix npm prepublish hook

vojtechszocs:
- simplify docs for publishing SDK packages
- bump `lib-utils` version to match what's currently published on npmjs :package: 
- use `prepublishOnly` hook instead of `prepublish` which has been deprecated by npm [1]

[1] https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts

Here's what happens when we run `npm publish` command (before the actual publish to npm registry):
1. `prepublishOnly` lifecycle hook kicks in :arrow_right: we run unit tests
2. `prepack` lifecycle hook kicks :arrow_right: we build the package

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,12 +30,6 @@ Each package is versioned and published independently from other packages.
 
 Make sure the `npm` package manager that comes with Node.js is up-to-date.
 
-To see the current installed version:
-
-```sh
-npm ls -g npm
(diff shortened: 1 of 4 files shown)


andrewballantyne (approved):
LGTM

PR #41, opened 2022-03-04
Fix webpack import by importing default, instead of star

karelhala:
### Adjust webpack import

Since webpack exports just default export we have to adjust the import to it. With the current import we'd have to access it as `webpack.default.`. This is generally a pitfall of using `*` import as it means `import everything` and if some module exports `default` and not named exports it will fail.

### DIFF
--- a/packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
+++ b/packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
@@ -7,7 +7,7 @@ import type { Extension } from '@openshift/dynamic-plugin-sdk';
 import * as _ from 'lodash-es';
 import * as glob from 'glob';
 import * as path from 'path';
-import * as webpack from 'webpack';
+import webpack from 'webpack';
 import type { PluginBuildMetadata } from '../types/plugin';
 import { parseJSONFile } from '../utils/json';
 import { pluginBuildMetadataSchema } from '../yup-schemas';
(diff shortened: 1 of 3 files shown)


packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -7,7 +7,7 @@ import type { Extension } from '@openshift/dynamic-plugin-sdk';
 import * as _ from 'lodash-es';
 import * as glob from 'glob';
 import * as path from 'path';
-import * as webpack from 'webpack';
+import webpack from 'webpack';
```
vojtechszocs, line 10: This seems related to https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports

If we want to change this, we should update all such imports
```
$ git grep " from 'webpack';"
packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts:import * as webpack from 'webpack';
packages/lib-webpack/src/webpack/GenerateManifestPlugin.ts:import * as webpack from 'webpack';
packages/lib-webpack/src/webpack/PatchContainerEntryPlugin.ts:import * as webpack from 'webpack';
```

What's important is how this gets reflected into `packages/lib-webpack/dist/index.js` output.

karelhala: Good point! updated.

vojtechszocs (approved):

PR #42, opened 2022-03-04
Fix plugin loading mechanism by including version as well

karelhala:
### Failing to load plugin

When loading the plugin from hac-core these plugins are stored in hash as `pluginName@pluginVersion` and it fails to load plugin bootstrapped with this webpack plugin since we are not including the version when calling the bootstrap function. This PR fixes this issue by passing down version of said plugin.


packages/lib-webpack/src/webpack/PatchContainerEntryPlugin.ts
```diff
@@ -2,7 +2,7 @@ import { REMOTE_ENTRY_SCRIPT, REMOTE_ENTRY_CALLBACK } from '@openshift/dynamic-p
 import * as webpack from 'webpack';
 
 export class PatchContainerEntryPlugin implements webpack.WebpackPluginInstance {
-  constructor(private readonly pluginName: string) {}
+  constructor(private readonly pluginName: string, private readonly pluginVersion: string) {}
```
andrewballantyne, line 5: Did we want this to be optional? `= ''` or something? Seems like you just upped the requirement to use it.

karelhala:
Closing in favor of https://github.com/openshift/dynamic-plugin-sdk/pull/44

PR #43, opened 2022-03-04
Allow passing classname to route

karelhala:
This is a mirror of https://github.com/openshift/hac-core/pull/38

### DIFF
--- a/packages/lib-core/src/extensions/pages.ts
+++ b/packages/lib-core/src/extensions/pages.ts
@@ -31,6 +31,8 @@ export type RoutePageProperties = {
   path: string | string[];
   /** When true, will only match if the path matches the `location.pathname` exactly. */
   exact?: boolean;
+  /** Option to add custom className to wrapper component */
+  className?: string;
 };
 
 export type ResourcePageProperties = {


andrewballantyne (approved):
I have no problems with this.

/hold
Not sure if it'll merge right away, but I'll let others comment if they see a desire to.

christianvogt:
What's the use case for allowing someone to override the classname?
The best extensions are those that are data driven and don't allow UI customizations that can lead to inconsistencies.

andrewballantyne:
> What's the use case for allowing someone to override the classname? The best extensions are those that are data driven and don't allow UI customizations that can lead to inconsistencies.

This isn't overriding, this is the wrapping class name... perhaps it needs better documentation

christianvogt:
Wouldn't the same be achieved by the route component assigning their own classname to their root element?

I don't think this should be a substitute for pre-fixing though if a page is extensible with the inclusion of externally defined elements being placed within your dom tree. There's still potential for clashing.

andrewballantyne:
> Wouldn't the same be achieved by the route component assigning their own classname to their root element?
>
> I don't think this should be a substitute for pre-fixing though if a page is extensible with the inclusion of externally defined elements being placed within your dom tree. There's still potential for clashing.

@karel made an argument for allowing more flexibility with the interfacing app rather than them configuring this at the root of very page they own. I think this is okay because webpack seem to be able to prefix the CSS from a compile step and not something they need to manually do everywhere. 

If they had to manually add this to everything, yeah I can see them managing this at the top level of each of their routes. However it seems to be a compile step, so this makes sense to me.

christianvogt:
Whether it's configured at the root of every route component or each route extension, it's still done per route.

andrewballantyne:
> Whether it's configured at the root of every route component or each route extension, it's still done per route.

Then I think this is something you need to take a stance on as Architect and inform HAC-Dev... this is a solution to the current implementation they have as far as I understand it.

christianvogt:
Thanks @andrewballantyne, I'll connect with hac-dev to hear their side as I still don't understand the exact need for this over the existing apis available.

christianvogt:
@karelhala I don't want to impose on devs the need to add the classname in their extensions to fulfill the prefix requirement imposed by the insights webpack config whereby it adds a sass-prefix.

Can you not obtain the prefix by other means similar to what's already being done for existing apps?

karelhala:
@christianvogt yeah, we can by all means use same principle - just the pluin name as class prefix. I was more thinking about being nice to the devs and give them option to change this. I've added a topic for next Arch sync to go trough more generic approach on this one.

Closing this as it's not really required.

PR #45, opened 2022-03-07
Add useK8sWatchResource hook

andrewballantyne:
https://issues.redhat.com/browse/HAC-489

Moves the `useK8sWatchResource` hook into the SDK (from OpenShift Console)

Notes:

- `useK8sWatchResources` (plural) was not moved -- it's a convenience function to do multiple of the watches, it brings in an external dependency that looks unneeded -- to avoid this slowdown this can be another ticket (https://issues.redhat.com/browse/HAC-791)
- Types were cleaned up to meet SDK standards -- hopefully it's not too strict on usage (for instance `useK8sWatchResource` now requires a value; whereas before it was okay with `null`
- Renamed `getIDAndDispatch` because the name was weird and would return a `dispatch` property that was then itself dispatched; renamed it to `getWatchData` as it is related to the watch hooks & the prop is now `action` to meet a known `dispatch(action)` syntax of Redux.

~Based off of https://github.com/openshift/dynamic-plugin-sdk/pull/37 -- Ignore first commit~


packages/lib-utils/src/k8s/hooks/index.ts
```diff
@@ -0,0 +1 @@
+export { useK8sWatchResource } from './useK8sWatchResource';
```
andrewballantyne, line 1: @vojtechszocs I'm pretty sure you said I don't need to export types used -- is that true or should I be exporting the type behind this?

packages/lib-utils/src/k8s/hooks/useDeepCompareMemoize.ts
```diff
@@ -0,0 +1,16 @@
+import * as _ from 'lodash-es';
+import * as React from 'react';
+
+export const useDeepCompareMemoize = <T = unknown>(value: T, stringify?: boolean): T => {
```
andrewballantyne, line 4: This was co-located with the hooks in the OpenShift Console SDK -- this can be more generically used. Not sure we want to export it though. But maybe as a utility method? Up to discussion.

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
```diff
@@ -0,0 +1,74 @@
...
+    return () => {
+      if (reduxID) {
+        dispatch(k8sActions.stopK8sWatch(reduxID.id));
+      }
+    };
+  }, [dispatch, reduxID]);
+
+  const resourceK8s = useSelector<SDKStoreState, unknown>((state) =>
+    reduxID ? getReduxIdPayload(state, reduxID.id) : null,
+  ) as ImmutableMap<string, unknown>; // TODO: Store state based off of Immutable is problematic
```
andrewballantyne, line 56: All uses of ImmutableMap cause issues because we do not properly type our K8s store. This needs to be fixed all together in a future PR.

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -207,18 +207,17 @@ export const getGroupVersionKindForModel: GetGroupVersionKindForModel = ({
...
  * * */
 export const getReference = ({
-  apiGroup,
-  apiVersion,
+  group,
+  version,
   kind,
-}: K8sResourceIdentifier): K8sResourceKindReference =>
-  [apiGroup || 'core', apiVersion, kind].join('~');
+}: K8sGroupVersionKind): K8sResourceKindReference => [group || 'core', version, kind].join('~');
```
andrewballantyne, line 220: @florkbr Not sure what you were up to with this change... but I have reverted it. `K8sResourceIdentifier` and `K8sGroupVersionKind` are not the same type.

andrewballantyne (commented):
Some review comments.

packages/lib-utils/src/k8s/hooks/useDeepCompareMemoize.ts
florkbr: Hm... Did we take https://www.npmjs.com/package/use-deep-compare and modify it to play nice with typescript?

packages/lib-utils/src/k8s/hooks/k8s-watcher.ts
```diff
@@ -0,0 +1,116 @@
...
+  if (!_.isEmpty(query)) {
+    qs = `---${JSON.stringify(query)}`;
+  }
+
+  return `${cluster ?? 'local-cluster'}${getReferenceForModel(
+    k8sKind || ({} as K8sModelCommon),
+  )}${qs}`;
+};
+
+export const makeQuery: MakeQuery = (namespace, labelSelector, fieldSelector, name, limit) => {
```
florkbr, line 27: I wonder if we could remove the multiple `if`'s and use something like 
```js
export const makeQuery: MakeQuery = (namespace, ...args) => {
  const query: Query = {
    ns: namespace,
    ...args
  };

  return _.pickBy(query, _.identity);
};
```

NOTE: if we expect to pass any falsey values back in the query object this won't work (we'd need to tweak the return filter).

packages/lib-utils/src/types/k8s.ts
```diff
@@ -108,6 +108,8 @@ export type FilterValue = Partial<{
 
 export type GetGroupVersionKindForModel = (model: K8sModelCommon) => K8sResourceIdentifier;
 
+export type K8sGroupVersionKind = { group?: string; version: string; kind: string };
```
florkbr, line 111: 👍 thank you @andrewballantyne !

packages/lib-utils/src/k8s/k8s-utils.ts
florkbr: Yep. Thanks! Not enough coffee ☕

packages/lib-utils/src/k8s/hooks/k8s-watcher.ts
andrewballantyne: Method args are not named... args spread would be an array, not a keyed item.

florkbr: Good point. I'm fine with leaving this.

packages/lib-utils/src/k8s/hooks/watch-resource-types.ts
```diff
@@ -0,0 +1,59 @@
+import type {
+  K8sGroupVersionKind,
+  K8sResourceCommon,
+  K8sResourceKindReference,
+  Selector,
+} from '../../types/k8s';
+
+export type WatchK8sResult<R extends K8sResourceCommon | K8sResourceCommon[]> = [
```
florkbr, line 8: Still need to clean up the type situation present in the SDK today (in root type folder versus next to the consuming JS). I'm fine with this for now!

packages/lib-utils/src/k8s/hooks/useDeepCompareMemoize.ts
florkbr: Chatted with Andrew and agreed to create a follow-up investigation item: https://issues.redhat.com/browse/HAC-795

packages/lib-utils/src/k8s/hooks/watch-resource-types.ts
andrewballantyne: Yeah, I did modify any existing structure, just co-located stuff as I saw fit for the objective -- based on that conversation

packages/lib-utils/src/k8s/hooks/index.ts
vojtechszocs: @andrewballantyne This ultimately depends on whether we want to have a direct export of `UseK8sWatchResource` type in `packages/lib-utils/src/index.ts` which is the main module of the library.

Taking `packages/lib-core/src/index.ts` as an example, we currently don't export dedicated hook function types, so I'd add it later if consumers need to reference the type directly.

vojtechszocs: > don't need to export types used

When a function `foo` refers to type `X` even if type `X` isn't directly exported in the main module, it's still included in the generated `dist/index.d.ts` typings bundle.

The decision whether or not to export `X` directly depends on whether or not consumers need to reference it directly.

packages/lib-utils/src/k8s/hooks/useDeepCompareMemoize.ts
vojtechszocs: @florkbr I think @rawagner wrote the initial impl. for this hook so he might know more about the intention behind it.

packages/lib-utils/src/k8s/hooks/watch-resource-types.ts
```diff
@@ -0,0 +1,59 @@
...
+  R,
```
vojtechszocs, line 9: Nit: TS 4+ allows discrete tuple (array) elements to have meaningful names.

See for example `UseResolvedExtensionsResult` in `packages/lib-core/src/store/useResolvedExtensions.ts`

```ts
type UseResolvedExtensionsResult<TExtension extends Extension> = [
  resolvedExtensions: LoadedExtension<ResolvedExtension<TExtension>>[],
  resolved: boolean,
  errors: unknown[],
];
```

Can be done later as an improvement.

vojtechszocs (changes requested):

packages/lib-utils/src/k8s/hooks/index.ts
andrewballantyne: Right, I get that ... but that's only helpful when we are talking docs... for other TS usages, it won't be available. Hence why I pushed out the Model Common and Resource Common types in my other PR -- those need to be used by the caller/user of the SDK.

packages/lib-utils/src/k8s/hooks/watch-resource-types.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export type WatchK8sResults<R extends ResourcesObject> = {
+  [k in keyof R]: WatchK8sResultsObject<R[k]>;
+};
+
+/** @deprecated Use groupVersionKind instead. The kind property will be removed in a future release. */
+type UseKind = {
+  kind: K8sResourceKindReference;
+  groupVersionKind?: never;
```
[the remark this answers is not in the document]
vojtechszocs: Ohh right, I see - thanks for clarification.

Another way to express this mutual property relationship is
```ts
type UseKindOrReference =
  | {
      kind: K8sResourceKindReference;
      groupVersionKind: never;
    }
  | {
      kind: never;
      groupVersionKind: K8sGroupVersionKind;
    };

export type WatchK8sResource = UseKindOrReference & {
  // ..
};
```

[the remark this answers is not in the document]
andrewballantyne: Addressed in https://github.com/openshift/dynamic-plugin-sdk/pull/52

PR #50, opened 2022-03-08
Allow WS code to support additional args

andrewballantyne:
I noticed a couple gaps in what we need to support WS.

1. The caller of the `WebSocketFactory` needs to be able to build a proper qualified `path` -- `getK8sResourceURL` is what they would use in OpenShift Console and thus we should export it for them to use.
2. The host app needs to be able to augment the URL to manage it with the host target... in the case of DevSandbox, a suffix of `?watch=true` is needed
3. Added JSDoc comments to the web socket params

Also changed
1. Also exported the Common types for Model/Resource -- as they are defined internally in the package but not exposed for the users of the SDK to use (which is important for the `getK8sResourceURL` call as well as other usages)
    image

2. I renamed `webSocket` folder to `web-socket` as we use kebab-case in the SDK packages


florkbr (approved):
LGTM 👍

PR #52, opened 2022-03-08
Support EitherNotBoth and EitherOrNone type helpers

andrewballantyne:
Useful types to help us write code that is more explicit in its intent. Made one change to support a comment @vojtechszocs made: https://github.com/openshift/dynamic-plugin-sdk/pull/45#discussion_r821975127

### DIFF
--- a/packages/common/src/types/common.ts
+++ b/packages/common/src/types/common.ts
@@ -23,3 +23,68 @@ export type AnyFunction = (...args: any) => any;
 export type ReplaceProperties<T extends AnyObject, R extends AnyObject> = {
   [K in keyof T]: K extends keyof R ? R[K] : T[K];
 } & AnyObject;
+
+/**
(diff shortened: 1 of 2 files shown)


packages/common/src/types/common.ts
```diff
@@ -23,3 +23,68 @@ export type AnyFunction = (...args: any) => any;
...
   [K in keyof T]: K extends keyof R ? R[K] : T[K];
 } & AnyObject;
+
+/**
+ * Never allow any properties of `Type`.
+ *
+ * Utility type, probably never a reason to export.
+ */
+type Never<Type> = {
+  [Property in keyof Type]?: never;
```
vojtechszocs, line 33: Nit: when referring to key types we can simply use `K`.

```suggestion
  [K in keyof Type]?: never;
```

vojtechszocs (approved):

PR #57, opened 2022-03-09
[WIP] Random fixes to lib-utils to make it work with HAC-Core

andrewballantyne:
WIP because:

- The fix for the store is not good... try/catch should not be used with a hook
- Still trying to test out things in HAC-Core (see https://github.com/openshift/hac-core/pull/43)

### DIFF
--- a/packages/lib-utils/src/app/AppInitSDK.tsx
+++ b/packages/lib-utils/src/app/AppInitSDK.tsx
@@ -10,8 +10,6 @@ import { initAPIDiscovery } from './api-discovery';
 import { useReduxStore } from './redux';
 
 type AppInitSDKProps = {
-  /** Only child is your Application */
-  children: React.ReactElement;
   configurations: {
     apiDiscovery?: InitAPIDiscovery;
     appFetch: UtilsConfig['appFetch'];
@@ -39,17 +37,25 @@ type AppInitSDKProps = {
  */
 const AppInitSDK: React.FC<AppInitSDKProps> = ({ children, configurations }) => {
   const { store, storeContextPresent } = useReduxStore();
+  const [canRender, setCanRender] = React.useState(false);
+
   const { appFetch, pluginStore, wsAppSettings, apiDiscovery = initAPIDiscovery } = configurations;
+
   React.useEffect(() => {
     try {
-      if (isUtilsConfigSet()) {
+      if (!isUtilsConfigSet()) {
         setUtilsConfig({ appFetch, wsAppSettings });
+        setCanRender(true);
       }
       apiDiscovery(store);
     } catch (e) {
       consoleLogger.warn('Error while initializing AppInitSDK', e);
     }
   }, [apiDiscovery, appFetch, store, wsAppSettings]);
+
+  if (!canRender) {
+    return null;
+  }
 
   return (
     <PluginStoreProvider store={pluginStore}>
--- a/packages/lib-utils/src/app/api-discovery/index.ts
+++ b/packages/lib-utils/src/app/api-discovery/index.ts
@@ -78,13 +78,15 @@ const getResources = async (): Promise<DiscoveryResources> => {
     },
     {},
   );
-  const all: Promise<APIResourceList>[] = _.flatten(
-    apiResourceData.groups.map((group) =>
-      group.versions.map((version) => `/apis/${version.groupVersion}`),
+  const all = _.flatten(
+    apiResourceData.groups.map<string[]>((group) =>
+      group.versions.map<string>((version) => `/apis/${version.groupVersion}`),
     ),
   )
     .concat(['/api/v1'])
-    .map((p) => commonFetchJSON<K8sModelCommon>(`api/kubernetes${p}`).catch((err) => err));
+    .map<Promise<APIResourceList>>((p: string) =>
+      commonFetchJSON<K8sModelCommon>(p).catch((err) => err),
+    );
 
   return Promise.all(all).then((data) => {
     const resourceSet = new Set<string>();
--- a/packages/lib-utils/src/app/redux/index.ts
+++ b/packages/lib-utils/src/app/redux/index.ts
@@ -25,7 +25,14 @@ type UseReduxStoreResult = {
  * ```
  */
 export const useReduxStore = (): UseReduxStoreResult => {
-  const storeContext = useStore();
+  let storeContext: Store | null = null;
+  try {
+    // It'll always be invoked -- it just might blow up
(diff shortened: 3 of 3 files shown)


packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -39,18 +37,26 @@ type AppInitSDKProps = {
...
+        setCanRender(true);
       }
       apiDiscovery(store);
     } catch (e) {
       consoleLogger.warn('Error while initializing AppInitSDK', e);
     }
   }, [apiDiscovery, appFetch, store, wsAppSettings]);
 
+  if (!canRender) {
+    return null;
```
andrewballantyne, line 57: https://coreos.slack.com/archives/C02EQERSV3N/p1646860045455809

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -39,18 +37,26 @@ type AppInitSDKProps = {
...
 const AppInitSDK: React.FC<AppInitSDKProps> = ({ children, configurations }) => {
   const { store, storeContextPresent } = useReduxStore();
+  const [canRender, setCanRender] = React.useState(false);
+
   const { appFetch, pluginStore, wsAppSettings, apiDiscovery = initAPIDiscovery } = configurations;
+
   React.useEffect(() => {
     try {
-      if (isUtilsConfigSet()) {
+      if (!isUtilsConfigSet()) {
```
andrewballantyne, line 46: Oops 🙂

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -10,8 +10,6 @@ import { initAPIDiscovery } from './api-discovery';
 import { useReduxStore } from './redux';
 
 type AppInitSDKProps = {
-  /** Only child is your Application */
-  children: React.ReactElement;
```
andrewballantyne, line 14: `React.FC` covers this... not sure why we overrode it 🤔 

Causes a weird call: https://github.com/openshift/hac-core/pull/43/files#diff-407ff414eb218ddfa29e668c31dfa67956c70968492131cf233a735742ca1addR16

packages/lib-utils/src/app/api-discovery/index.ts
```diff
@@ -78,13 +78,15 @@ const getResources = async (): Promise<DiscoveryResources> => {
...
-  const all: Promise<APIResourceList>[] = _.flatten(
-    apiResourceData.groups.map((group) =>
-      group.versions.map((version) => `/apis/${version.groupVersion}`),
+  const all = _.flatten(
+    apiResourceData.groups.map<string[]>((group) =>
+      group.versions.map<string>((version) => `/apis/${version.groupVersion}`),
     ),
   )
     .concat(['/api/v1'])
-    .map((p) => commonFetchJSON<K8sModelCommon>(`api/kubernetes${p}`).catch((err) => err));
```
andrewballantyne, line 87: Fairly certain `api/kubernetes` is the OpenShift Console backend proxy...

packages/lib-utils/src/app/redux/index.ts
```diff
@@ -25,7 +25,14 @@ type UseReduxStoreResult = {
...
 export const useReduxStore = (): UseReduxStoreResult => {
-  const storeContext = useStore();
+  let storeContext: Store | null = null;
+  try {
+    // It'll always be invoked -- it just might blow up
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    storeContext = useStore();
+  } catch (e) {
+    // TODO: Handle this better...
+  }
```
andrewballantyne, line 35: This blows up with or without the provider used in HAC-Core's AppEntry file. Not sure what is happening here. This is how I fixed it locally. This is NOT a solution though.

andrewballantyne (commented):
Reasonings behind the changes.

packages/lib-utils/src/app/redux/index.ts
karelhala: Kinda curious why we need this, I've been running both hac-core and hac-dev locally for some time and didn't see this error before :thinking:

packages/lib-utils/src/app/AppInitSDK.tsx
florkbr: I could've sworn this has come up before and I removed `children` then added it back... but now I can't find the comment. I'm fine removing it.

packages/lib-utils/src/app/AppInitSDK.tsx
florkbr: 🙈

packages/lib-utils/src/app/AppInitSDK.tsx
florkbr: As a very temporary fix to the hook firing order this is fine

packages/lib-utils/src/app/api-discovery/index.ts
florkbr: Yep. My bad

packages/lib-utils/src/app/redux/index.ts
```diff
@@ -25,7 +25,14 @@ type UseReduxStoreResult = {
...
  */
 export const useReduxStore = (): UseReduxStoreResult => {
-  const storeContext = useStore();
+  let storeContext: Store | null = null;
+  try {
+    // It'll always be invoked -- it just might blow up
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    storeContext = useStore();
+  } catch (e) {
+    // TODO: Handle this better...
```
florkbr, line 34: @andrewballantyne would like to add a `consoleLogger.error` log here if you're making more changes.

packages/lib-utils/src/app/AppInitSDK.tsx
andrewballantyne: Right, well I think we wanted to control the flow -- but this makes it very difficult for others to use the API. I guess if we want this as a stance, I can modify the props in my component to only accept one and thus thread that logic forward... but I'm wondering why we did it as there is nothing I can think of that won't work if we just do it this way.

If I recall correctly, there was a lint error on passing children without Fragments... but it wasn't necessarily a "design choice" so much as a "fix for the eslint disable"

packages/lib-utils/src/app/redux/index.ts
andrewballantyne: Yup, definitely can

florkbr:
Closing in favor of https://github.com/openshift/dynamic-plugin-sdk/pull/61

PR #58, opened 2022-03-10
Prevent API discovery from sending too many requests

florkbr:
New flow:
1. Wait API_DISCOVERY_INIT_DELAY milliseconds before triggering API discovery
2. Actually wait API_DISCOVERY_POLL_INTERVAL milliseconds from the time the last batch finished (previously this was in parallel while the other batch was in flight)
3. During each loop, request API_DISCOVERY_REQUEST_BATCH_SIZE batch resources at a time.

Tested against HAC core running against dev sandbox using @andrewballantyne's PRs:
https://github.com/openshift/dynamic-plugin-sdk/pull/57
https://github.com/openshift/hac-core/pull/43


karelhala (commented):
Looking good! @florkbr once you fix those merge issues we are good to go.

karelhala (approved):
/lgtm

PR #60, opened 2022-03-22
[WIP] Use storybook to test the testK8s page from hac-core

florkbr:


andrewballantyne:
@florkbr Why are we bringing over the test page from HAC-Core? Shouldn't we look to make this an e2e plugin for HAC-Core? (I'd like we'd also co-locate that code within it)

florkbr:
I am closing this and will reopen with a better example - agree with Andrew.

PR #61, opened 2022-03-24
Random fixes to lib-utils to make it work with HAC-Core

florkbr:
Taking/running the torch from @andrewballantyne and https://github.com/openshift/dynamic-plugin-sdk/pull/57

WIP because:

* The fix for the store is not good... try/catch should not be used with a hook
* Still trying to test out things in HAC-Core (see https://github.com/openshift/hac-core/pull/43)

### DIFF
--- a/packages/lib-utils/src/app/redux/index.ts
+++ b/packages/lib-utils/src/app/redux/index.ts
@@ -31,7 +31,8 @@ export const useReduxStore = (): UseReduxStoreResult => {
     // eslint-disable-next-line react-hooks/rules-of-hooks
     storeContext = useStore();
   } catch (e) {
-    // TODO: Handle this better...
+    // TODO: remove once proven not needed (redux versioning issue)
+    consoleLogger.error(e);
   }
   const [storeContextPresent, setStoreContextPresent] = React.useState(false);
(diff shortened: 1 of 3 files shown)


florkbr:
@karelhala we lost the original review on https://github.com/openshift/hac-core/pull/43. Can you please review? Thanks.

packages/lib-utils/src/app/redux/index.ts
```diff
@@ -25,7 +25,15 @@ type UseReduxStoreResult = {
...
-  const storeContext = useStore();
+  let storeContext: Store | null = null;
+  try {
+    // It'll always be invoked -- it just might blow up
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    storeContext = useStore();
+  } catch (e) {
+    // TODO: Handle this better...
+    consoleLogger.error(e);
+  }
```
karelhala, line 36: I don't think we need this anymore (given we update the redux to same version hosted from hac-core). But let's keep it here and remove later down the line. Can you add `TODO: remove once proven not needed`?

karelhala (commented):
Looking good, just one minor update wrt `TODO`

PR #62, opened 2022-03-25
HAC-268: Add sample plugin and improve other packages

vojtechszocs:
Fixes [HAC-268](https://issues.redhat.com/browse/HAC-268)

This PR adds a minimal sample plugin as well as a whole bunch of other improvements.

- ESLint is now executed once instead of per-package (it's faster and more flexible in terms of linted file paths)
- added `build-libs` script which builds all distributable SDK packages (`lib-core`, `lib-utils`, `lib-webpack`)
- common development-related dependencies moved into monorepo root package manifest
- `lib-core` now declares `react-router` and `redux` as optional via `peerDependenciesMeta`
- `lib-webpack` is now bundled using CommonJS module format, with `lodash-es` :arrow_right: `lodash` replacement
- `lib-webpack` no longer peer-depends on `lib-core` package, it simply includes the necessary bits of code
- `main` package manifest field is now set via `publishConfig` right before a package is published to npmjs
- Yarn executable bumped to latest version :memo: [3.2.0](https://dev.to/arcanis/yarn-32-libc-yarn-explain-next-major--o22)
- Yarn `plugin-workspace-tools` bumped to latest version
- update & reorganize project documentation


packages/common/package-utils.js
```diff
@@ -0,0 +1,32 @@
+import { mkdirSync, writeFileSync } from 'fs';
+import _ from 'lodash';
+
+/**
+ * Get the distributable package manifest ready for publishing.
+ *
+ * @param {'esm' | 'cjs'} format
+ */
+export const getDistPackageManifest = (pkg, format = 'esm') => {
+  const distPkg = _.cloneDeep(pkg);
```
karelhala, line 10: Mhh, we could probably use [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax). We are not mutating any deep values. Also we could use destructor here

```JS
const { devDependencies, scripts, dependencies, ...distPkg } = pkg;
```

And later when we remove `lodash-es` we can delete it from `dependencies`. And when returning the package
```JS
return {
  ...distPkg,
  dependencies
}
```

packages/common/rollup-configs.js
```diff
@@ -20,46 +21,79 @@ const getExternalModuleTester = (pkg) => {
...
 
 /**
+ * @returns {import('rollup').Plugin}
+ */
+const replaceLodashEsRequire = () => ({
+  name: 'replace-lodash-es-require',
+
+  renderChunk(code, chunk) {
+    return {
+      code: code.replace("require('lodash-es')", "require('lodash')"),
```
karelhala, line 47: This will not work for direct imports, better to change this to less strict replace
```
code.replace("require('lodash-es", "require('lodash")
```

packages/lib-core/package.json
```diff
@@ -3,46 +3,33 @@
...
-    "@types/react": "^17.0.37",
-    "@types/react-router-dom": "^5.1.2",
-    "react": "^17.0.2",
-    "redux": "^4.1.2",
-    "rollup": "^2.61.1",
-    "rollup-plugin-analyzer": "^4.0.0",
-    "rollup-plugin-dts": "^4.0.1",
-    "tslib": "^2.3.1",
-    "typescript": "~4.4.4"
+  "peerDependenciesMeta": {
```
karelhala, line 27: I like the usage of this!

packages/lib-utils/package.json
```diff
@@ -37,21 +29,6 @@
...
-    "react": "^17.0.2",
-    "react-redux": "^7.2.2",
-    "redux": "^4.1.2",
-    "redux-thunk": "^2.4.1",
-    "rollup": "^2.61.1",
-    "rollup-plugin-analyzer": "^4.0.0",
-    "rollup-plugin-dts": "^4.0.1",
-    "tslib": "^2.3.1",
-    "typescript": "~4.4.4"
+    "@openshift/dynamic-plugin-sdk": "link:../lib-core/dist"
```
karelhala, line 32: We can now use `*` in here. No need to use the `link` value.
```suggestion
    "@openshift/dynamic-plugin-sdk": "*"
```

packages/lib-webpack/package.json
```diff
@@ -3,48 +3,27 @@
...
     "test": "echo 'Write some tests!'"
   },
   "peerDependencies": {
-    "@openshift/dynamic-plugin-sdk": "^1.0.0",
     "webpack": "^5.69.1"
   },
   "dependencies": {
     "glob": "^7.2.0",
+    "lodash": "^4.17.21",
     "lodash-es": "^4.17.21",
```
karelhala, line 24: We should pick one and remove the other. Are we using `lodash-es` in this package?

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -1,12 +1,13 @@
...
+  PLUGIN_MANIFEST,
   REMOTE_ENTRY_SCRIPT,
   REMOTE_ENTRY_CALLBACK,
-  extensionArraySchema,
-} from '@openshift/dynamic-plugin-sdk';
-import type { Extension } from '@openshift/dynamic-plugin-sdk';
+} from '@openshift/dynamic-plugin-sdk/src/constants';
+import { extensionArraySchema } from '@openshift/dynamic-plugin-sdk/src/yup-schemas';
+import type { EncodedExtension } from '@openshift/dynamic-plugin-sdk/src/types/extension';
 import * as _ from 'lodash-es';
```
karelhala, line 8: I would use `lodash` here directly since we are exposing `cjs` package. And while on it we can import just what we use in here.

packages/sample-plugin/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "private": true,
+  "scripts": {
+    "clean": "rm -rf dist",
+    "build": "yarn clean && NODE_ENV=production yarn webpack",
+    "build-dev": "yarn clean && yarn webpack",
+    "webpack": "node -r ts-node/register $PROJECT_CWD/node_modules/.bin/webpack"
+  },
+  "devDependencies": {
+    "@openshift/dynamic-plugin-sdk": "link:../lib-core/dist",
+    "@openshift/dynamic-plugin-sdk-webpack": "link:../lib-webpack/dist"
```
karelhala, line 14: We could use asterix here as well.

packages/common/package-utils.js
vojtechszocs: @karelhala Note that I've removed this file in a later update.

`main` field of package manifest is now set via `publishConfig`, other fields like `files` (which have no impact when importing from within another monorepo package) are kept as before.

`types` field was removed. When absent, TypeScript checks the `main` field and, if present, infers the `.d.ts` from it, see [here](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html#editing-the-packagejson) for reference.

> `“main”:“index.js”` :arrow_right: `index.d.ts`

As for Lodash, `lib-webpack` is built against `lodash-es` since the standard CommonJS `lodash` version has tons of issues when it comes to tree shaking and module bundling in general.

At the end of the build, we simply replace the `require()` for `lodash-es` with `lodash` in the generated JS bundle.

packages/common/rollup-configs.js
vojtechszocs: This is also the reason why there's no `lodash-es` in `lib-webpack`'s `dependencies`.

@karelhala @florkbr In each "lib" package, `peerDependencies` and `dependencies` reflect the consumer side of the package.

packages/lib-core/package.json
vojtechszocs: @karelhala Thanks, the reason I've used it is because it allows devs to write simple plugins that use only a small subset of extension types, i.e. not necessarily extension types that depend on Redux etc.

This is assuming that plugins use TypeScript to declare their extensions, since `react-router` and `redux` dependencies here are utilized only via type imports.

packages/lib-utils/package.json
vojtechszocs: @karelhala This dev-dependency results in Yarn creating `packages/lib-utils/node_modules/@openshift/dynamic-plugin-sdk` specifically pointing to the generated distributable output of `lib-core` package.

I believe this is what we want in general - when building `lib-utils` which has a peer-dependency on `lib-core`, consume `lib-core` via its generated distributable outputs (same behavior as when used by specific plugins or host apps).

packages/lib-webpack/package.json
vojtechszocs: This should be fixed, `lib-webpack` is built against `lodash-es` but the CommonJS `require()` statement is modified to import `lodash`.

`lodash-es` is ES modules friendly (and therefore modern, tree shaking and optimization friendly) build of the standard CommonJS `lodash` build. Their APIs are identical.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
vojtechszocs: @karelhala For consistency, I'd prefer to use `lodash-es` everywhere in our sources, and adapt generated outputs as necessary.

With `lodash-es` you can do `import * as _` and the module bundler like Rollup will still be able to figure out which parts of this library are actively used in the compilation. Agreed that it'd be good to look into a good convention on how to handle Lodash imports, but I'd do it as a follow-up :small_airplane:

packages/sample-plugin/package.json
vojtechszocs: Check out my comment above, I've tried to explain why linking to `dist` of consumed monorepo packages would make sense, let me know what you think.

packages/common/rollup-configs.js
karelhala: @vojtechszocs I was more concerned about direct and named imports. If we were to use
```JS
import defaultsDeep from 'lodash-es/defaultsDeep'
```

That would be transplied as `var defaultsDeep = require('lodash-es');` but we can discuss the usage of this later down the line. No need to stall this PR on this anymore.

packages/lib-core/package.json
karelhala: Yeah, exactly! I've been scratching my head why we mark them as optional and went looking at the code and realised we use just types out of these packages.

packages/lib-utils/package.json
karelhala: I am more of a concerned how that looks when published to NPM. Wouldn't this cause an issue with unknown version? But since this is just dev dependency I don't think we have to spend more time on this. The code will be bundled in the final build after all.

karelhala (approved):
/lgtm

PR #68, opened 2022-04-12
HAC-791: Move useK8sWatchResources hook into the Core SDK

vidyanambiar:
This PR migrates the useK8sWatchResources hook over to the core SDK.

Please see the PR https://github.com/openshift/hac-core/pull/61 for how this can be tested using HAC Core's testK8s page.


vidyanambiar:
Note: This PR includes a change that reverts an update from this commit - https://github.com/openshift/console/commit/8b673199efe928a95a9fd21bfa47c0243c424972#diff-1347c512bf8e06590508b192459dfc8e04ce7a930fa955e986a3e266895d0ad7R41 that was part of the changes for fixing a bug https://bugzilla.redhat.com/show_bug.cgi?id=2027311 for watching core resources in the console.

The issue is that based on the commit linked above, when `getK8sModel` is called it looks for a model in the key path `RESOURCES -> models` , and the input param that we're giving it in the form of `k8s` is not `SDKStoreState.k8s`, we're instead passing the models already extracted from `k8s -> RESOURCES -> models ` (`allK8sModels` and `prevK8sModels`)  to this function as the value of input param `k8s`. Since models are already extracted from k8s , the `getK8sModel` function is unable to find the model using that original keypath. As a result the hook is unable to fetch data for the resource.

As discussed with @florkbr, we will need a follow-up work item to make sure this change doesn't cause any discrepancy in the console when the console is updated to use the core SDK.

packages/lib-utils/src/index.ts
```diff
@@ -3,6 +3,7 @@ export { SDKReducers } from './app/redux/reducers';
 export { UtilsConfig, isUtilsConfigSet, setUtilsConfig, getUtilsConfig } from './config';
 export { commonFetch, commonFetchText, commonFetchJSON } from './utils/common-fetch';
 export { useK8sWatchResource } from './k8s/hooks';
+export { useK8sWatchResources } from './k8s/hooks';
```
vojtechszocs, line 6: Can we use a single export here?

```ts
export { useK8sWatchResource, useK8sWatchResources } from './k8s/hooks';
```

packages/lib-utils/src/k8s/hooks/useK8sWatchResources.ts
```diff
@@ -0,0 +1,184 @@
...
+ * @return A map where keys are as provided in initResouces and value has three properties data, loaded and error.
+ * @example
+ * ```ts
+ * const Component: React.FC = () => {
+ *   const watchResources = {
+        'deployment': {...},
+        'pod': {...}
+        ...
+      }
+ *   const {deployment, pod}  = useK8sWatchResources(watchResources)
```
vojtechszocs, line 37:
```suggestion
 *   const {deployment, pod} = useK8sWatchResources(watchResources)
```

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
```diff
@@ -23,12 +24,14 @@ const NOT_A_VALUE = '__not-a-value__';
...
       }
- *   const [data, loaded, error] = UseK8sWatchResource(watchRes)
+ *   const [data, loaded, error] = useK8sWatchResource(watchRes)
  *   return ...
  * }
  * ```
  */
-export const useK8sWatchResource: UseK8sWatchResource = (initResource) => {
+export const useK8sWatchResource = <R extends K8sResourceCommon | K8sResourceCommon[]>(
+  initResource: WatchK8sResource | null,
```
florkbr, line 33: @vidyanambiar i'm not sure we want `| null` here. I think we'd want to enforce a K8s resource is always passed in - as it doesn't make much sense to watch nothing (unless I'm missing something). We can follow up on this in our TS cleanup: https://issues.redhat.com/browse/HAC-801

packages/lib-utils/src/k8s/hooks/useK8sWatchResources.ts
```diff
@@ -0,0 +1,184 @@
...
+  );
+
+  const prevK8sModels = usePrevious(allK8sModels);
+  const prevResources = usePrevious(resources);
+
+  // Ref to immutable map mapping model's kind (string) to the model object (ImmutableMap)
+  const k8sModelsRef = React.useRef<ImmutableMap<string, K8sModelCommon>>(ImmutableMap());
+
+  if (
+    prevResources !== resources ||
```
florkbr, line 60: This conditional is _huge_. I know we migrated it from console and it is working, but I'd like to move some of the `|| `and `&&` and `.some` logic to some variables for readability.

florkbr: @vidyanambiar let's follow up on this. Could you create a ticket for cleanup?

packages/lib-utils/src/k8s/hooks/useK8sWatchResources.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  type WatchModel = ReturnType<GetWatchData> & { noModel: boolean };
+
+  // reduxIDs -- Map of keys from "resources" to the {id, action} for watching the specific resource
+  const reduxIDs = React.useMemo<{
+    [key: string]: WatchModel;
+  } | null>(
+    () =>
+      modelsLoaded
+        ? Object.keys(resources).reduce(
```
florkbr, line 93: Same here as above, there is quite a bit of logic nested in a ternary here and it makes readability hard.

packages/lib-utils/src/k8s/hooks/usePrevious.ts
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+
+export const usePrevious = <P = unknown>(value: P, deps?: unknown[]): P | undefined => {
```
florkbr, line 3: I wonder if we could also improve our types here - I don't want the shared `usePrevious` utility to be a reason why we do not focus on explicit types in components/hooks.

packages/lib-utils/src/k8s/hooks/watch-resource-types.ts
```diff
@@ -7,13 +7,13 @@ import type {
 } from '../../types/k8s';
 
 export type WatchK8sResult<R extends K8sResourceCommon | K8sResourceCommon[]> = [
-  data: R,
+  data: R | unknown,
```
florkbr, line 10: I would also like to revisit these `|`'s. We really should be returning predictable types on these hooks.

florkbr:
There are a few issues I'd like us to follow up on, however I don't feel they are blocking this PR.

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
vidyanambiar: That's a good point, Bryan. I think I introduced this in the commit that was attempting to resolve the issue with `any`  for the plural hook, but this isn't needed. I'll get rid of it.

packages/lib-utils/src/k8s/hooks/useK8sWatchResources.ts
vidyanambiar: Here's the ticket, Bryan. I created it as a story but please let me know if it's better to have it as a task.
https://issues.redhat.com/browse/HAC-1162

packages/lib-utils/src/k8s/hooks/watch-resource-types.ts
vidyanambiar: I don't believe these are required for the build to pass. I'll get rid of them.

packages/lib-utils/src/k8s/hooks/usePrevious.ts
vidyanambiar: @florkbr I've added this as an additional point in the Acceptance Criteria for the ticket here: https://issues.redhat.com/browse/HAC-801

PR #69, opened 2022-04-13
Adjust resources to properly fetch data with watch

karelhala:


packages/lib-utils/src/k8s/hooks/k8s-watcher.ts
```diff
@@ -77,14 +77,14 @@ export const getReduxData = (immutableData, resource: WatchK8sResource) => {
...
-  if (!k8sModel || !resource || !resource.name || !resource.namespace) {
+  if (!k8sModel || !resource || !k8sModel.name || !resource.namespace) {
     return null;
   }
   const query = makeQuery(
     resource.namespace,
     resource.selector,
     resource.fieldSelector,
-    resource.name,
+    k8sModel.name,
```
vidyanambiar, line 87: Why are we replacing `resource.name` with `k8sModel.name` here? Since resource.name corresponds to the name of the specific watched resource which we pass to the hook.

packages/lib-utils/src/app/api-discovery/index.ts
```diff
@@ -42,6 +43,7 @@ const defineModels = (list: APIResourceList): K8sModelCommon[] => {
       ({ name, singularName, namespaced, kind, verbs, shortNames }) =>
         <K8sModelCommon>{
           ...(apiGroup ? { apiGroup } : {}),
+          name,
```
vidyanambiar, line 46: Why do we need `name` here? Since this is the model and not a specific resource based on that model. I could have understood this wrong.

packages/lib-utils/src/k8s/hooks/k8s-watcher.ts
```diff
@@ -77,14 +77,14 @@ export const getReduxData = (immutableData, resource: WatchK8sResource) => {
 };
 
 export const getWatchData: GetWatchData = (resource, k8sModel, cluster) => {
-  if (!k8sModel || !resource || !resource.name || !resource.namespace) {
+  if (!k8sModel || !resource || !k8sModel.name || !resource.namespace) {
```
florkbr, line 80: @karelhala I think we modified this line to satisfy a Typescript error when migrating from console. This is what it is in console:
https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/src/utils/k8s/hooks/k8s-watcher.ts#L74

```tsx
export const getIDAndDispatch: GetIDAndDispatch<SDKStoreState> = (resource, k8sModel, cluster) => {
  if (!k8sModel || !resource) {
    return null;
  }
``

File: packages/lib-utils/src/k8s/hooks/k8s-watcher.ts (hunk 1 above)
Comment (florkbr) [inline packages/lib-utils/src/k8s/hooks/k8s-watcher.ts:87]: Hm.... wondering why this needed to be changed. We use the resource.name in console: https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/src/utils/k8s/hooks/k8s-watcher.ts#L81

File: packages/lib-utils/src/app/api-discovery/index.ts (hunk 2 above)
Comment (karelhala) [reply packages/lib-utils/src/app/api-discovery/index.ts:46]: Ah good point!

File: packages/lib-utils/src/k8s/hooks/k8s-watcher.ts (hunk 1 above)
Comment (karelhala) [reply packages/lib-utils/src/k8s/hooks/k8s-watcher.ts:87]: When we are getting list the name is not required and line 80 was failing. Since the name is not required for list methods I'll change it so the `if` is more aligned. Goot poin however!

File: packages/lib-utils/src/k8s/hooks/k8s-watcher.ts [hunk 4]
```diff
@@ -77,7 +77,7 @@ export const getReduxData = (immutableData, resource: WatchK8sResource) => {
 };
 
 export const getWatchData: GetWatchData = (resource, k8sModel, cluster) => {
-  if (!k8sModel || !resource || !resource.name || !resource.namespace) {
+  if (!k8sModel || !resource?.namespace) {
```
vidyanambiar, line 80: This would prevent us from watching resources that are not namespaced, correct? Is that expected? I know this is also currently the case so it's not a comment on this PR but a general question.

PR #71, opened 2022-04-13
Allow passthru param on ws.destory()

andrewballantyne:
Fix from https://github.com/openshift/console/pull/11288 propagated into the SDK. 

Effectively the issue in OpenShift Console was the `.onDestory` callback was waiting for a param that was supposed to be provided by the `.destory()` call made a few lines above. 

Kinda surprised the `GenericHandler<never>` type didn't cause a TS issue with us having a variable in the `.onDestory(timedOut)` callback 🤔


andrewballantyne:
/cc @florkbr 

This is the issue I mentioned yesterday to you.

florkbr:
:shipit:

PR #72, opened 2022-04-20
Add filtering to Table component

fhlavac:
[HAC-1012](https://issues.redhat.com/browse/HAC-1012)

filtering

Added multi-column filtering to the table component. User can apply filters to multiple columns at the same time and clear the filters. Component allows passing a custom `onFilter` component. If no `filters` array is passed, user can use the component completely without filtering - pagination will be added to it in a follow-up.

Also added comments to `Table` and `StatusBox` props.

@florkbr

### DIFF
--- a/packages/lib-utils/src/components/status/StatusBox.tsx
+++ b/packages/lib-utils/src/components/status/StatusBox.tsx
@@ -3,11 +3,14 @@ import * as _ from 'lodash-es';
 import * as React from 'react';
 
 export type LoadError = {
+  /** Error message. */
   message: string;
+  /** Error HTTP status code. */
   status: number;
 };
 
 type AccessDeniedProps = {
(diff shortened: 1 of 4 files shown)


packages/lib-utils/src/components/status/StatusBox.tsx
```diff
@@ -3,11 +3,14 @@ import * as _ from 'lodash-es';
 import * as React from 'react';
 
 export type LoadError = {
+  /* Error message. */
```
vojtechszocs, line 6: Nit: please use `/**` comment blocks, since `/*` ones won't appear in JSdoc.

fhlavac: Missed that, thank you @vojtechszocs

PR #73, opened 2022-04-20
Consume preferenceList to sort API discovery based on

karelhala:
## Watch improvements
There are 2 improvements to make watching for the k8s resources a lot faster (currently the app would have to wait for about 1-2 minutes before starting the watch functionality)

### Use prioritized list of APIs
In order to start polling for the resource sooner, before we get all resources we can pass in a list of APIs we want to query for first. This will put the prioritized APIs in the front and we'd still use batch requests not to break the API.

### Dispatch each resource to redux
Whenever we receive the resources from server we can emediately dispatch them to redux in order to unblock resource watching. So once the resource is in redux store the app can start watching for it.

### DIFF
--- a/packages/lib-utils/src/app/api-discovery/index.ts
+++ b/packages/lib-utils/src/app/api-discovery/index.ts
@@ -69,7 +69,7 @@ type APIResourceData = {
   }[];
 };
 
-const getResources = async (): Promise<DiscoveryResources> => {
+const getResources = async (preferenceList: string[]): Promise<DiscoveryResources> => {
   const apiResourceData: APIResourceData = await commonFetchJSON('/apis');
   const groupVersionMap = apiResourceData.groups.reduce(
     (acc: AnyObject, { name, versions, preferredVersion: { version } }) => {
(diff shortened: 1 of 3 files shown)


packages/lib-utils/src/app/api-discovery/index.ts
```diff
@@ -187,7 +193,7 @@ const startAPIDiscovery = () => (dispatch: DispatchWithThunk) => {
     .catch((err) => consoleLogger.error('API discovery startAPIDiscovery polling failed:', err));
 };
 
-export const initAPIDiscovery: InitAPIDiscovery = (storeInstance) => {
+export const initAPIDiscovery: InitAPIDiscovery = (storeInstance, preferenceList = []) => {
   consoleLogger.info(`API discovery waiting ${API_DISCOVERY_INIT_DELAY} ms before initializing`);
   const initDelay = new Promise((resolve) => {
     window.setTimeout(resolve, API_DISCOVERY_INIT_DELAY);
```
florkbr, line 199: @karelhala can we reduce this initial timeout? 15 seconds is too long.

karelhala: Sure thing!

PR #75, opened 2022-04-21
Export all core extensions and add nav section type

karelhala:
### Export all core extensions

Since we want to allow apps to utilize fully all types we defined, let's export them.

### New navigation section

To fully support all navigation features we need to add nav section as well.


vojtechszocs:
/lgtm

In future I'd pick types we need to expose for SDK consumers, for now we can expose everything.

vojtechszocs (approved):

PR #76, opened 2022-04-25
Enforce types - useK8sWatchResource and useK8sWatchResources

vidyanambiar:
This is a follow-on to PR https://github.com/openshift/dynamic-plugin-sdk/pull/68 and addresses 2 review comments about ~enforcing the type on the input parameter to the useK8sWatchResource hook, see https://github.com/openshift/dynamic-plugin-sdk/pull/68#discussion_r857627499~  as well as on the return type WatchK8sResult, see https://github.com/openshift/dynamic-plugin-sdk/pull/68#discussion_r857639488

### DIFF
--- a/packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
+++ b/packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
@@ -30,7 +30,7 @@ const NOT_A_VALUE = '__not-a-value__';
  * ```
  */
 export const useK8sWatchResource = <R extends K8sResourceCommon | K8sResourceCommon[]>(
-  initResource: WatchK8sResource | null,
+  initResource: WatchK8sResource,
 ): WatchK8sResult<R> => {
   const cluster = useSelector<SDKStoreState, string>((state) => getActiveCluster(state));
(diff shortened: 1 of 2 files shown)


packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
```diff
@@ -30,7 +30,7 @@ const NOT_A_VALUE = '__not-a-value__';
  * ```
  */
 export const useK8sWatchResource = <R extends K8sResourceCommon | K8sResourceCommon[]>(
-  initResource: WatchK8sResource | null,
```
andrewballantyne, line 33: This is explicitly needed if the `initResource` is dynamically calculated. 

I created a PR to specifically add it in: https://github.com/openshift/dynamic-plugin-sdk/pull/51

andrewballantyne (changes requested):

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
vidyanambiar: Thanks @andrewballantyne! Thank you for linking your PR as well. I mistakenly took this for a different change I had introduced in my PR. 
Reverted 👍

PR #79, opened 2022-04-27
Bump lib-core version after publish

vojtechszocs:
`lib-core` got a new version published so bumping the version in repo as well.


florkbr:
Karel beat me 🏎️ 🏎️ 🏎️

PR #80, opened 2022-04-27
Add pagination to the TableView

fhlavac:
https://issues.redhat.com/browse/HAC-1013

- added pagination to the TableView component
- moved filter chips to a separate component

pagination

@florkbr


packages/lib-utils/src/components/table-view/FilterChips.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const groupedFilters = Object.keys(filterValues).map((key) => (
+    <ChipGroup
+      className="dps-table-view-chips"
+      key={`group-${key}`}
+      categoryName={filters.find((item) => item.id === key)?.label}
+    >
+      <Chip
+        key={filterValues[key]}
+        onClick={(event) => {
+          event.stopPropagation();
```
florkbr, line 32: @fhlavac do we also need to use [`preventDefault` ](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)? 

We could create a common click handler for this to use in other places in a utility.

packages/lib-utils/src/components/table-view/TableView.tsx
```diff
@@ -17,22 +15,27 @@ import { omit } from 'lodash';
...
 export type TableViewProps<D> = TableProps<D> & {
-  /** Optional custom onFilter callback. */
+  /* Optional custom onFilter callback. */
   onFilter?: (filterValues: Record<string, string>, activeFilter?: FilterItem) => D[];
-  /** Optional array of filterBy options. */
+  /* Optional array of filterBy options. */
   filters?: FilterItem[];
 };
 
+const calculatePage = (limit = 10, offset = 0) => Math.floor(offset / limit) + 1;
```
florkbr, line 35: @fhlavac see my comment about using the virtualized extension.

packages/lib-utils/src/components/table-view/TableView.tsx
```diff
@@ -144,6 +162,16 @@ const TableView: React.FC<TableViewProps<Record<string, unknown>>> = ({
         emptyLabel={emptyLabel}
         aria-label={ariaLabel}
       />
+      <Pagination
```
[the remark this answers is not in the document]
fhlavac: @florkbr yeah, as I am looking into it it should. If we want infinite scrolling and managing a lot of rows, it's probably worth using it.

packages/lib-utils/src/components/table-view/filter-chips.scss
```diff
@@ -0,0 +1,15 @@
+.pf-c-chip-group.dps-table-view-chips.pf-m-category {
+  margin-bottom: var(--pf-global--spacer--sm);
+  margin-top: var(--pf-global--spacer--sm);
+  margin-right: var(--pf-global--spacer--sm);
+}
+
+.pf-c-button.pf-m-link.dps-table-view-clear {
+  margin-left: var(--pf-global--spacer--xs);
```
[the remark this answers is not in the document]
fhlavac: @florkbr yeah, the console's implementation is hiding columns from the end when you narrow the window. Default PF behavior is to display a row as a stack of cells when there is not enough space. It just depends what we want.

By default 
image
becomes:
image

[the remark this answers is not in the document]
florkbr: @fhlavac lets favor the patternfly implementation. We don't have to address it in this PR - I was just thinking about it :)

package.json
```diff
@@ -66,5 +68,8 @@
     "tslib": "^2.3.1",
     "typescript": "~4.4.4",
     "webpack": "^5.69.1"
+  },
+  "overrides": {
+    "react-dom": "^17.0.2"
```
florkbr, line 73: @fhlavac do we have an explicit reason why we are adding this override? What package is pulling in the wrong version of react-dom?

package.json
```diff
@@ -18,6 +18,7 @@
   "devDependencies": {
     "@patternfly/react-core": "^4.202.16",
     "@patternfly/react-table": "^4.71.16",
+    "@patternfly/react-virtualized-extension": "^4.53.2",
```
florkbr, line 21: @vojtechszocs are we expecting to have to add new dependencies in both the root package.json and the sub-package.json? That seems like a lot of overhead....

florkbr, line 21: @vojtechszocs are we expecting to have to add new dependencies in both the root package.json and the sub-package.json? That seems like a lot of overhead....

packages/lib-utils/src/components/table/VirtualizedTableBody.tsx
```diff
@@ -0,0 +1,129 @@
...
+  height,
+  isScrolling,
+  onChildScroll,
+  Row,
+  scrollTop,
+  width,
+}: VirtualizedTableBodyProps<D>) => {
+  const cellMeasurementCache = new CellMeasurerCache({
+    fixedWidth: true,
+    minHeight: 44,
```
florkbr, line 74: Curious where we got 44 from?

package.json
fhlavac: @florkbr 
```
➤ YN0000: @monorepo/root@workspace:. provides react-dom@npm:17.0.2 [2e9c5] with version 17.0.2, which doesn't satisfy the following requirements:

➤ YN0000: react-virtualized@npm:9.22.3 [2e9c5] → ^15.3.0 || ^16.0.0-alpha ✘
```

Unfortunately, `react-virtualized` library does not support the new version of `react-dom` (in latest stable version which is more than a year old). In the last months they don't publish stable releases for some reason :worried:

packages/lib-utils/src/components/table/VirtualizedTableBody.tsx
fhlavac: @florkbr yeah, that's a bit weird that there is such a hardcoded value.
It's used in the official example https://github.com/patternfly/patternfly-react/blob/main/packages/react-virtualized-extension/src/components/Virtualized/examples/VirtualizedTable.md
and both https://github.com/openshift/console/blob/33f02a8f454519e10fd498a3037e25c26f3d0c0a/frontend/public/components/factory/Table/VirtualizedTableBody.tsx#L44 and https://github.com/openshift/hac-dev/blob/eb1864d25c7a235838c0c1c338b8abf1b356a7fe/src/shared/components/table/VirtualBody.tsx#L47

package.json
florkbr: Hm. Alright. We may need to bring this up with the PF team - and maybe possibly contribute ourselves. CC @psav

PR #81, opened 2022-04-28
Export watch resource types

karelhala:
### Missing types

While investigating hac-dev it was discovered that some types are missing in the dynamic plugin. This PR fixes such issue by adding the most common types to exported list.


karelhala:
ping @florkbr

karelhala:
@vojtechszocs looks like I need /approve as well

PR #82, opened 2022-04-28
HAC-800: Clean up reducers

vidyanambiar:
1. Removed the Core reducer/actions as they are not currently used in hac-core or hac-dev and are not needed on the k8s get calls.
2. Removed `| undefined` on the SDK K8sState type and addressed Typescript errors thrown in AppInitSDK>combineReducers by utilizing a different function overload of `combineReducers`

This PR can be tested against hac-core's testK8s page (by running `yarn build` in `packages/lib-utils` and copying the `packages/lib-utils/dist` folder into the `@openshift/dynamic-plugin-sdk-utils` node module in `hac-core/frontend/`)


packages/lib-utils/src/app/redux/actions/k8s.ts
```diff
@@ -184,11 +182,10 @@ export const watchK8sList =
           return;
         }
 
-        const { subprotocols } = getImpersonate(getState()) || {};
         WS[id] = k8sWatch(
           k8skind,
           { ...queryWithCluster, resourceVersion },
-          { subProtocols: subprotocols, timeout: 60 * 1000 },
+          { timeout: 60 * 1000 },
```
florkbr, line 188: @vidyanambiar could we move this timeout value to a constant somewhere? Perhaps near the top of the file near `paginationLimit`? 

It might be good to also indicate paginationLimit is a constant by changing it to: `PAGINATION_LIMIT`

packages/lib-utils/src/app/redux/actions/k8s.ts
```diff
@@ -304,9 +297,7 @@ export const watchK8sObject =
       return;
     }
 
-    const { subprotocols } = getImpersonate(getState()) || {};
-
-    WS[id] = k8sWatch(k8sType, queryWithCluster, { subProtocols: subprotocols }).onBulkMessage(
+    WS[id] = k8sWatch(k8sType, queryWithCluster).onBulkMessage(
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
florkbr, line 301: @vidyanambiar since we're in this area, could you have a look and see if we could remove this Any? I think we added it temporarily and it would be nice to enforce the correct types.

florkbr: We have this same Any on a similar line above as well.

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
```diff
@@ -32,17 +31,13 @@ const NOT_A_VALUE = '__not-a-value__';
 export const useK8sWatchResource = <R extends K8sResourceCommon | K8sResourceCommon[]>(
   initResource: WatchK8sResource | null,
 ): WatchK8sResult<R> => {
-  const cluster = useSelector<SDKStoreState, string>((state) => getActiveCluster(state));
```
florkbr, line 35: Discussed with Vidya offline - stripping out the `activeCluster` logic is going to create a gap with the core SDK and the console SDK. For now we are ok with this as there is no immediate requirement from hac-core/hac-dev for this.

@vidyanambiar did we create a story under https://issues.redhat.com/browse/HAC-557?

**EDIT** -- looks like we have one already here: https://issues.redhat.com/browse/HAC-1168 for "Implementing usage of active cluster and user"

vidyanambiar: Yes @florkbr - I had created that one: https://issues.redhat.com/browse/HAC-1168

packages/lib-utils/src/app/redux/actions/k8s.ts
vidyanambiar: @florkbr Thanks. I've pushed this change up.

packages/lib-utils/src/app/redux/actions/k8s.ts
vidyanambiar: @florkbr I've addressed this comment in the new commit. Please let me know if you have any suggestions for improving how I've implemented it. Thank you!

packages/lib-utils/src/app/redux/actions/k8s.ts
```diff
@@ -62,7 +63,25 @@ const WS: { [id: string]: WebSocketFactory } = {};
...
+type EventType = { object: K8sResourceCommon };
+
+const isOfEventType = (e: MessageDataType): e is EventType => {
+  return (e as EventType).object !== undefined;
+};
+
+type EventsType = K8sEvent[] & K8sResourceCommon[];
+
+const isOfEventsType = (events: MessageDataType[]): events is EventsType => {
+  const event = (events as EventsType)[0];
```
florkbr, line 78: @vidyanambiar do we ever expect there to be more than one event? I wonder if we should future proof this function by using `_.every` or similar?

vidyanambiar: @florkbr That's a good point. I agree we should use `_.every` here. I pushed up an update 👍

PR #85, opened 2022-05-05
[HAC-1230] Tests for useK8sWatchResource(s) hooks

vidyanambiar:
This PR implements unit tests for the useK8sWatchResource(s) hooks.


packages/common/test-configs/setup-react.ts
```diff
@@ -1,14 +1,13 @@
+// This adds custom Jest matchers for working with the DOM
+// https://github.com/testing-library/jest-dom#custom-matchers
 import '@testing-library/jest-dom';
 
-jest.mock('react', () => {
-  const originalModule = jest.requireActual('react');
+import { AnyObject } from '../src/types/common';
```
karelhala, line 5: This is failing the Prow test, try importing just the `type` or rebase on top of https://github.com/openshift/dynamic-plugin-sdk/pull/91 (will be merged in a few so could be just enough to rebase on main)

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
```diff
@@ -0,0 +1,179 @@
...
+    conditions: [
+      {
+        lastTransitionTime: '2022-04-29T13:41:22Z',
+        message: 'Application has been successfully created',
+        reason: 'OK',
+        status: 'True',
+        type: 'Created',
+      },
+    ],
+    devfile:
```
florkbr, line 58: @vidyanambiar are we actually using all of these fields in the test? How did we get these objects? It looks like maybe we pulled them from the K8s responses from server?

Point being - I think we should be careful with these tests - any field that is not explicitly required in the type definition nor used in the test should be omitted.

florkbr, line 58: @vidyanambiar are we actually using all of these fields in the test? How did we get these objects? It looks like maybe we pulled them from the K8s responses from server?

Point being - I think we should be careful with these tests - any field that is not explicitly required in the type definition nor used in the test should be omitted.

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      'metadata:\\n  attributes:\\n    appModelRepository.context: /\\n    appModelRepository.url: https://github.com/redhat-appstudio-appdata/-vnambiar-touch-drive\\n    gitOpsRepository.context: /\\n    gitOpsRepository.url: https://github.com/redhat-appstudio-appdata/-vnambiar-touch-drive\\nschemaVersion: 2.1.0\\n',
+  },
+};
+
+let undefinedModelMock: K8sModelCommon | undefined;
+
+// Mock modules
+
+jest.mock('./k8s-watcher', () => ({
```
florkbr, line 67: @vidyanambiar I'll need to look at our Jest config, but we should ensure these mocks are cleared/reset between test files to prevent cross-contamination.

florkbr, line 67: @vidyanambiar I'll need to look at our Jest config, but we should ensure these mocks are cleared/reset between test files to prevent cross-contamination.

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
```diff
@@ -0,0 +1,179 @@
...
+jest.mock('react-redux', () => ({
+  useDispatch: jest.fn(() => jest.fn()),
+  useSelector: jest.fn(),
+}));
+
+describe('useK8sWatchResource', () => {
+  beforeEach(() => {
+    jest.resetModules();
+
+    (getWatchData as jest.Mock).mockReturnValue(null);
```
florkbr, line 90: @vidyanambiar I see we are using `as jest.Mock` in quite a few places. Could we perhaps use the [mocked helper](https://jestjs.io/docs/jest-object#jestmockedtitem-t-deep--false) jest provides to get the right typings instead?

florkbr, line 90: @vidyanambiar I see we are using `as jest.Mock` in quite a few places. Could we perhaps use the [mocked helper](https://jestjs.io/docs/jest-object#jestmockedtitem-t-deep--false) jest provides to get the right typings instead?

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      boolean,
+    ]);
+    (getWatchData as jest.Mock).mockReturnValue(mockWatchData);
+
+    const { result } = renderHook(() => useK8sWatchResource(watchedResourceMock));
+    const [data, loaded, error] = result.current;
+
+    expect(useSelector as jest.Mock).toHaveBeenCalled();
+    const resourceData = data as K8sResourceCommon;
+    expect(resourceData?.metadata?.creationTimestamp).toEqual('2022-04-29T13:41:21Z');
```
florkbr, line 175: @vidyanambiar we may want to consider comparing more than just the timestamps to cover all of the bases and to also avoid hard-coding the right hand side of this assertion.

We could achieve this through the [objectContaining](https://jestjs.io/docs/expect#expectobjectcontainingobject) assertion perhaps.

florkbr: Otherwise if we update the creation timestamp on line 39 this test and any other tests with the hard-coded value of `2022-04-29T13:41:21Z` will fail.

packages/lib-utils/src/k8s/hooks/useK8sWatchResources.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    (getWatchData as jest.Mock).mockReturnValue(mockWatchData);
+    (getReduxData as jest.Mock).mockReturnValue(resourceDataMock);
+
+    const { result } = renderHook(() => useK8sWatchResources(watchedResourcesMock));
+    const { application } = result.current;
+    const { data, loaded, loadError } = application;
+    expect(loaded).toEqual(true);
+    expect(loadError).toEqual('');
+    expect((data as K8sResourceCommon).metadata?.creationTimestamp).toEqual('2022-04-29T13:41:21Z');
+    expect(useSelectorSpy).toHaveBeenCalledTimes(2);
```
florkbr, line 158: Not sure how helpful it is to assert that `useSelector` is called twice here versus a simple `toHaveBeenCalled()`. If we go to add another `useSelector` call in the future this test will fail.

packages/lib-utils/src/k8s/hooks/useK8sWatchResources.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    const { data, loaded, loadError } = application;
+
+    expect(data).toMatchObject({});
+    expect(loaded).toEqual(true);
+    expect((loadError as NoModelError).message).toEqual('Model does not exist');
+  });
+
+  test('should return data for the watched resource', () => {
+    const mockWatchData: WatchData = {
+      id: 'appstudio.redhat.com~v1alpha1~Application---{"ns":"test-ns","name":"test"}',
```
florkbr, line 129: It looks like this ID is shared with the payload below. Can we de-dupe it/put it in a constant?

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
vidyanambiar: @florkbr Yes, that's right. These were pulled from k8s responses from the server (testK8s). This response data is of type `K8sResourceCommon` which combines `K8sResourceIdentifier` and a [Partial](https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/lib-utils/src/types/k8s.ts#L27) and fields like `status.devfile` that you see above are valid under the type definition for `status: { [key: string]: unknown };` within the partial. However, I can pare this down to just what I'm verifying in the test `metadata.creationTimestamp` if that's what you're referring to.

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
vidyanambiar: @florkbr We do have `clearMocks` set to `true` in the [jest config](https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/common/test-configs/jest-config-base.ts#L15) at the moment. I recall @vojtechszocs had recommended it's better to explicitly clear it in the test code. 
I have `jest.clearAllMocks` in the [afterEach](https://github.com/openshift/dynamic-plugin-sdk/pull/85/files#diff-37b388a5fd51bb4dc2241412abafe4a68c75dc9bc1fe30389fd40d7c5ab1ceb7R94) hook.  @vojtechszocs @florkbr What's your recommendation? Shall I remove the global `clearMocks: true` from the jest.config or get rid of the afterEach hook where I'm clearing mocks?

florkbr: @vidyanambiar I think we can revisit this, however my points would be:
1. `clearMocks` set to `true` in the jest config is good - however that won't remove the mock implementations themselves. That's what `resetMocks` is for.
2. It's not clear from the doc if the global mocks will be removed (IE the mocks on react/etc). We should be able to test that - IE add another test file without mocking the global modules we have in these 2 files and see if the test passes.

I personally think keeping clearMocks is a good thing (as a fallback if we fail to clean up some mocks). However I agree with @vojtechszocs that we should also clean up after ourselves in the tests.

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    const { result } = renderHook(() => useK8sWatchResource(watchedResourceMock));
+    const [data, loaded, error] = result.current;
+
+    expect(useSelectorMock).toHaveBeenCalled();
+    const resourceData = data as K8sResourceCommon;
+
+    expect(resourceData).toEqual(
+      expect.objectContaining({
+        apiVersion: expect.any(String),
```
florkbr, line 153: @vidyanambiar this is not quite what I meant - we should enforce that the values are what we expect as well. I was thinking of something a bit more dynamic. Let me check out your branch and I'll post an update

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    useK8sModelMock.mockReturnValue([resourceModelMock, false] as [K8sModelCommon, boolean]);
+    getWatchDataMock.mockReturnValue(mockWatchData);
+
+    const { result } = renderHook(() => useK8sWatchResource(watchedResourceMock));
+    const [data, loaded, error] = result.current;
+
+    expect(useSelectorMock).toHaveBeenCalled();
+    const resourceData = data as K8sResourceCommon;
+
+    expect(resourceData).toEqual(
```
florkbr, line 151: So you should be able to replace this entire expect (starting on `151`) with:

```suggestion
    expect(data as K8sResourceCommon).toMatchObject(resourceDataMock);
```

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
florkbr: See above suggestion!

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
florkbr: I also confirmed this does not effect the coverage and will be a bit more dynamic if we add/remove fields.

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
vidyanambiar: @florkbr As you suggested, I added another test file without mocking the global modules we have in these 2 test files to see if the test passes. The test does not pass and jest appears to be using the unmocked version of the module as expected. I found that this still happens even if we comment out clearAllMocks in the previous files as well as in the jest config.
So it appears that clearAllMocks applies within the context of a test file, if we need to clear the mock between individual tests in a file rather than between test files. I found some more info here. https://stackoverflow.com/questions/44141276/does-jest-automatically-restore-mocked-modules-between-test-modules#:~:text=You%20don't%20have%20to,in%20its%20own%20sandboxed%20thread. Let me know what you think.
PS: The only global mock that we seem to have right now is `react` which was added as part of [setup-react.ts](https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/common/test-configs/setup-react.ts).

packages/lib-utils/src/k8s/hooks/useK8sWatchResource.test.ts
vidyanambiar: That's a good point, @florkbr. Thanks! Addressed 👍

PR #91, opened 2022-05-11
Fix renderHook failing on React import

vojtechszocs:
This fixes the issue where `renderHook` from `@testing-library/react-hooks` would fail on
```
TypeError: Cannot read property 'createElement' of undefined
```

We use TypeScript to compile both JS and TS sources (via `"allowJs": true` option) when building our packages.

In the same fashion, we compile both JS and TS test sources (via `preset: 'ts-jest/presets/js-with-ts'` option) when running our unit tests. Specifying `__esModule: true` when mocking `react` shouldn't be necessary.

Dependency changes:
- de-bump Jest and `ts-jest` from 28 to 27
- add `@types/ts-jest` for Jest 27
- add `react-test-renderer` for React 17
- remove `jest-environment-jsdom` (Jest 27 pulls it in automatically)


vidyanambiar:
Thank you @vojtechszocs! I tested these changes with the tests for the useWatchK8sResource hook. It worked well and coverage was collected.

karelhala (approved):
/lgtm

PR #92, opened 2022-05-13
Enable code coverage reports to codecov

karelhala:


test.sh
```diff
@@ -24,3 +26,19 @@ yarn test
...
+       echo "detected PR code coverage job for #${PULL_NUMBER}"
+       REF_FLAGS="-P ${PULL_NUMBER} -C ${PULL_PULL_SHA}"
+elif [[ "${JOB_TYPE}" == "postsubmit" || "${JOB_TYPE}" == "periodic" ]]; then
+       REF_FLAGS=""
+else
+       echo "Coverage not enabled on Job Type :${JOB_TYPE}"
+fi
+
+if [[ "${JOB_TYPE}" != "local" ]]; then
+  curl -Os https://uploader.codecov.io/latest/linux/codecov
```
florkbr, line 41: Will this fail the job if we fail to upload the codecov?

test.sh
```diff
@@ -24,3 +26,19 @@ yarn test
 
 # Run end-to-end tests
 # TODO
+
+# Upload code coverage
+if [[ "${JOB_TYPE}" == "presubmit" ]]; then
+       echo "detected PR code coverage job for #${PULL_NUMBER}"
+       REF_FLAGS="-P ${PULL_NUMBER} -C ${PULL_PULL_SHA}"
+elif [[ "${JOB_TYPE}" == "postsubmit" || "${JOB_TYPE}" == "periodic" ]]; then
```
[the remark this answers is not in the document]
karelhala: We could add an env variable in the postsubmit script and check for it here. But I wouldn't spend time on it, since if we were to switch CI/CD we'd have to change this script accordingly anyway.

test.sh
karelhala: I don't think so, the token was missing previously and it just didn't published the report.

florkbr: Hm... I'm torn if we actually should fail the pipeline if the codecov report fails to upload. I guess the built binary could be fine. I just don't want to get into a situation where the published binary doesn't have an associated codecov. 

I think the codecov bot will block the PR if codecov is too low

test.sh
```diff
@@ -24,3 +26,19 @@ yarn test
 
 # Run end-to-end tests
 # TODO
+
+# Upload code coverage
+if [[ "${JOB_TYPE}" == "presubmit" ]]; then
+       echo "detected PR code coverage job for #${PULL_NUMBER}"
```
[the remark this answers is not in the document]
florkbr: @karelhala ☝️

test.sh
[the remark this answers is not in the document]
vojtechszocs: I agree with @florkbr to put the Prow CI workflow specific bits into a separate script, but that can be done as a follow-up.

test.sh
vojtechszocs: @florkbr This shell script (and hopefully any other scripts too) uses the so-called [Bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/)

```bash
set -euo pipefail
```

> The `set -e` option instructs bash to immediately exit if any command has a non-zero exit status.

So if the above `curl` command (or any other command) fails with non-zero exit code, this script as a whole will fail with that exit code. IMO this is a very sensible default for user scripts.

> Hm... I'm torn if we actually should fail the pipeline if the codecov report fails to upload.

We may treat the commands inside the if-block `[[ "${JOB_TYPE}" != "local" ]]` as ones that could potentially fail without this script failing as a result. I'd do this later as a follow-up change.

vojtechszocs:
/approve

LGTM

florkbr:
I agree with @vojtechszocs's comments above. @karelhala can we follow up? For now this LGTM.

PR #93, opened 2022-05-16
[HAC-1295] Add basic feature flag support

vidyanambiar:
This PR introduces support for feature flags in the Core SDK.

- Extensions will be able to specify feature flags within the optional `flags` attribute. (At the moment we are not distinguishing between flags that are app-specific and flags that may be managed by external services like Unleash.)
- The extensions in use will be determined based on the feature flags specified as `required` and `disallowed`.
- Consumer apps will need to implement the `isFeatureFlagEnabled` function (see `packages/lib-core/src/store/PluginStore.ts`). 
- The function `updateExtensions` should be called by the consumer app/plugin whenever a change is detected in any feature flags that may be used by plugins to gate their extensions.

### DIFF
--- a/packages/lib-core/src/store/PluginStore.ts
+++ b/packages/lib-core/src/store/PluginStore.ts
@@ -224,7 +224,7 @@ export class PluginStore implements PluginConsumer, PluginManager {
    * @param {Extension} extension The extension to check for
    * @returns {boolean} returns `true` if the extension is in use, and `false` if it is not in use
    */
-  isExtensionInUse(extension: Extension) {
+  private isExtensionInUse(extension: Extension) {
     return (
       (extension.flags?.required?.every((f) => this.options.isFeatureFlagEnabled(f)) ?? true) &&
       (extension.flags?.disallowed?.every((f) => !this.options.isFeatureFlagEnabled(f)) ?? true)
--- a/packages/lib-core/src/store/PluginStore.ts
+++ b/packages/lib-core/src/store/PluginStore.ts
@@ -210,7 +210,8 @@ export class PluginStore implements PluginConsumer, PluginManager {
     const prevExtensions = this.extensions;
 
     this.extensions = Array.from(this.loadedPlugins.values()).reduce(
-      (acc, p) => (p.enabled ? [...acc, ...p.extensions.filter(this.isExtensionInUse)] : acc),
+      (acc, p) =>
+        p.enabled ? [...acc, ...p.extensions.filter(_.bind(this.isExtensionInUse, this))] : acc,
       [] as LoadedExtension[],
     );
 
--- a/packages/lib-core/src/store/PluginStore.ts
+++ b/packages/lib-core/src/store/PluginStore.ts
@@ -13,6 +13,8 @@ export type PluginStoreOptions = Partial<{
   autoEnableLoadedPlugins: boolean;
   /** Post-process loaded extension objects before adding associated plugin to the {@link PluginStore}. */
   postProcessExtensions: (extensions: LoadedExtension[]) => LoadedExtension[];
+  /** Determine whether the given feature flag is currently enabled. */
+  isFeatureFlagEnabled: (flag: string) => boolean;
 }>;
 
 /**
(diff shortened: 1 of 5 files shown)


packages/lib-core/src/yup-schemas.ts
```diff
@@ -41,13 +41,34 @@ const extensionTypeSchema = yup
...
+ * Examples:
+ *
+ * ```
+ * FOO, FOO_BAR
+ * ```
+ */
+const featureFlagNameSchema = yup
+  .string()
+  .required()
+  .matches(/^[A-Z]+[A-Z_]*$/, 'feature flag name');
```
christianvogt, line 56: Do we not allow numbers in the flag name?

packages/lib-core/src/store/PluginStore.ts
```diff
@@ -195,20 +196,41 @@ export class PluginStore implements PluginConsumer, PluginManager {
...
+      this.invokeListeners(PluginEventType.ExtensionsChanged);
     }
   }
 
+  /**
+   * Checks whether an extension is in use based on the values of required and disallowed feature flags
+   * @param {Extension} extension The extension to check for
+   * @returns {boolean} returns `true` if the extension is in use, and `false` if it is not in use
+   */
+  isExtensionInUse(extension: Extension) {
```
christianvogt, line 227:
```suggestion
  private isExtensionInUse(extension: Extension) {
```

packages/lib-core/src/types/store.ts
```diff
@@ -81,4 +81,15 @@ export type PluginManager = {
...
+  /**
+   * Update extensions which are currently in use.
+   *
+   * This function should be called whenever a change is detected in any feature flags that
+   * may be used by plugins to gate their extensions, including:
+   * - flags managed by the host application
+   * - flags contributed by plugins
+   * - flags managed by external services like Unleash
+   */
+  updateExtensions: () => void;
```
christianvogt, line 94: Are there other reasons to call `updateExtensions` instead of flags? 
Keeping the name generic is probably a good call. But if it's only for flags we can be more specific.

christianvogt (commented):
overall the change seems good for supporting feature flags

packages/lib-core/src/types/store.ts
vidyanambiar: We originally had this function as `featureFlagChanged` and then we renamed it to `updateExtensions` as we're effectively triggering an update of extensions when the flags change. Since feature flag change is the only use case to trigger these updates (at least for now), maybe we could rename it to something like `handleFeatureFlagUpdate` if that makes it clearer? What do you think @cvogt @vojtechszocs ?

packages/lib-core/src/yup-schemas.ts
vojtechszocs: Sounds reasonable, let's allow numbers as well

```diff
-  .matches(/^[A-Z]+[A-Z_]*$/, 'feature flag name');
+  .matches(/^[A-Z]+[A-Z0-9_]*$/, 'feature flag name');
```

packages/lib-core/src/store/PluginStore.ts
vojtechszocs: I'm OK with this change, but later on we'll want to unit test this method :smiley: 

Note that TypeScript `private` modifier is not the same as ES [private class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) which we currently don't use.

packages/lib-core/src/types/store.ts
vojtechszocs: We called it `updateExtensions` because it effectively forces the `PluginStore` to recompute the list of extensions which are currently in use.

We avoided adding a method like `setFeatureFlags: (flags: { [name: string]: boolean }) => void` since this would force the host application to store all flags in a single object that would be passed to this method. Instead, the `PluginStore` will use the `isFeatureFlagEnabled: (flag: string) => boolean` option to determine the value of the given flag when needed.

vojtechszocs: I'd keep it as `updateExtensions` - it's more generic and future proof :smiley: but yeah, right now, a change in feature flags is basically the only thing that should trigger this function.

christianvogt: I'm good with this.

florkbr: FYI @vidyanambiar will be following up on this PR based on our discussion here:  https://coreos.slack.com/archives/C02EQERSV3N/p1653499298943189 (issue https://issues.redhat.com/browse/HAC-1406). For now adding my LGTM

florkbr:
Confirmed with Vidya these changes work in hac-core as-is. We still have some more tweaking to do - but for now this is a good start.

PR #100, opened 2022-05-23
Implement basic publish script for the SDK lib packages

florkbr:
Rudimentary publish script that will publish the lib packages. The following criteria must be met (up for debate):
1. A PR must be open and the label `publish-sdk` present on the PR
2. The PR must bump the version of the lib in the package.json it wishes to have published.

The PR will then execute a dry-run publish in the presubmit, and an actual publish in the post-submit.

I've also included a `FORCE_PUBLISH` env variable for convenience of the SDK maintainers if manual local publishing is necessary. This bypasses the PR label check and the job type check(s).


publish.sh
```diff
@@ -0,0 +1,42 @@
...
+    echo "$PACKAGE_NAME NPM_VERSION: $NPM_VERSION"
+    GIT_VERSION=$(jq -r .version < ./package.json)
+    echo "$PACKAGE_NAME GIT_VERSION: $GIT_VERSION"
+    if [[ "${NPM_VERSION}" == "${GIT_VERSION}" ]]; then
+      echo 'NPM version and GIT version identical - nothing to do'
+    else
+      echo "publishing ${PACKAGE_NAME} version ${GIT_VERSION}"
+      npm publish --no-git-tag-version "$DRY_RUN"
+    fi
+    cd ../../
```
andrewballantyne, line 38: Probably better to hold onto `cwd` before the loop and just reset back to where you started rather than some partial path... upkeep of this may be annoying in it's current state.

test.sh
```diff
@@ -42,3 +42,6 @@ if [[ "${JOB_TYPE}" != "local" ]]; then
   chmod +x codecov
   ./codecov -t ${CODECOV_TOKEN} -r "openshift/dynamic-plugin-sdk" ${REF_FLAGS} --dir ./coverage
 fi
+
+# Publish
+./publish.sh
```
andrewballantyne, line 47: I use `./test.sh` to verify everything... will this will prompt a publish to npm? I don't have access, but will it ask me for npm creds?

andrewballantyne (changes requested):
Doesn't a code change need to happen before we can publish?

test.sh
florkbr: Agreed. I can make these updates. I was primarily focused on the path to publishing first.

publish.sh
andrewballantyne: 🤯  TIL.

publish.sh
```diff
@@ -0,0 +1,42 @@
...
+PACKAGE_GLOB=(./packages/lib-*)
+NPM_REGISTRY="https://registry.npmjs.org/"
+NPM_REGISTRY_AUTH_KEY="//registry.npmjs.org/:_authToken"
+LABELS=$(curl -s \
+  -H "Accept: application/vnd.github.v3+json" \
+  "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/pulls/${PULL_NUMBER}" \
+  | jq '.labels | map(.name)')
+HAS_PUBLISH=$(printf '%s\n' "${LABELS}" | jq 'index("'${LABEL_PUBLISH}'")')
+
+if [[ "${HAS_PUBLISH}" != "null" ]]; then
```
karelhala, line 14: This will always be non-null isn't it?

publish.sh
```diff
@@ -0,0 +1,42 @@
...
+  echo 'publish NPM packages'
+  if [[ "${JOB_TYPE}" != "postsubmit" ]]; then
```
karelhala, line 16: Do we even want to run the publish on non-postsubmit jobs? The dry run could be usefull, but I am a bit worried about the length of this task to run.

test.sh
vojtechszocs: Same as @andrewballantyne - I also use `test.sh` to verify my changes locally, the publish can be triggered when we are in CI env.

publish.sh
florkbr: @karelhala since this is gated by the logic above checking if the PR has the label `publish-sdk`, it will only run for the pre and post submits for PRs that include the `publish-sdk` label. Or am I missing something?

publish.sh
florkbr: Yes - specifically jq returns a string `null` when it cannot find the index.

publish.sh
florkbr: Done! Switched to `pushd` and `popd`

test.sh
florkbr: Updated so this script will not impact local runs of `./test.sh` (aside from a single line of output: `skipping publish npm packages`)

florkbr:
Bringing this PR back from the dead ☠️

publish.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+set -euo pipefail
+
+JOB_TYPE=${JOB_TYPE:-local}
```
florkbr, line 4: TODO (follow up if possible): document how to use this script in https://github.com/openshift/dynamic-plugin-sdk/blob/main/docs/publish-packages.md from a developers perspective (pull request and adding the `publish-sdk` label) and from a maintainer perspective (manual publish and using `FORCE_PUBLISH`).

vojtechszocs, line 4: Nit: can we add a link to Prow docs (comment `# https://..`) where these env. variables are documented?

publish.sh
```diff
@@ -0,0 +1,49 @@
...
+REPO_NAME=${REPO_NAME:-dynamic-plugin-sdk}
+PULL_NUMBER=${PULL_NUMBER:-0}
+NPM_TOKEN=${NPM_TOKEN:-notoken}
+FORCE_PUBLISH=${FORCE_PUBLISH:-false}
+
+LABEL_PUBLISH="publish-sdk"
+PACKAGE_GLOB=(./packages/lib-*)
+NPM_REGISTRY="https://registry.npmjs.org/"
+NPM_REGISTRY_AUTH_KEY="//registry.npmjs.org/:_authToken"
+LABELS=$(curl -s \
```
vojtechszocs, line 15: This triggers HTTP call to GitHub API even if we run the script locally as part of manual `test.sh` execution.

Can we make this HTTP call conditional, i.e. use an empty array if `PULL_NUMBER` is not `0` ?

publish.sh
```diff
@@ -0,0 +1,49 @@
...
+  -H "Accept: application/vnd.github.v3+json" \
+  "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/pulls/${PULL_NUMBER}" \
+  | jq '.labels? | map(.name)?')
+HAS_PUBLISH=$(printf '%s\n' "${LABELS}" | jq 'index("'${LABEL_PUBLISH}'")')
```
vojtechszocs, line 19: You can also use Bash `<<<` operator that redirects literal string to stdin, for example:

```sh
jq 'index("'${LABEL_PUBLISH}'")' <<< "${LABELS}"
```

publish.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if [[ ( "${JOB_TYPE}" != "local" && "${HAS_PUBLISH}" != "null" ) || "${FORCE_PUBLISH}" == "true" ]]; then
+  echo 'publish npm packages'
+  if [[ "${JOB_TYPE}" != "postsubmit" && "${FORCE_PUBLISH}" != "true" ]]; then
+    echo 'non postsubmit job detected - performing dry run'
+    DRY_RUN="--dry-run"
```
vojtechszocs, line 25: Instead of `DRY_RUN`, I'd suggest using `PUBLISH_ARGS` initialized like
```sh
PUBLISH_ARGS=""
```
above the whole `if` block, and update it here like so:
```sh
PUBLISH_ARGS="${PUBLISH_ARGS} --dry-run"
```

publish.sh
```diff
@@ -0,0 +1,49 @@
...
+  else
+    DRY_RUN=""
+  fi
+  npm set registry $NPM_REGISTRY
+  npm set $NPM_REGISTRY_AUTH_KEY $NPM_TOKEN
```
vojtechszocs, line 30: We can add an `if` block here that checks if `NPM_TOKEN` has the right value, i.e. not equal to `notoken`.

publish.sh
```diff
@@ -0,0 +1,49 @@
...
+    if [[ "${NPM_VERSION}" == "${GIT_VERSION}" ]]; then
+      echo 'npm version and git version identical - nothing to do'
+    else
+      echo "publishing ${PACKAGE_NAME} version ${GIT_VERSION}"
+      npm publish --no-git-tag-version "$DRY_RUN"
+    fi
+    popd > /dev/null
+  done
+else
+  echo "skipping publish npm packages"
```
vojtechszocs, line 48:
```suggestion
  echo "skipping publish of npm packages"
```

test.sh
```diff
@@ -27,3 +27,6 @@ yarn test
 
 # Upload code coverage
 ./prow-codecov.sh 2>/dev/null
+
+# Publish
```
vojtechszocs, line 31:
```suggestion
# Publish packages to npm registry
```

vojtechszocs (changes requested):

florkbr:
@vojtechszocs changes made! Please re-review

florkbr:
@vojtechszocs @karelhala @Hyperkid123 with our push to release the SDK/HAC I'd like to revisit this to streamline updates in the future and reduce the manual steps to publish the SDK. Mind having a look?

PR #102, opened 2022-05-24
Add feature flag example to the sample plugin and app

vidyanambiar:
This PR builds on top of https://github.com/openshift/dynamic-plugin-sdk/pull/107. It introduces a very basic example of a feature flag "TELEMETRY_FLAG" that is used to gate the "telemetry/listener" extension in the sample plugin.

Screen Shot 2022-05-26 at 8 42 10 AM
Screen Shot 2022-05-26 at 8 41 56 AM


florkbr:
Synced up with Vidya offline - seeing an issue when enabling/disabling the flag via the button. There is an error in the console and the button will not allow me to re-enable the flag:
image

florkbr:
@vidyanambiar this is working for me now! Though I think we accidentally checked in a screenshot... Can we remove it?

PR #104, opened 2022-05-25
[HAC-1170] Initial Cypress configuration to be able to run e2e tests

vidyanambiar:
To run tests:
1. `yarn install`
2. Start the servers for the sample plugin and the sample app:
`npm run http-server`  within `packages/sample-app` and `packages/sample-plugin`
3. `npm run e2e-test`

Some considerations:
- The cypress documentation recommends the `cypress/*` folder structure I went with it but we can change it if necessary and point the configuration files to our custom directory names.
- I left some boilerplate code in related to cypress `support` and `plugins` that can help us extend the internal behavior of Cypress. These folders are auto-generated when we run cypress unless we explicitly disable them in the Cypress configuration. The only reason I kept this is so that we know it's an option to use going forward, but I can disable them too and we can enable these features as needed going forward. @florkbr what do you think?
- I intentionally disabled the setting for capturing videos for now because I suspected that would probably trigger the out of memory issue we'd seen earlier on Prow. We can turn this on if needed once we have more tests (Screenshots get captured by default on errors)


packages/common/test-configs/cypress.json
```diff
@@ -0,0 +1,5 @@
+{
+  "baseUrl": "http://localhost:9000",
+  "video": false,
+  "defaultCommandTimeout": 5000
+}
```
yapei, line 5: Shall we move `cypress.json` into `cypress` folder?

cypress/integration/sample.spec.ts
```diff
@@ -0,0 +1,10 @@
+// <reference types="cypress" />
+
+describe('example to-do app', () => {
+  it('displays two todo items by default', () => {
+    cy.visit('/app-minimal.html');
+    cy.get('div.pf-c-toolbar__item > button.pf-c-button')
+      .first()
+      .should('have.text', 'Load plugin');
+  });
+});
```
yapei, line 10: normally we may want put tests files in `tests` folder, with more tests are added we may want another folder for common selectors

packages/common/test-configs/cypress.json
yapei: If we put `cypress.json` in `cypress` folder we may need to explicitly set `integrationFolder` in the configuration file

package.json
```diff
@@ -13,7 +13,9 @@
     "test": "jest --collectCoverage",
     "test-print-config": "jest --showConfig",
     "eslint": "eslint --max-warnings 0 --ext .js,.jsx,.ts,.tsx",
-    "eslint-print-config": "eslint --print-config"
+    "eslint-print-config": "eslint --print-config",
+    "e2e-test": "cypress run --config-file packages/common/test-configs/cypress.json",
+    "e2e-test:open": "cypress open --config-file packages/common/test-configs/cypress.json"
```
yapei, line 18: if `cypress.json` is placed in the same folder we may not need specify `--config-file` param

yapei:
shall we add support to trigger e2e tests upon every PR creation?

packages/common/test-configs/cypress.json
vidyanambiar: @florkbr had suggested having the configurations in the `packages/common` directory where all the test configurations are. What are your thoughts Bryan?
Cypress would normally look for `cypress.json` by default in the root directory from where we trigger the cypress run. @yapei any particular reason for having it within the `cypress` directory?

vidyanambiar:
> shall we add support to trigger e2e tests upon every PR creation?

@yapei This story and PR are specifically for the cypress configuration to enable us to start writing E2E tests. Triggering the e2e tests will require some more discussion and will be covered in https://issues.redhat.com/browse/HAC-1412.

packages/common/test-configs/cypress.json
yapei: Ah yeah, you are right @vidyanambiar `cypress.json` should be in root directory where we trigger the tests, it still make sense if we have other considerations

cypress/integration/sample.spec.ts
yapei: maybe it is enough to replace `integration` with `tests`?

packages/common/test-configs/cypress.json
vidyanambiar: @yapei Is your recommendation to move it into the root directory then? 
FYI @florkbr

yapei: @vidyanambiar  I'm ok to leave it in `packages/common` if you have talked about it : )

cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
...
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
+ */
+// eslint-disable-next-line no-unused-vars
+module.exports = (on, config) => {
```
florkbr, line 19: I think it's ok for the time being to keep this boiler plate - @vidyanambiar we should have a look at any plugins that we use in OCP Console, HAC Core, ConsoleDot etc to see if we could benefit from using them or not.

package.json
florkbr: @vidyanambiar a small cleanup suggestion to keep our scripts consistent:
1. Use `-` in favor of `:` delimiter
2. Can we update the E2E scripts to start with `test-e2e` so they appear closer alphabetically to the other test scripts?

cypress/tests/sample.spec.ts
```diff
@@ -0,0 +1,10 @@
+// <reference types="cypress" />
+
+describe('example to-do app', () => {
```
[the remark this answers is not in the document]
vidyanambiar: @florkbr Sure, we can discuss that. In this case we need both the sample plugin and the sample app to be running for the E2E tests and the tests also rely on both these packages since the plugin is integrated into the app. We may also need a script to start those servers up first. I went with the recommended directory structure from cypress as an initial step but we can change it up as needed.

cypress/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["es5", "dom"],
+    "types": ["cypress"]
```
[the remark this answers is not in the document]
vidyanambiar: @florkbr I'm not seeing any common options between this config and the files under https://github.com/openshift/dynamic-plugin-sdk/tree/main/packages/common/tsconfig-bases. The only common field is "include" but the path would be different.

cypress/plugins/index.js
vidyanambiar: @florkbr I see hac-dev [here](https://github.com/openshift/hac-dev/blob/main/integration-tests/plugins/index.ts) using `plugins/index.js` to set some environment variables in the cypress configuration to be used in test utils. I found some examples of it under 3 plugins in `frontend/packages/*` in the console repo where in addition to setting cypress environment variables, some cypress tasks are set up and there is customization around how the test code is transpiled using cypress/webpack-preprocessor. I don't believe we need these just yet, but we might come across the need for env variables or common tasks as we go through the process of writing specs and then we can update the plugins file. WDYT?

cypress/tsconfig.json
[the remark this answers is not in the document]
vojtechszocs: We could use `app-react-esm.json` as the base config here, it's already used by both sample plugin and sample app packages.

This ^^ config basically represents a TS web project targeting reasonably recent web browsers (`es2020`).

I don't think we need an explicit `lib` setting (`dom` should be included by default, but I might be wrong).

Also I think we can skip the `types` setting since the default is very reasonable

> By default all visible `”@types”` packages are included in your compilation.

```json
{
  "extends": "@monorepo/common/tsconfig-bases/app-react-esm.json",
  "compilerOptions": {
    "outDir": "dist"
  },
  "include": ["**/*.ts"],
  "exclude": ["node_modules"]
}
```

cypress/plugins/index.js
vidyanambiar: > Also, some of these files are empty or do nothing, do we really need them?

@vojtechszocs Yes, as mentioned above they just include some boilerplate code for now. I see plugins used in other projects so I just left these in case we want to build them out more as we start writing tests. I can remove the files if they feel unnecessary at this point. LMK

vidyanambiar:
@vojtechszocs I've pushed up a commit to address your comments. The reason I kept `cypress.json` in the `e2e-sample-app` package rather than `e2e-common` is that `cypress.json` does not have an `extends` option so it seemed like it's better to place it in the package that triggers the e2e tests and this way we can specify package specific configs like baseUrl and screenshots directory. Let me know if you have any concerns with this.
Please re-review . Thanks!
cc: @florkbr

florkbr:
This PR looks good to me. I checked it out and the E2E seems to pass locally:
image

We will sync as a team this week to decide where we want our E2E to live.

vidyanambiar:
@florkbr Thanks, Bryan. It appears that this is an open issue still with cypress not providing a clean way to disable the auto-generation of directories - https://github.com/cypress-io/cypress/issues/501 (duplicate issue https://github.com/cypress-io/cypress/issues/619 is still open).
Since I don't see a use for the fixtures directory as yet, I've pushed up a commit to disable it in the config for now. So you should no longer see that generated directory for `test-e2e-open`.

packages/e2e-common/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  extends: [
+    'plugin:@monorepo/eslint-plugin-internal/react-typescript-prettier',
+    'plugin:cypress/recommended',
```
vojtechszocs, line 4: In future, we can add a new ESLint preset to `packages/eslint-plugin-internal/src/index.js` e.g. `react-typescript-prettier-cypress` that would allow us to define any Cypress ESLint rule overrides if we need them.

PR #105, opened 2022-05-26
add core.redux-provider extension and support

christianvogt:
fixes: https://issues.redhat.com/browse/HAC-1296

New extension `core.redux-provider`.
Provides a configuration for establishing a Redux instance scoped for the contributing plugin.

`ReduxExtensionProvider` component may be used by applications to support redux providers at the root.

Each redux provider will be scoped by the extension with a unique store and context.

Usage of redux apis must supply the context to gain access to the respective store. eg `createSelectorHook(myContext);`

### DIFF
--- a/packages/lib-utils/src/app/redux/ReduxExtensionProvider.test.tsx
+++ b/packages/lib-utils/src/app/redux/ReduxExtensionProvider.test.tsx
@@ -1,0 +1,118 @@
+import type {
+  ReduxProvider,
+  ResolvedExtension,
+  LoadedExtension,
+} from '@openshift/dynamic-plugin-sdk';
+import { useResolvedExtensions } from '@openshift/dynamic-plugin-sdk';
+import { render } from '@testing-library/react';
+import * as React from 'react';
+import type { ReactReduxContextValue } from 'react-redux';
+import { createSelectorHook } from 'react-redux';
+import type { Store } from 'redux';
+import { createStore } from 'redux';
+import ReduxExtensionProvider from './ReduxExtensionProvider';
+
+type LRReduxReducer = LoadedExtension<ResolvedExtension<ReduxProvider>>;
+
+jest.mock('@openshift/dynamic-plugin-sdk', () => ({
+  useResolvedExtensions: jest.fn(),
+}));
+
+const useResolvedExtensionsMock = jest.mocked(useResolvedExtensions, false);
+
+describe('ReduxExtensionProvider', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('should register multiple providers with separate redux contexts', () => {
+    type State = {
+      value: number;
+    };
+
+    const store1: Store = createStore((state) => state ?? { value: 1 });
+    const store2: Store = createStore((state) => state ?? { value: 2 });
+
+    // create 2 core.redux-provider extensions
+    const extensions: LRReduxReducer[] = [
+      {
+        uid: '1',
+        pluginName: 'test',
+        type: 'core.redux-provider',
+        properties: {
+          store: store1,
+          context: React.createContext<ReactReduxContextValue>({
+            store: store1,
+            storeState: store1.getState(),
+          }),
+        },
+      },
+      {
+        uid: '2',
+        pluginName: 'test',
+        type: 'core.redux-provider',
+        properties: {
+          store: store2,
+          context: React.createContext<ReactReduxContextValue>({
+            store: store2,
+            storeState: store2.getState(),
+          }),
+        },
+      },
+    ];
+
+    // mock that returns the test extensions
+    useResolvedExtensionsMock.mockReturnValue([extensions, true, []]);
+
+    // create 2 contextual redux selectors; one for each extension
+    const useSelector1 = createSelectorHook<State>(extensions[0].properties.context);
+    const useSelector2 = createSelectorHook<State>(extensions[1].properties.context);
+
+    let value1 = -1;
+    let value2 = -1;
+
+    // this component will select values separately from each redux store
+    const Test: React.FC = () => {
+      value1 = useSelector1((state) => state.value);
+      value2 = useSelector2((state) => state.value);
+      return null;
+    };
+
+    render(
+      <ReduxExtensionProvider>
+        <Test />
+      </ReduxExtensionProvider>,
+    );
+
+    // assert that the values received by each selector relate to their corresponding redux state
+    expect(value1).toBe(1);
+    expect(value2).toBe(2);
+  });
+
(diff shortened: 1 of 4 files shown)


packages/lib-utils/src/app/redux/ReduxExtensionProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+      {reduxProviderExtensions.reduce(
+        (c, e) => (
+          <Provider store={e.properties.store} context={e.properties.context}>
+            {c}
+          </Provider>
+        ),
+        children,
+      )}
+    </>
+  ) : null;
```
vojtechszocs, line 24: There are currently two scenarios which may cause extensions to be put into or out of use at any point in time:
- change in a plugin's enabled status
- change in the feature flags, which in turn impacts the availability of individual extensions

Extensions in use are computed as all extensions from plugins that are enabled, which also meet the feature flag criteria according to the current feature flag values.

Therefore,
```ts
useResolvedExtensions<ReduxProvider>(isReduxProvider)
```
may cause this hook to be re-run multiple times, providing different results on each run.

I think we need to guard against this, otherwise we may "un-render" and then "re-render" a Redux `Provider` which would effectively reset it (?)

packages/lib-utils/src/app/redux/ReduxExtensionProvider.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+jest.mock('@openshift/dynamic-plugin-sdk', () => ({
+  useResolvedExtensions: jest.fn(),
+}));
+
+const useResolvedExtensionsMock = jest.mocked(useResolvedExtensions, false);
+
+describe('ReduxExtensionProvider', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
```
vojtechszocs, line 25: I'd suggest to do any mock clear or reset in `beforeEach` instead of `afterEach`.

Note that we already have
```js
clearMocks: true,
```
in `packages/common/test-configs/jest-config-base.ts` file.

I personally prefer to be specific about how and when a test suite clears or resets its mocks, like it's done above, instead of a global `clearMocks: true` setting (which I'd like to remove in the future).

packages/lib-utils/src/app/redux/ReduxExtensionProvider.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+    // mock return value of empty set of extensions
+    useResolvedExtensionsMock.mockReturnValue([[], true, []]);
+
+    const { container } = render(
+      <ReduxExtensionProvider>
+        <span id="test" />
+      </ReduxExtensionProvider>,
+    );
+
+    expect(container.firstElementChild?.tagName).toBe('SPAN');
```
vojtechszocs, line 102: Just wondering, does the tag name have to be uppercase?

vojtechszocs (changes requested):

packages/lib-utils/src/app/redux/ReduxExtensionProvider.tsx
```diff
@@ -0,0 +1,27 @@
+import { useResolvedExtensions, isReduxProvider } from '@openshift/dynamic-plugin-sdk';
+import type { ReduxProvider } from '@openshift/dynamic-plugin-sdk';
+import * as React from 'react';
+import { Provider } from 'react-redux';
+
+/**
+ * Consumes the core.redux-
```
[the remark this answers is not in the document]
christianvogt: whoops, i guess i never finished my thought lol

packages/lib-utils/src/app/redux/ReduxExtensionProvider.test.tsx
christianvogt: yes, `tagName` is always upper case

packages/lib-utils/src/app/redux/ReduxExtensionProvider.tsx
christianvogt: The context and store should be the same instance on a re-render. So even if the provider was re-rendered, those who rely on the redux store would still be accessing the same redux context and store.

packages/lib-utils/src/app/redux/ReduxExtensionProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+ */
+const ReduxExtensionProvider: React.FC = ({ children }) => {
```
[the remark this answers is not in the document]
christianvogt: Is that a convention we're following? I like provider because those who use redux are familiar with the setup component named `Provider`.

packages/lib-utils/src/app/redux/ReduxExtensionProvider.tsx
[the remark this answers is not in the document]
florkbr: @christianvogt FYI @vojtechszocs is OOO for the remainder of the week. Were you going to address these comments? I can re-review so we can try to get this in.

christianvogt:
Addressed review comments except the renaming of `ReduxExtensionProvider`. Unless there's a convention we're following, the name currently aligns with redux api.

packages/lib-utils/src/app/redux/ReduxExtensionProvider.tsx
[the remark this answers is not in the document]
vojtechszocs: Not a convention, I'm OK with the current name - just realized that the `Provider` suffix means this component renders a Redux `Provider` component for each extension instance.

(I've based my original suggestion around the fact this component handles/interprets `ReduxProvider` extensions, but I agree that `Provider` suffix sounds more familiar to those already using Redux.)

packages/lib-utils/src/app/redux/ReduxExtensionProvider.tsx
vojtechszocs: One more question, if a `ReduxProvider` extension is put out of use (due to change in flags or due to the owner plugin getting disabled) but some component down in the tree still attempts to access the associated context or store, the responsibility to handle this is on the consuming component, correct?

christianvogt: Correct. If a nested component relies on redux and the provider is not present, it's their responsibility. Under normal circumstances the system will ensure its presence in time for consumption.

A redux provider that is disabled by a flag, while possible, this scenario is less likely to happen as the presence would have a no-op effect if unused and therefore using a flag wouldn't have any real benefit.

If the plugin is disabled, then all contributions from said plugin would also be disabled and therefore there would be no reliance on the provider.

PR #107, opened 2022-05-27
Utility for managing feature flags - HAC-1406

vidyanambiar:


packages/lib-core/src/store/PluginStoreContext.tsx
```diff
@@ -31,3 +31,12 @@ export const usePluginStore = (): PluginConsumer & PluginManager => {
...
 };
+
+/**
+ * React hook that provides access to the feature flags.
+ */
+export const useFeatureFlags = (): [FeatureFlags, (newFeatureFlags: FeatureFlags) => void] => {
+  const store = usePluginStore();
+
+  return [store.getFeatureFlags(), store.setFeatureFlags];
+};
```
christianvogt, line 42: Often within a Component we're only looking at a single feature flag.
An API that simplifies this, similar to `useState` may be best?

```
const [flag, setFlag] = useFeatureFlag('MY_FLAG');
...
setFlag(true);
```

vs
```
const [flags, setFeatureFlag] = useFeatureFlags();
const flag = flags['MY_FLAG'];
...
setFeatureFlags({ 'MY_FLAG': true });
```

WDYT?
cc @vojtechszocs

vidyanambiar: @christianvogt That's a good point. If it's more common for a single flag to be manipulated at a time, I agree that useFeatureFlag makes more sense. Do you foresee apps needing access to all the available feature flags at some point? in which case we could probably keep both hooks - for manipulating a single flag as well as multiple flags?
Let me know what you think @christianvogt @vojtechszocs. Thanks!

christianvogt: Flag usage is very often specific to individual checks. I don't see the need for a single component wanting access to all flags at once. There will be instances where more than one flag is checked at a time, in which case multiple uses of the hook will be sufficient.

packages/lib-core/src/store/PluginStore.ts
```diff
@@ -227,11 +228,28 @@ export class PluginStore implements PluginConsumer, PluginManager {
...
   private isExtensionInUse(extension: Extension) {
     return (
-      (extension.flags?.required?.every((f) => this.options.isFeatureFlagEnabled(f)) ?? true) &&
-      (extension.flags?.disallowed?.every((f) => !this.options.isFeatureFlagEnabled(f)) ?? true)
+      (extension.flags?.required?.every((f) => this.featureFlags[f]) ?? true) &&
+      (extension.flags?.disallowed?.every((f) => !this.featureFlags[f]) ?? true)
     );
   }
 
+  setFeatureFlag = (flagName: string) => (newFlagValue: boolean) => {
```
christianvogt, line 236: I'd prefer an API that takes 2 params. The hook can memoize a setter function.
@vojtechszocs  wdyt?

packages/lib-core/src/store/PluginStoreContext.tsx
```diff
@@ -31,3 +31,19 @@ export const usePluginStore = (): PluginConsumer & PluginManager => {
...
+ *
+ * @example
+ * ```ts
+ * const [flag, setFlag] = useFeatureFlag('MY_FLAG');
+ * ...
+ * setFlag(true);
+ * ```
+ */
+export const useFeatureFlag = (flagName: string): [boolean, (newValue: boolean) => void] => {
+  const store = usePluginStore();
```
vojtechszocs, line 46:
```suggestion
  const pluginStore = usePluginStore();
```

I'd generally prefer `pluginStore` since `store` might be too ambiguous.

packages/lib-core/src/store/PluginStoreContext.tsx
```diff
@@ -31,3 +31,19 @@ export const usePluginStore = (): PluginConsumer & PluginManager => {
...
+
+  return [store.getFeatureFlag(flagName), store.setFeatureFlag(flagName)];
```
vojtechszocs, line 48:
```ts
const currentValue = pluginStore.getFeatureFlag(name);

const setValue = React.useCallback((value: boolean) => {
  pluginStore.setFeatureFlag(name, value);
}, [pluginStore]);

return [currentValue, setValue];
```

Also, should the hook react to changes in feature flags? If so, we'll need some way to subscribe to `PluginStore` for changes in flags so we can re-render the component with an up-to-date `currentValue`. cc @christianvogt 

In other words, assuming `FOO` flag is currently not `true` (or undefined) and you do this in your component:
```ts
const [fooFlag, setFooFlag] = useFeatureFlag('FOO');
setFooFlag(true);
```
then your component should re-render, with `fooFlag` having the value `true`.

packages/lib-core/src/store/PluginStore.ts
```diff
@@ -227,11 +228,28 @@ export class PluginStore implements PluginConsumer, PluginManager {
...
+    const prevFeatureFlags = this.featureFlags;
+
+    this.featureFlags = {
+      ...this.featureFlags,
+      [flagName]: newFlagValue,
+    };
+
+    if (!_.isEqual(prevFeatureFlags, this.featureFlags)) {
```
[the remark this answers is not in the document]
vidyanambiar: @vojtechszocs Since we had discussed supporting 2 states (true/false), shall we check for one of those two states before we update the flag in the store? cc @christianvogt

PR #108, opened 2022-06-01
Sync search params with URL

fhlavac:
https://issues.redhat.com/browse/HAC-1433

Added search params sync with the URL for the table component so that search params change in the URL is reflected in search bar and vice versa.

syncURL

### DIFF
--- a/packages/lib-utils/src/utils/url-sync.ts
+++ b/packages/lib-utils/src/utils/url-sync.ts
@@ -1,15 +1,16 @@
 import type { History } from 'history';
+import * as _ from 'lodash';
 
 export const applyFiltersToURL = (
   history: History,
   keys: string[],
-  filterValues: Record<string, string>,
+  filterValues: Record<string, string[]>,
 ) => {
   const searchParams = new URLSearchParams(history.location.search);
   keys.forEach((key: string) => {
     searchParams.delete(key);
     if (filterValues[key]) {
-      searchParams.set(key, filterValues[key]);
+      filterValues[key].map((value) => searchParams.append(key, value));
     }
   });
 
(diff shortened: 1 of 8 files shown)


fhlavac (commented):
@vojtechszocs just not sure about this - it's probably eslint false positive
https://github.com/openshift/dynamic-plugin-sdk/pull/108/files#diff-d7b35d11b677828ab3332b86d539dfcc0c092a5afd51e65778371f871676929aR64
https://github.com/openshift/dynamic-plugin-sdk/pull/108/files#diff-d7b35d11b677828ab3332b86d539dfcc0c092a5afd51e65778371f871676929aR84

packages/lib-utils/src/utils/url-sync.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+  });
+
+  history.replace({
+    pathname: history.location.pathname,
+    search: searchParams.toString(),
+  });
+};
+
+export const syncFiltersWithUrl = (
```
vojtechszocs, line 22:
```suggestion
export const syncFiltersWithURL = (
```

packages/lib-utils/src/utils/url-sync.ts
```diff
@@ -0,0 +1,59 @@
...
+  keys: string[],
+  defaults: Record<string, string> = {},
+) => {
+  const searchParams = new URLSearchParams(history.location.search);
+
+  let filters: Record<string, string> = keys.reduce((acc, key) => {
+    const values = searchParams.getAll(key);
+    return {
+      ...acc,
+      [key]: values.length > 1 ? values : values[0],
```
florkbr, line 33: @fhlavac I wonder if we should update the logic below to handle the array in either scenario you're covering here - that way we are consistent and don't need as may `.length` checks?

packages/lib-utils/src/utils/url-sync.ts
```diff
@@ -0,0 +1,59 @@
...
+    };
+  }, {});
+
+  Object.keys(defaults).forEach((key: string) => {
+    const value = defaults[key];
+    filters = {
```
florkbr, line 39: @fhlavac is the only thing we're doing here is reassigning the value of `filters[key]` here to a default value if it's not set? We might want to favor chaining for readability - and we could also merge the objects together (defaults and parsed filters) using [lodash merge](https://lodash.com/docs/#merge). Something like: `_.merge(defaults, filters)`

fhlavac:
@vojtechszocs updated the PR according to your suggestions, works much better. Thank you!

PR #109, opened 2022-06-09
[HAC-1465] Fix query params and encode label selector correctly

vidyanambiar:
Issue: [HAC-1465](https://issues.redhat.com/browse/HAC-1465)

I tested this against hac-core [https://prod.foo.redhat.com:1337/beta/hac/testK8s](https://prod.foo.redhat.com:1337/beta/hac/testK8s) by first creating a few Application resources, adding a label `test-label=test-value` to one of them and then changing the `watchedResource` as follows on https://github.com/openshift/hac-core/blob/main/frontend/src/Routes/testK8s/HookTest.tsx#L16:
```
  const watchedResource: WatchK8sResource = {
    isList: true,
    groupVersionKind: {
      group: 'appstudio.redhat.com',
      version: 'v1alpha1',
      kind: 'Application',
    },
    namespace,
    selector: {
      matchLabels: {
        'test-label': 'test-value',
      },
    }
  };
```
 also tested with `fieldSelector`:
```
  const watchedResource: WatchK8sResource = {
    isList: true,
    groupVersionKind: {
      group: 'appstudio.redhat.com',
      version: 'v1alpha1',
      kind: 'Application',
    },
    namespace,
    fieldSelector: 'metadata.name=test'
  };
```

### DIFF
--- a/packages/lib-utils/src/app/redux/actions/k8s.ts
+++ b/packages/lib-utils/src/app/redux/actions/k8s.ts
@@ -5,9 +5,9 @@ import type { ActionType as Action } from 'typesafe-actions';
 import { action } from 'typesafe-actions';
 import type { Query } from '../../../k8s/hooks/k8s-watch-types';
 import { k8sListResource, k8sGetResource } from '../../../k8s/k8s-resource';
-import { k8sWatch } from '../../../k8s/k8s-utils';
+import { k8sWatch, selectorToString } from '../../../k8s/k8s-utils';
 import type { DiscoveryResources } from '../../../types/api-discovery';
-import type { K8sModelCommon, K8sResourceCommon, FilterValue } from '../../../types/k8s';
+import type { K8sModelCommon, K8sResourceCommon, FilterValue, Selector } from '../../../types/k8s';
 import type { ThunkDispatchFunction } from '../../../types/redux';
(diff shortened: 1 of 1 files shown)


packages/lib-utils/src/app/redux/actions/k8s.ts
```diff
@@ -120,11 +120,23 @@ export const watchK8sList =
           }
         : {};
 
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { ns, ...queryParameters } = queryWithCluster; // Omit ns from query parameters
```
vojtechszocs, line 124:
```suggestion
      const queryParameters = _.omit(queryWithCluster, 'ns');
```

packages/lib-utils/src/app/redux/actions/k8s.ts
```diff
@@ -120,11 +120,23 @@ export const watchK8sList =
...
+
+      const { labelSelector } = queryParameters;
+      if (labelSelector) {
+        const encodedSelector = selectorToString(<Selector>labelSelector);
```
vojtechszocs, line 128:
```suggestion
        const encodedSelector = selectorToString(labelSelector as Selector);
```

PR #116, opened 2022-06-27
Add tests for Table component

fhlavac:
https://issues.redhat.com/browse/HAC-1551

Added test coverage for
- `VirtualizedTableBody` - non virtualized
- `VirtualizedTable` - non virtualized
- `ListView`
- `StatusBox`
- `FilterChips`

@florkbr

### DIFF
--- a/packages/lib-utils/src/components/list-view/ListView.test.tsx
+++ b/packages/lib-utils/src/components/list-view/ListView.test.tsx
@@ -1,0 +1,208 @@
+import { screen, render, fireEvent } from '@testing-library/react';
+import * as React from 'react';
+import { act } from 'react-dom/test-utils';
+import { MemoryRouter } from 'react-router-dom';
+import type { TableTestItem } from '../table/test-data';
+import {
+  testData as data,
+  TestRow as Row,
+  testColumns as columns,
+  testFilters,
+} from '../table/test-data';
+import ListView, { filterDefault } from './ListView';
+
+describe('TableView - non-virtualized', () => {
+  it('should render with data', () => {
+    render(
+      <MemoryRouter initialEntries={[{ pathname: '/hac' }]}>
+        <ListView
+          data={data}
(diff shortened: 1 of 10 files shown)


packages/lib-utils/src/components/list-view/ListView.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+    await act(async () => {
+      render(
+        <MemoryRouter initialEntries={[{ pathname: '/hac' }]}>
+          <ListView
+            data={data}
+            columns={columns}
+            Row={Row}
+            filters={testFilters}
+            loaded
+            isRowSelected={undefined}
```
florkbr, line 56: Not blocking - I'd like for us to investigate de-duplicating these props across our tests (at least the ones that remain consistent between tests). Otherwise it's very hard to identify which props are changing between tests (if any at all)

packages/lib-utils/src/components/table/VirtualizedTable.tsx
```diff
@@ -23,21 +23,19 @@ export type VirtualizedTableProps<D> = {
   /** Optional load error object. */
   loadError?: LoadError;
   /** Table columns array. */
-  columns: TableColumn<D>[];
+  columns: TableColumn[];
```
florkbr, line 26: @fhlavac we will need to confirm if any of these prop updates will impact `hac-infra`.

florkbr:
Overall looks good to me and the tests _finally_ are passing!! 🎉

packages/lib-utils/src/components/list-view/ListView.test.tsx
vojtechszocs: Yes, and I'm not sure if we really need explicit `someProp={undefined}` ones as well, can be looked into later on.

vojtechszocs (approved):

PR #117, opened 2022-06-29
Rename TableView to ListView

fhlavac:
https://issues.redhat.com/browse/HAC-1686

Renamed `TableView` to `ListView`.


packages/lib-utils/src/components/table-view/TableView.tsx
```diff
@@ -25,14 +25,14 @@ export type FilterItem = {
   id: string;
 };
 
-export type TableViewProps<D> = VirtualizedTableProps<D> & {
+export type ListViewProps<D> = VirtualizedTableProps<D> & {
```
florkbr, line 28: @fhlavac I'm wondering if we should update the name of the files/directories as well to avoid confusion?

fhlavac: Sure, it's updated. Thank you :slightly_smiling_face:

florkbr:
@fhlavac needs rebase 😢

florkbr:
@fhlavac FYI the change to export the table merged before this one - so we will need to update the names in the index/exports: https://github.com/openshift/dynamic-plugin-sdk/pull/118

PR #121, opened 2022-07-07
Add support for only one filter

fhlavac:
https://issues.redhat.com/browse/HAC-1721

When there is only one filter defined, filter selection disappears and only text input is visible.
onlysearch

@florkbr not sure about the search icon - this is what PF search input looks like by default. Do we want to have the search icon on the right as in the mocks? Should it work as a submit button?

### DIFF
--- a/packages/eslint-plugin-internal/src/rules/react-typescript-prettier.js
+++ b/packages/eslint-plugin-internal/src/rules/react-typescript-prettier.js
@@ -13,4 +13,7 @@ module.exports = {
 
   // Enforce a defaultProps definition for every prop that is not a required prop
   'react/require-default-props': 'off',
+
+  // Enforce that block statements are wrapped in curly braces
+  curly: ['error', 'all'],
 };
--- a/packages/lib-utils/src/components/list-view/ListView.tsx
+++ b/packages/lib-utils/src/components/list-view/ListView.tsx
@@ -87,25 +87,27 @@ const ListView: React.FC<ListViewProps<Record<string, unknown>>> = ({
         <ToolbarContent>
           {filters ? (
             <>
-              <ToolbarItem key="filter-select">
-                <Select
-                  toggleIcon={<FilterIcon />}
-                  variant={SelectVariant.single}
-                  onToggle={(value) => setFilterSelectExpanded(value)}
-                  onSelect={(e, selection) => {
-                    setActiveFilter(filters.find((item) => item.id === selection));
-                    setFilterSelectExpanded(false);
-                  }}
-                  placeholderText={activeFilter?.label}
-                  isOpen={isFilterSelectExpanded}
-                >
-                  {filters.map((option) => (
-                    <SelectOption key={option.id} value={option.id}>
-                      {option.label}
-                    </SelectOption>
-                  ))}
-                </Select>
-              </ToolbarItem>
+              {filters.length > 1 && (
+                <ToolbarItem key="filter-select">
+                  <Select
+                    toggleIcon={<FilterIcon />}
+                    variant={SelectVariant.single}
+                    onToggle={(value) => setFilterSelectExpanded(value)}
+                    onSelect={(e, selection) => {
+                      setActiveFilter(filters.find((item) => item.id === selection));
+                      setFilterSelectExpanded(false);
+                    }}
+                    placeholderText={activeFilter?.label}
+                    isOpen={isFilterSelectExpanded}
+                  >
+                    {filters.map((option) => (
+                      <SelectOption key={option.id} value={option.id}>
+                        {option.label}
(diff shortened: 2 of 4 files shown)


karelhala (commented):
Looking good! WRT search icon, I'd keep it on the left to be consistent with PF and would fire a filter after 2 seconds of inactive typing or after pressing enter. No need to have a dedicated button.

vidyanambiar (commented):
Works well!

In the list's empty state, the HAC Infra Marvel design has the placeholder as simply "Search". Can we have the placeholder fallback to just `Search` if the `activeFilter.label` is provided as an empty string? 


image

image

vidyanambiar:
@fhlavac Just to be clear I'm suggesting continuing with the same implementation that you have where the placeholder would say `Search by ${activeFilter.label}` but if `activeFilter.label` is an empty string then the placeholder would fall back to simply `Search`. As of now it fall back to `Search by`.

fhlavac:
@vidyanambiar ah, ok, that would make sense. I'll add it.

vidyanambiar:
Works great! Thanks @fhlavac

fhlavac:
@karelhala added the 2s debounce

packages/lib-utils/src/components/list-view/ListView.tsx
```diff
@@ -32,6 +32,25 @@ export type ListViewProps<D> = VirtualizedTableProps<D> & {
   filters?: FilterItem[];
 };
 
+export function filterDefault<D>(data: D[], filterValues: Record<string, string[]>): D[] {
+  return data.filter((item) => {
+    let isRelevant = true;
+    Object.keys(filterValues).forEach((key) => {
```
vojtechszocs, line 38: This seems like a use case for [Array.every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) method to simplify the logic (I hope I understood it correctly).

```ts
return data.filter((item) => {
  return Object.entries(filterValues).every(([key, values]) => {
    return typeof item[key] === 'string' &&
      values.every((v) => item[key].toLowerCase().includes(v));
  });
});
```

Also `D` type parameter should be narrowed like `filterDefault<D extends Record<string, unknown>>` otherwise the function will make implicit assumptions about its structure.

packages/lib-utils/src/components/list-view/ListView.tsx
```diff
@@ -53,78 +72,83 @@ const ListView: React.FC<ListViewProps<Record<string, unknown>>> = ({
   const [filteredData, setFilteredData] = React.useState(data);
   const [isFilterSelectExpanded, setFilterSelectExpanded] = React.useState(false);
   const [searchParams, setSearchParams] = useSearchParams();
+  const [useUrl, setUseUrl] = React.useState(true);
```
vojtechszocs, line 75:
```suggestion
  const [useURL, setUseURL] = React.useState(true);
```

packages/lib-utils/src/components/list-view/ListView.tsx
```diff
@@ -53,78 +72,83 @@ const ListView: React.FC<ListViewProps<Record<string, unknown>>> = ({
...
-                          ? { ...filterValues.current, [activeFilter.id]: [value] }
-                          : omit(filterValues.current, activeFilter.id);
-                      setFiltersToURL(
-                        searchParams,
-                        setSearchParams,
-                        filters.map((filter) => filter.id),
-                        newValues,
-                      );
-                    }
+                    if (useUrl) setUseUrl(false);
```
vojtechszocs, line 146: Normally this would trigger an ESLint warning saying you should encapsulate `if` body inside `{` `}` braces to make it more readable - if it doesn't, maybe we should enforce this rule?

vojtechszocs (changes requested):

packages/lib-utils/src/components/list-view/ListView.tsx
fhlavac: @vojtechszocs added the curly braces, just not sure the eslint rule. Added it to the config but it seems not to work. Could you please take a look? Thank you!

packages/eslint-plugin-internal/src/rules/react-typescript-prettier.js
```diff
@@ -13,4 +13,7 @@ module.exports = {
 
   // Enforce a defaultProps definition for every prop that is not a required prop
   'react/require-default-props': 'off',
+
+  // Enforce that block statements are wrapped in curly braces
+  curly: [2, 'all'],
```
vojtechszocs, line 18: I'd suggest using string-based rule severity values instead of their numeric counterparts for better readability.

```suggestion
  curly: ['error', 'all'],
```

PR #122, opened 2022-07-07
Override react dependency in react-virtualized

fhlavac:


package.json
```diff
@@ -97,5 +97,9 @@
     "typescript": "~4.4.4",
     "uuid": "^8.3.2",
     "webpack": "^5.69.1"
+  },
+  "resolutions": {
```
vojtechszocs, line 101: npm v8 https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides
- nested dependencies represented as nested objects
- values can be semver ranges

Yarn (next) https://yarnpkg.com/configuration/manifest/#resolutions
- nested dependencies represented as strings, using `/` separator
- can only be used at the (monorepo) root of the project
- docs indicate the values should be specific versions

PR #124, opened 2022-07-11
Export useK8sModel hook and its type

karelhala:
### Description

We are missing `useK8sModel` and `useK8sModels` hooks to tap in redux and pull either single or all models. This PR will export these functions on top level of sdk utils package so we can consume it later down the line.


christianvogt:
I think it's good to expose these hooks so that plugins don't depend directly on redux. Furthermore if we change the context in which the model data is stored, plugins won't need to make any changes as the hook itself can be updated to use the new context under the covers.

PR #126, opened 2022-07-12
Initial storybook for list view

florkbr:
This PR will allow us to use storybook while developing components in the SDK. This allows for rapid isolated development, hot reloading, and eventually additional plugins around accessibility, etc.

1. Run `yarn install` after cloning
2. Run `yarn storybook`
3. Browser should auto open (if not open http://localhost:6006/)

Screen Shot 2022-07-12 at 7 26 26 PM


.storybook/preview.js
```diff
@@ -0,0 +1,9 @@
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
```
vojtechszocs, line 2: https://storybook.js.org/docs/react/writing-stories/parameters#global-parameters

> Setting a global parameter is a common way to configure addons.

So this config is consumed by the Storybook addons, is that right?

packages/lib-utils/src/components/list-view/ListView.stories.tsx
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable react/jsx-props-no-spreading */
+import '@patternfly/react-core/dist/styles/base.css';
```
vojtechszocs, line 2: > If you have global imported styles, create a file called [.storybook/preview.js](https://storybook.js.org/docs/react/configure/overview#configure-story-rendering) and import the styles there. They will be added by Storybook automatically for all stories.

I think we can import ^^ from the preview config file, so we don't have to repeat it in every stories file.

packages/lib-utils/src/components/list-view/ListView.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import ListView from './ListView';
+
+export default {
+  title: 'ListView',
+  component: ListView,
+  argTypes: {
+    backgroundColor: { control: 'color' },
+  },
+} as ComponentMeta<typeof ListView>;
```
vojtechszocs, line 16: Why do we need a force cast on a [default stories metadata export](https://storybook.js.org/docs/react/writing-stories/introduction#default-export)?

This does not ensure that given object conforms to `ComponentMeta` interface, it just force casts it to it.

```ts
type A = { a: string };
export default {} as A; // passes

// vs.

const x: A = {}; // fails
export default x;
```

Update: now I see that official Storybook docs actually recommend this approach (which is weird).

I'd suggest to make it type-safe, otherwise the `as` force cast is basically redundant.

```ts
const meta: ComponentMeta<typeof ListView> = {
title: 'ListView',
  component: ListView,
  argTypes: {
    backgroundColor: { control: 'color' },
  },
};

export default meta;
```

vojtechszocs (changes requested):

.storybook/main.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  "stories": [
+    "../packages/lib-utils/**/*.stories.mdx",
+    "../packages/lib-utils/**/*.stories.@(js|jsx|ts|tsx)"
+  ],
+  "addons": [
+    "@storybook/addon-links",
+    "@storybook/addon-essentials",
```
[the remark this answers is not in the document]
florkbr: I do not think the order matters. The doc shows the same order: https://storybook.js.org/docs/react/configure/overview#using-storybook-api

.storybook/preview.js
florkbr: Yes; specifically the [actions addon](https://storybook.js.org/docs/react/essentials/actions)

PR #127, opened 2022-07-20
Enable info tooltip/popover in ListView

fhlavac:
https://issues.redhat.com/browse/HAC-1855

Enabled info option for `ListView` column

Screenshot from 2022-07-20 10-08-29

@florkbr 
@vidyanambiar


vidyanambiar (commented):
Works great, thanks @fhlavac!
/lgtm

PR #128, opened 2022-07-21
Fix incorrect storing key for core resources

karelhala:
### Description

There's a bug where apps can't watch for core resources, because these aren't stored under proper key `$apiVersion~~$kind`, whereas it should be stored as `core~$apiVersion~$kind`. This is caused by splitting the `groupVersion` by `/`, but if no such string is found within the groupVersion it means it's just a version. So let's prepend this string with `/` if not found to properly split the string.

Since we are using batched requests let's pull for core ones first as they can't be properly sorted and are somewhat crutial to apps.

### DIFF
--- a/packages/lib-utils/src/app/api-discovery/index.ts
+++ b/packages/lib-utils/src/app/api-discovery/index.ts
@@ -30,7 +30,9 @@ const pluralizeKind = (kind: string): string => {
 
 const defineModels = (list: APIResourceList): K8sModelCommon[] => {
   const { groupVersion } = list;
-  const [apiGroup, apiVersion] = groupVersion.split('/');
+  const [apiGroup, apiVersion] = `${groupVersion.includes('/') ? '' : '/'}${groupVersion}`.split(
+    '/',
+  );
   if (!list.resources || list.resources.length < 1) {
     return [];
   }
@@ -134,13 +136,13 @@ const getResources = async (
     },
     {},
   );
-  const all = _.flatten(
-    apiResourceData.groups.map<string[]>((group) =>
-      group.versions.map<string>((version) => `/apis/${version.groupVersion}`),
-    ),
-  )
-    .concat(['/api/v1'])
-    .sort((api) => (preferenceList.find((item) => api.includes(`/apis/${item}`)) ? -1 : 0));
+  const all = ['/api/v1'].concat(
+    _.flatten(
+      apiResourceData.groups.map<string[]>((group) =>
+        group.versions.map<string>((version) => `/apis/${version.groupVersion}`),
+      ),
+    ).sort((api) => (preferenceList.find((item) => api.includes(`/apis/${item}`)) ? -1 : 0)),
+  );
 
   // let batchedData: APIResourceList[] = [];
   const batches = _.chunk(all, API_DISCOVERY_REQUEST_BATCH_SIZE);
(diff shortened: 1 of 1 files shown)


fhlavac (approved):

packages/lib-utils/src/app/api-discovery/index.ts
```diff
@@ -30,7 +30,9 @@ const pluralizeKind = (kind: string): string => {
 
 const defineModels = (list: APIResourceList): K8sModelCommon[] => {
   const { groupVersion } = list;
-  const [apiGroup, apiVersion] = groupVersion.split('/');
+  const [apiGroup, apiVersion] = `${groupVersion.includes('/') ? '' : '/'}${groupVersion}`.split(
+    '/',
+  );
```
christianvogt, line 35: Alternatively:
```suggestion
  const [apiVersion, apiGroup] = groupVersion.split('/').reverse();
```

karelhala: Is it better? You gotta think about the reverse in a different way, whereas the current specifally says if no slash is found prepend it so we get an empty `apiGroup`.

christianvogt: An `apiGroup` whose value is `''` or `undefined` is treated the same below where the value of `apiGroup` is read. Therefore the net result should be the same. Note that in my suggestion, the deconstruction assignment is reversed compared to the original line of code and `apiGroup` will be `undefined` if the split results in an array of length 1.

florkbr: @karelhala I tend to favor @christianvogt's suggestion for readability and simplicity. We could also set a default value for `apiGroup` if it turns out to be `undefined` if necessary.

packages/lib-utils/src/app/api-discovery/index.ts
```diff
@@ -134,13 +136,13 @@ const getResources = async (
...
     {},
   );
-  const all = _.flatten(
-    apiResourceData.groups.map<string[]>((group) =>
-      group.versions.map<string>((version) => `/apis/${version.groupVersion}`),
-    ),
-  )
-    .concat(['/api/v1'])
-    .sort((api) => (preferenceList.find((item) => api.includes(`/apis/${item}`)) ? -1 : 0));
+  const all = ['/api/v1'].concat(
```
florkbr, line 139: @karelhala I know this was an issue before your change, but we should consider removing the hard-coded version here (IE if we switch to version `V2` in the distant future). Not blocking.

packages/lib-utils/src/app/api-discovery/index.ts
karelhala: @florkbr `apiGroup` already has a default value. 

@christianvogt @florkbr if you prefer to read the `reverse` I am fine with it. For me it makes a bit more sense to "prepend" the string with `/` if not found. Both things are doing the same, it's just the first one is more straight forward, whereas the `reverse` with flipped destruct requires a bit more thinking of what it actually does.

packages/lib-utils/src/app/api-discovery/index.ts
karelhala: Which api we'd call for fetching the core models? We can add a constant by all means, just curious how'd you like to remove the version.

florkbr: I'm just suggesting we should have a singular location for the version of the core APIs - not that the API is changing or will change. For now this LGTM.

PR #129, opened 2022-07-22
Improve api discovery by properly storing data in localStorage

karelhala:
### Description

There are three main changes to this PR

- Load data from localStorage before waiting for API - this ensures the models are loaded in redux before application even starts
- Store resources for each batch - whenever we request models from API we can emediately store them in localStorage, that way when user refreshes a page mid API discovery they'll get at least some resources cached
- Make a proper merge of resources - for some reason lodash's merge was messing with arrays so we stored just the most recent resources

### DIFF
--- a/packages/lib-utils/src/app/api-discovery/discovery-cache.ts
+++ b/packages/lib-utils/src/app/api-discovery/discovery-cache.ts
@@ -1,23 +1,51 @@
 import { consoleLogger } from '@monorepo/common';
+import * as _ from 'lodash-es';
+import { getReferenceForModel } from '../../k8s/k8s-utils';
 import type { DiscoveryResources } from '../../types/api-discovery';
+import type { K8sModelCommon } from '../../types/k8s';
 
 const SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY = 'sdk/api-discovery-resources';
 
-export const cacheResources = (resources: DiscoveryResources) => {
+const mergeByKey = (prev: K8sModelCommon[], next: K8sModelCommon[]) =>
+  Object.values(_.merge(_.keyBy(prev, getReferenceForModel), _.keyBy(next, getReferenceForModel)));
+
+export const cacheResources = (resources: DiscoveryResources[]) => {
+  let allResources;
   try {
-    localStorage.setItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY, JSON.stringify(resources));
+    allResources = [
+      ...[JSON.parse(localStorage.getItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY) || '{}')],
+      ...resources,
+    ].reduce((acc, curr) => {
+      const { models, configResources, clusterOperatorConfigResources, ...rest } = curr || {};
+      return {
+        ..._.merge(acc, rest),
+        configResources: mergeByKey(acc?.configResources, configResources || []),
+        models: mergeByKey(acc?.models, models || []),
+        clusterOperatorConfigResources: mergeByKey(
+          acc?.clusterOperatorConfigResources,
+          clusterOperatorConfigResources || [],
+        ),
+      };
+    }, {});
+  } catch (e) {
+    consoleLogger.error('Error caching API resources in localStorage', e);
+    throw e;
+  }
+  try {
+    localStorage.setItem(
+      SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY,
+      JSON.stringify(allResources),
+    );
   } catch (e) {
     consoleLogger.error('Error caching API resources in localStorage', e);
     throw e;
   }
 };
 
-export const getCachedResources = async () => {
+export const getCachedResources = () => {
   const resourcesJSON = localStorage.getItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY);
   if (!resourcesJSON) {
-    throw new Error(
-      `No API resources found in localStorage for key ${SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY}`,
-    );
+    return null;
   }
 
   // Clear cached resources after load as a safeguard. If there's any errors
--- a/packages/lib-utils/src/app/api-discovery/discovery-cache.ts
+++ b/packages/lib-utils/src/app/api-discovery/discovery-cache.ts
@@ -9,42 +9,44 @@ const SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY = 'sdk/api-discovery-resourc
 const mergeByKey = (prev: K8sModelCommon[], next: K8sModelCommon[]) =>
   Object.values(_.merge(_.keyBy(prev, getReferenceForModel), _.keyBy(next, getReferenceForModel)));
 
-export const cacheResources = (resources: DiscoveryResources[]) => {
-  let allResources;
+const getLocalResources = () => {
   try {
-    allResources = [
-      ...[JSON.parse(localStorage.getItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY) || '{}')],
-      ...resources,
-    ].reduce((acc, curr) => {
-      const { models, configResources, clusterOperatorConfigResources, ...rest } = curr || {};
-      return {
-        ..._.merge(acc, rest),
-        configResources: mergeByKey(acc?.configResources, configResources || []),
-        models: mergeByKey(acc?.models, models || []),
-        clusterOperatorConfigResources: mergeByKey(
-          acc?.clusterOperatorConfigResources,
-          clusterOperatorConfigResources || [],
-        ),
-      };
-    }, {});
+    return JSON.parse(localStorage.getItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY) || '{}');
   } catch (e) {
-    consoleLogger.error('Error caching API resources in localStorage', e);
+    consoleLogger.error('Cannot load cached API resources', e);
     throw e;
   }
+};
+
+const setLocalResources = (resources: DiscoveryResources[]) => {
   try {
-    localStorage.setItem(
-      SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY,
-      JSON.stringify(allResources),
-    );
+    localStorage.setItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY, JSON.stringify(resources));
   } catch (e) {
     consoleLogger.error('Error caching API resources in localStorage', e);
     throw e;
   }
 };
+export const cacheResources = (resources: DiscoveryResources[]) => {
+  const allResources = [...[getLocalResources()], ...resources].reduce(
+    (acc, curr) =>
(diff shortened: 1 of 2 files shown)


packages/lib-utils/src/app/api-discovery/discovery-cache.ts
```diff
@@ -1,23 +1,51 @@
...
+  Object.values(_.merge(_.keyBy(prev, getReferenceForModel), _.keyBy(next, getReferenceForModel)));
+
+export const cacheResources = (resources: DiscoveryResources[]) => {
+  let allResources;
+  try {
+    allResources = [
+      ...[JSON.parse(localStorage.getItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY) || '{}')],
+      ...resources,
+    ].reduce((acc, curr) => {
+      const { models, configResources, clusterOperatorConfigResources, ...rest } = curr || {};
```
karelhala, line 19: For some reason lodash messes merge and if `models`, `configResources` or `clusterOperatorConfigResources` were in merged object they were rewritten and only last ones were used.

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
```diff
@@ -1,23 +1,51 @@
...
+        configResources: mergeByKey(acc?.configResources, configResources || []),
+        models: mergeByKey(acc?.models, models || []),
+        clusterOperatorConfigResources: mergeByKey(
+          acc?.clusterOperatorConfigResources,
+          clusterOperatorConfigResources || [],
+        ),
+      };
+    }, {});
+  } catch (e) {
+    consoleLogger.error('Error caching API resources in localStorage', e);
```
vojtechszocs, line 31: I assume that only JSON parse & local storage access may result in an error, otherwise it's just data processing.

I'd suggest adding functions to handle the locally cached API resources, for example:

```ts
const getLocalResources = () => {
  try {
    return JSON.parse(localStorage.getItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY) || '{}');
  } catch (e) {
    throw new Error('Cannot load cached API resources', e);
  }
};

const setLocalResources = (resources) => {
  try {
    localStorage.setItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY, JSON.stringify(resources));
  } catch (e) {
    throw new Error('Cannot write cached API resources', e);
  }
};
```

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
```diff
@@ -1,23 +1,51 @@
...
 };
 
-export const getCachedResources = async () => {
+export const getCachedResources = () => {
   const resourcesJSON = localStorage.getItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY);
   if (!resourcesJSON) {
-    throw new Error(
-      `No API resources found in localStorage for key ${SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY}`,
-    );
+    return null;
```
florkbr, line 48: @karelhala for debugging purposes we may want to consider keeping some indication in the console this condition is hit. Especially since we clear the cache just below it (if there is a caching issue that occurs every page refresh).

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
```diff
@@ -1,23 +1,51 @@
 import { consoleLogger } from '@monorepo/common';
+import * as _ from 'lodash-es';
+import { getReferenceForModel } from '../../k8s/k8s-utils';
```
florkbr, line 3: @karelhala this style of resource reference (tilda `~`) is deprecated. I'd like to avoid relying on it more if possible.

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
florkbr: @karelhala hm... I was hoping for a cleaner implementation of the cache merge... Perhaps this is due to the fact that lodash mutates the target object passed in. Have you tried cloning the objects before merging with spread?

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
florkbr: @vojtechszocs agreed

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
```diff
@@ -1,23 +1,51 @@
...
 
-export const cacheResources = (resources: DiscoveryResources) => {
+const mergeByKey = (prev: K8sModelCommon[], next: K8sModelCommon[]) =>
+  Object.values(_.merge(_.keyBy(prev, getReferenceForModel), _.keyBy(next, getReferenceForModel)));
+
+export const cacheResources = (resources: DiscoveryResources[]) => {
+  let allResources;
+  try {
+    allResources = [
+      ...[JSON.parse(localStorage.getItem(SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY) || '{}')],
```
florkbr, line 16: For readability we probably could move this to a separate line/variable.

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
karelhala: We can avoid it, but we need to construct a key by which to merge the arrays, the models are defined by group, version and kind so I found this convinient to use. We could construct a key like `${group}/${version}-${kind}` to avoid using the deprecated thing.

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
karelhala: Sure thing, I'll use the consoleLogger error.

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
karelhala: The problem is that `merge` from lodash is merging the data over keys and if we merge multiple arrays together they get overwritten by the second (because arrays have same keys). That's why I had to come up with the `mergeByKey` just for the arrays.

If I kept the `models`, `configResources` or `clusterOperatorConfigResources` in the current object the spread was ignored. That's because lodash is performing the merge in recursive manner, so when the reduce finishes it uses the later value (because these are nested and lodash is doing some perfomance goodies on top of this). We could change the `reduce` arguments to
```JS
.reduce(
  {
    models: accModels,
    configResources: accConfigResources,
    clusterOperatorConfigResources: accClusterOperatorConfigResources,
    ...acc
  } = {},
  { 
    models: currModels,
    configResources: currConfigResources,
    clusterOperatorConfigResources: currClusterOperatorConfigResources,
    ...curr
  } ={}
)
```
Not sure if this would make things easier to read. I can try to play around with `mergeWith` and write a bit more complex compare function.

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
karelhala: Good point!

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
vojtechszocs: I think it's OK to use a composite group/version/kind string as a key, since each element is a string on its own.

@florkbr AFAIK, Console doesn't want to promote `group~version~kind` key as a public facing API but it's OK to use it internally, within Console app or within our SDK utilities.

vojtechszocs: @karelhala @florkbr In future we can consider renaming `getReference` to `getInternalReference` and `getReferenceForModel` to `getInternalReferenceForModel` while undeprecating them and ensuring they are not exposed to SDK consumers.

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
```diff
@@ -1,23 +1,53 @@
...
   } catch (e) {
     consoleLogger.error('Error caching API resources in localStorage', e);
     throw e;
   }
 };
+export const cacheResources = (resources: DiscoveryResources[]) => {
+  const allResources = [...[getLocalResources()], ...resources].reduce(
+    (acc, curr) =>
+      _.mergeWith(acc, curr, (first, second) => {
+        if (Array.isArray(first) && first[0]?.constructor?.name === 'Object') {
```
vojtechszocs, line 33: Can we replace `first[0]?.constructor?.name === 'Object'` with `_.isPlainObject(first)` ?

packages/lib-utils/src/app/api-discovery/index.ts
```diff
@@ -153,6 +153,8 @@ const getResources = async (
       allResources.push(resource);
       dispatch(receivedResources(resource));
     });
+    // // Cache the resources whenever discovery completes to improve console load times.
```
vojtechszocs, line 156:
```suggestion
    // Cache the resources whenever discovery completes to improve console load times.
```

vojtechszocs (changes requested):

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
karelhala: Not really, the `isPlainObject` will return true for `new String('')`. This one directly returns true only for variables which are key value pair.

karelhala: Actually, testing it will return the same value. So the question is, do we want to use library's function or plain JS one?

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
florkbr: Agreed - since this is internal use we can keep as is

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
florkbr: Lodash implements this more or less the same way (with more checks to see if the constructor is in fact a function): https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L12036-L12047

We've been trying to use ES6/vanilla JS utilities when possible over Lodash in Openshift Console.

@karelhala one thing we may want to consider - if we do not use Lodash's implementation here, perhaps we should introduce a utility function that can be reused that does similar.

packages/lib-utils/src/app/api-discovery/discovery-cache.ts
```diff
@@ -1,23 +1,53 @@
 import { consoleLogger } from '@monorepo/common';
+import * as _ from 'lodash-es';
+import { getReferenceForModel } from '../../k8s/k8s-utils';
 import type { DiscoveryResources } from '../../types/api-discovery';
+import type { K8sModelCommon } from '../../types/k8s';
 
 const SDK_API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY = 'sdk/api-discovery-resources';
 
-export const cacheResources = (resources: DiscoveryResources) => {
+const mergeByKey = (prev: K8sModelCommon[], next: K8sModelCommon[]) =>
```
florkbr, line 9: @karelhala we will need to add unit tests for these new functions

florkbr:
For now this LGTM. @karelhala please let me know if you'd like me to create tickets for the unit tests and the lodash helper function @vojtechszocs pointed out.

PR #130, opened 2022-07-25
Fix VirtualizedTable columns widths

fhlavac:
Fixes: https://issues.redhat.com/browse/HAC-1818

- Fixed broken column widths across `VirtualizedTable`.
- Removed `info` option from table columns - it is broking this fix. We should rather pass the info button within the `title` element.

image

@vidyanambiar 
@florkbr


vidyanambiar (commented):
@fhlavac I'm still seeing this issue. In my case the row selection feature (checkboxes) is disabled. But I also tried with row selection enabled.
image
image

vidyanambiar:
@fhlavac Sure, you can refer to the hac-infra code in this branch - https://github.com/openshift/hac-infra/compare/main...vidyanambiar:hac-infra:ws-list-mockdata?expand=1
specifically `src/components/WorkspaceList/WorkspaceList.tsx` and `src/components/WorkspaceList/WorkspaceListConfig.tsx`
I applied the sdk utils package built from this PR on both HAC Core and HAC Infra's `@openshift/dynamic-plugin-sdk-utils` node_module. Then started HAC Core with `ENVIRONMENT=prod yarn dev` and hac-infra with `npm run start:federated` and opened https://prod.foo.redhat.com:1337/beta/hac/workspaces in the browser.

fhlavac:
@vidyanambiar ah I probably forgot to mention. For the fix to work, you need to use `Td` element exposed from the SDK instead of the PF one. We need to add some classes there to fix the styling. I know it's not the best solution but temporarily there is probably no better way. Please try it if you have time.

fhlavac:
@vidyanambiar it seems to be working with that change 
image

vidyanambiar:
Thanks @fhlavac. It looks much better with the `Td` from the sdk 👍 . 
However, `Td` is not exposed via `index.ts`. To test this I had to add an `index.ts` file under `packages/lib-utils/src/components/table` to export `Td` and also update `packages/lib-utils/src/index.ts` accordingly. Could you please include that change in the PR?
Also, I do still see a partially hidden row action selector at the end of the row though. Could you take a look at that as well please?
Screen Shot 2022-07-26 at 12 34 36 PM


I think it would help to include a Readme with guidance on special instructions such as this one (about using the Td from the SDK) as more folks are starting to use the ListView. In a future PR perhaps? WDYT? @fhlavac 
cc: @florkbr

fhlavac:
@vidyanambiar I added the exports after your first comment, it should be already exposed. Yeah, the row actions overflow is the bug I mentioned, preparing a fix. I will also update README. Thank you :slightly_smiling_face:

vidyanambiar:
oops, Sorry I missed it!

packages/lib-utils/src/components/table/VirtualizedTable.tsx
```diff
@@ -167,20 +168,20 @@ const VirtualizedTable: React.FC<VirtualizedTableProps<AnyObject>> = ({
       CustomNoDataEmptyState={CustomNoDataEmptyState}
     >
       <div role="grid" aria-label={ariaLabel} aria-rowcount={data?.length || 0}>
-        <TableComposable aria-label={ariaLabel} role="presentation">
+        <TableComposable aria-label={ariaLabel} role="presentation" gridBreakPoint="">
```
florkbr, line 171: Out of curiosity, what does setting `gridBreakPoint=""` fix? Will this change cause issues with the table wrapping in smaller browser viewports?

packages/lib-utils/src/components/table/VirtualizedTable.tsx
```diff
@@ -167,20 +168,20 @@ const VirtualizedTable: React.FC<VirtualizedTableProps<AnyObject>> = ({
...
-                select={{
-                  onSelect: (event, rowSelected) => onSelect?.(event, rowSelected, data),
-                  isSelected: data.every((item) => isRowSelected?.(item)),
-                }}
-              />
+              {onSelect && (
+                <Th
+                  className="pf-m-truncate dps-list-view__table-text"
+                  select={{
+                    onSelect: (event, rowSelected) => onSelect?.(event, rowSelected, data),
```
florkbr, line 178: Is the optional chaining necessary with the null check on 174?

packages/lib-utils/src/components/table/VirtualizedTableBody.tsx
```diff
@@ -38,6 +36,18 @@ export type TableRowProps = {
   className?: string;
 };
 
+export const Td = React.forwardRef<
```
florkbr, line 39: @fhlavac is there any added functionality to exposing this component or can apps/plugins just use the `PFTd` component exposed directly from patternfly?

fhlavac: @florkbr our Td component has additional classes which are trying to fix the styling issues. There is no other reason but also probably no better solution for now.

packages/lib-utils/src/components/table/VirtualizedTable.tsx
fhlavac: That's again an issue with compatibility of the virtualized extension with table composable since when it switches to mobile view, we would miss the table header:
image

So I disabled it:
image

florkbr:
@fhlavac there is a linter error :(
```
/go/src/github.com/openshift/dynamic-plugin-sdk/packages/lib-utils/src/components/table/VirtualizedTable.tsx
  3:10  warning  'Td' is defined but never used  @typescript-eslint/no-unused-vars

✖ 1 problem (0 errors, 1 warning)
```

fhlavac:
@florkbr fixed the error, thank you. I'll take another look to try fixing the action cell width

florkbr:
FWIW the table column header misalignment issue only occurs with the virtualized table (disconnect between the table body and the header). For now this LGTM

PR #131, opened 2022-07-25
Add sample plugin docker config

karelhala:
### Description

We need to have an option to spin the sample plugin with docker, this PR adds docker config with option to pass custom variables to it in order to serve the plugin on HCC.

### DIFF
--- a/packages/sample-plugin/Dockerfile
+++ b/packages/sample-plugin/Dockerfile
@@ -1,0 +1,9 @@
+FROM quay.io/cloudservices/caddy-ubi:357c825
+
+COPY ./Caddyfile /opt/app-root/src/Caddyfile
+COPY dist /opt/app-root/src/dist/
+COPY ./package.json /opt/app-root/src
+WORKDIR /opt/app-root/src
+ENV PLUGIN_URL=/
+ENV FALLBACK_URL=/
+CMD ["caddy", "run", "--config", "/opt/app-root/src/Caddyfile"]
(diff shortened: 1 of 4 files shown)


packages/sample-plugin/README.md
```diff
@@ -0,0 +1,16 @@
+# Sample plugin
+
+### Docker config
+
+There's docker config with caddy server for easier build and run of the application. The Caddy config utilizes two environment variables
```
florkbr, line 5: @karelhala should we refer to this as the `plugin` and not an `application`? @vojtechszocs any input on the readme? I know you were going to look at adding them for the sample app and plugin.

packages/sample-plugin/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM quay.io/cloudservices/caddy-ubi:357c825
+
+COPY ./Caddyfile /opt/app-root/src/Caddyfile
+COPY dist /opt/app-root/src/dist/
+COPY ./package.json /opt/app-root/src
+WORKDIR /opt/app-root/src
+ENV PLUGIN_URL=/
+ENV FALLBACK_URL=/
```
BlakeHolifield, line 8: I like the ENV idea here :+1:

BlakeHolifield (approved):
Looks good from my end. Nice Caddyfile upgrade!

packages/sample-plugin/README.md
karelhala: Good point! Will change it to `plugin`

PR #132, opened 2022-07-27
Hide "select all" checkbox when no onSelect fn passed

fhlavac:
Fixes https://coreos.slack.com/archives/D03BC1Y958R/p1657549005264639

Fixed a bug where  "Select all" checkbox was visible even if no onSelect function has been passed to ListView.

@vidyanambiar


vidyanambiar:
Thanks @fhlavac!
/lgtm

PR #133, opened 2022-07-27
Support both basic and virtualized table in ListView

fhlavac:
https://issues.redhat.com/browse/HAC-2006
Enabled both virtualized and non-virtualized table variants for `ListView`

Virtualization can be enabled by `virtualized` boolean property on both `ListView` and `VirtualizedTable`

Screenshot:
image

[Video non-virtualized](https://user-images.githubusercontent.com/50696716/181518972-b2045250-aaa9-4b26-9efd-e367e7e39111.webm)

[Video virtualized](https://user-images.githubusercontent.com/50696716/182166320-763f5286-d07a-46c3-b6bb-6481fcdeba17.webm)

@florkbr @vidyanambiar

### DIFF
--- a/packages/lib-utils/src/components/list-view/ListView.stories.tsx
+++ b/packages/lib-utils/src/components/list-view/ListView.stories.tsx
@@ -1,10 +1,18 @@
 /* eslint-disable react/jsx-props-no-spreading */
-import { Td } from '@patternfly/react-table';
+import type { AnyObject } from '@monorepo/common';
+import { Button, Tooltip } from '@patternfly/react-core';
+import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
+import { sortable, Td } from '@patternfly/react-table';
 import type { ComponentStory, ComponentMeta } from '@storybook/react';
 import React from 'react';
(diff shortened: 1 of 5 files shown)


packages/lib-utils/src/components/list-view/ListView.stories.tsx
```diff
@@ -15,10 +23,24 @@ const meta: ComponentMeta<typeof ListView> = {
...
   return (
     <BrowserRouter>
       <div style={{ overflowY: 'auto' }}>
-        <ListView {...args} />
+        <ListView
+          {...args}
+          onSelect={(e, isRowSelected, selectedData) =>
+            isRowSelected
+              ? setSelected([
+                  ...new Set([...selected, ...selectedData.map((i) => i.name as string)]),
```
vojtechszocs, line 36: The `i.name as string` force cast may indicate an issue with `selectedData` type, can we adjust it so that the force cast is not needed?

fhlavac: @vojtechszocs rReplaced it with `String(...)` or do you have a better idea? Also fixed an issue in default pagination logic causing sorting being applied only to page data - not all data

PR #134, opened 2022-07-28
Add plugin dependency resolution

vojtechszocs:
This PR adds support for resolving plugin dependencies in a way that is compatible with openshift/console#11626.

Fixes [HAC-1720](https://issues.redhat.com/browse/HAC-1720)

Added `PluginLoader.loadPlugin` method to encapsulate the logic behind loading a plugin from the specified URL.

1. Fetch and validate the manifest (`plugin-manifest.json`).
2. Resolve plugin dependencies, if any.
3. Load and process the entry script (`plugin-entry.js`).

A plugin can declare its dependencies via the optional `dependencies` metadata field, for example:
```json
"dependencies": {
  "foobar": "~1.2.3",
  "sample-app": "^1.0.0"
},
```

There are two kinds of dependencies:
- _Regular dependencies_ refer to actual plugins that must be loaded with the right version prior to loading the plugin that depends on them.
- _Fixed dependencies_ refer to any non-plugin dependencies like host application version, runtime platform version, etc. They are specified by the host application via the `PluginLoader` option `fixedPluginDependencyResolutions`.

```ts
const pluginLoader = new PluginLoader({
  sharedScope: getSharedScope(),
  fixedPluginDependencyResolutions: { 'sample-app': '1.0.0' },
});
```

Given the examples above, `foobar` is a regular dependency while `sample-app` is a fixed dependency that represents the host application version. If the `fixedPluginDependencyResolutions` option is missing, it falls back to an empty object meaning all dependencies refer to actual plugins.

_Regular dependencies are currently not evaluated eagerly._ Loading the above plugin will not automatically trigger loading of plugin `foobar` since we currently don't have a concept of a "plugin registry" that would map plugin names to corresponding version & URL resolutions. This current behavior is consistent with openshift/console#11626 which basically requires the host application to initiate loading of all the plugins, including any dependencies of those plugins.

Dependency resolution is designed to fail early if there are any unsuccessful or unmet resolutions. If the `sample-app` fixed dependency has `2.0.0` resolution and we try to load the above plugin, the `"sample-app": "^1.0.0"` dependency will be immediately rejected as non-matching, so we won't wait for plugin `foobar` to load but fail immediately:
```
Detected 1 resolution errors with 1 pending resolutions (foobar)

Dependency sample-app not met: required range ^1.0.0, resolved version 2.0.0
```

### Code improvements

- `PluginLoader` now invokes its listeners early upon an error. This in turn allows `PluginStore` to register failed plugins even before reaching the "Load and process the entry script" step.
- When `PluginLoader` fails to load a plugin, we pass `errorMessage` and `errorCause` properties as part of the load result. These properties are exposed to consumers via `PluginStore.getPluginInfo` method.
- `PluginConsumer` and `PluginManager` interfaces have been merged into a single `PluginStoreInterface` describing the public API of `PluginStore`. Consumer APIs like the `usePluginStore` hook now work with this interface.
- Pending info entries removed from `PluginStore.getPluginInfo` method result since the `PluginLoader` currently notifies its subscribers only when a plugin has finished loading successfully or has failed to load.
- `PluginStore.setPluginsEnabled` has been split into two methods, `enablePlugins` and `disablePlugins`. The latter allows consumers to provide an optional reason why the given plugin(s) got disabled. This is also used internally when disabling a plugin due to code reference resolution errors. The `disableReason` property is exposed to consumers via `PluginStore.getPluginInfo` method.
- `sample-app` component `PluginInfoTable` improved to render `disableReason` (if present) via info icon tooltip.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "packages/*"
   ],
   "scripts": {
-    "build": "yarn workspaces foreach run build",
+    "build": "yarn build-libs && yarn build-samples",
     "build-libs": "yarn workspaces foreach --from '@openshift/*' run build",
     "build-samples": "yarn workspaces foreach --from '@monorepo/sample-*' run build",
     "run-samples": "yarn workspaces foreach -pvi --from '@monorepo/sample-*' run http-server",
@@ -53,6 +53,7 @@
     "@types/react-router": "^5.1.17",
     "@types/react-router-dom": "^5.3.3",
     "@types/react-virtualized": "^9.21.21",
+    "@types/semver": "^7.3.10",
     "@types/uuid": "^8.3.4",
     "@typescript-eslint/eslint-plugin": "^5.3.1",
     "@typescript-eslint/parser": "^5.3.1",
--- a/packages/lib-core/src/types/store.ts
+++ b/packages/lib-core/src/types/store.ts
@@ -1,11 +1,11 @@
 import type { LoadedExtension } from './extension';
-import type { LoadedPlugin } from './plugin';
+import type { LoadedPlugin, FailedPlugin } from './plugin';
 
 export enum PluginEventType {
   /**
    * Triggers when the list of extensions, which are currently in use, changes.
    *
-   * Associated data getter: {@link PluginConsumer.getExtensions}
+   * Associated data getter: {@link PluginStoreInterface.getExtensions}
    */
   ExtensionsChanged = 'ExtensionsChanged',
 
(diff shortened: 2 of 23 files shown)


packages/lib-core/src/types/store.ts
```diff
@@ -15,39 +15,35 @@ export enum PluginEventType {
    * - plugin failed to load, or there was an error while processing its extensions
    * - plugin was enabled or disabled (*)
    *
-   * (*) this triggers event {@link PluginEventType.ExtensionsChanged}
+   * (*) this may trigger event {@link PluginEventType.ExtensionsChanged}
    *
-   * Associated data getter: {@link PluginConsumer.getPluginInfo}
+   * Associated data getter: {@link PluginStoreInterface.getPluginInfo}
```
florkbr, line 20: @vojtechszocs sorry for the delay in reviewing. There is a lot here, but overall looks good.

Since we are changing several public APIs I'd like to test these changes in hac-core and associated plugins. 

Do you have areas you suspect that may be breaking for existing consumers I can focus on?

florkbr:
@vojtechszocs some linter violations as well :(

```

/go/src/github.com/openshift/dynamic-plugin-sdk/packages/lib-core/src/store/PluginStore.ts
  233:7  error  Assignment to property of function parameter 'plugin'  no-param-reassign
  239:7  error  Assignment to property of function parameter 'plugin'  no-param-reassign
/go/src/github.com/openshift/dynamic-plugin-sdk/packages/lib-core/src/yup-schemas.ts
  85:3  error  Prop type "object" is forbidden  react/forbid-prop-types
/go/src/github.com/openshift/dynamic-plugin-sdk/packages/sample-app/src/components/common/PluginInfoTable.tsx
  16:10  error  Identifier 'global_info_color_100' is not in camel case  camelcase
  16:10  error  Identifier 'global_info_color_100' is not in camel case  camelcase
✖ 5 problems (5 errors, 0 warnings)
```

yaacov:
adding @sjd78 and @rszwajko for visibitlity

vojtechszocs:
Note to self, include the change from https://github.com/openshift/console/pull/12078 in this PR.

packages/lib-core/src/types/store.ts
vojtechszocs: > Since we are changing several public APIs I'd like to test these changes in hac-core and associated plugins.

:+1:

vojtechszocs: > Do you have areas you suspect that may be breaking for existing consumers I can focus on?

@florkbr From a host app perspective, this should only impact how plugin dependencies are handled at runtime.

If an existing plugin has no dependencies, there should be no impact. You can try to modify an existing plugin and add some dependencies, then put `debugger` statements in the resolution logic to see that everything works as expected. You can test both "regular" and "fixed" dependency types, as documented in PR description.

package.json
```diff
@@ -5,7 +5,7 @@
     "packages/*"
   ],
   "scripts": {
-    "build": "yarn workspaces foreach run build",
+    "build": "yarn build-libs && yarn build-samples",
```
vojtechszocs, line 8: Success of building sample packages depends on the success of building lib packages.

This change allows the top-level build script to fail early, instead of attempting to build all the packages.

vojtechszocs:
/hold 

openshift/console#12078 still pending discussions, better wait until we reach a decision.

packages/sample-app/src/components/common/ErrorBoundary.tsx
```diff
@@ -7,7 +7,7 @@ export type ErrorBoundaryFallbackProps = {
   errorInfo: React.ErrorInfo;
 };
 
-type ErrorBoundaryProps = AnyObject;
+type ErrorBoundaryProps = React.PropsWithChildren<AnyObject>;
```
florkbr, line 10: FWIW there is a ticket in OCP console to expose our common error boundary fallback: https://issues.redhat.com/browse/CONSOLE-3276. Maybe we could eventually use a common one here in lieu of a custom sample one.

packages/lib-core/src/yup-schemas.ts
```diff
@@ -81,6 +81,10 @@ export const extensionArraySchema = yup.array().of(extensionSchema).required();
 export const pluginRuntimeMetadataSchema = yup.object().required().shape({
   name: pluginNameSchema,
   version: semverStringSchema,
+  // TODO(vojtech): Yup lacks native support for map-like structures with arbitrary keys
+  // TODO(vojtech): suppress false positive https://github.com/jsx-eslint/eslint-plugin-react/pull/3326
```
florkbr, line 85: 👍

florkbr:
I tested these changes in HAC Core with the HAC Dev plugin and things continue to behave as normal.

The sample app/plugin dependency test also looks good to me locally.

:shipit:

PR #136, opened 2022-08-03
Bump lib versions after npm publish

vojtechszocs:


florkbr:
@vojtechszocs thanks! I was literally just about to hit the publish button and noticed the versions had changed from under me :)

PR #137, opened 2022-08-08
Improve the build of lib packages

vojtechszocs:
- create reusable Rollup plugins for common build tasks
- remove banner in the `.js` bundle in favor of `dist/build-meta.json`
- remove `dist/types` after generating the `.d.ts` bundle
- use [`downlevel-dts`](https://github.com/sandersn/downlevel-dts) tool to generate TS 3.8 compliant `.d.ts` bundle
- use TypeScript [`typesVersions`](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions) to consume the right type declarations (for `lib-core` only)

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@typescript-eslint/parser": "^5.3.1",
     "babel-loader": "^8.2.5",
     "cypress": "^9.7.0",
+    "downlevel-dts": "^0.10.0",
     "eslint": "^7.32.0",
     "eslint-config-airbnb": "^19.0.0",
     "eslint-config-airbnb-base": "^15.0.0",
@@ -70,7 +71,9 @@
     "eslint-plugin-promise": "^5.1.1",
     "eslint-plugin-react": "^7.27.0",
     "eslint-plugin-react-hooks": "^4.3.0",
+    "fs-extra": "^10.1.0",
     "jest": "^27.5.1",
+    "lodash": "^4.17.21",
     "lodash-es": "^4.17.21",
     "prettier": "^2.4.1",
     "react": "^17.0.2",
@@ -90,6 +93,7 @@
     "ts-jest": "^27.1.4",
     "ts-node": "^10.7.0",
     "tslib": "^2.3.1",
+    "type-fest": "^2.18.0",
     "typescript": "~4.4.4",
     "webpack": "^5.69.1"
   }
--- a/packages/common/rollup-configs.js
+++ b/packages/common/rollup-configs.js
@@ -1,55 +1,50 @@
-import { execSync } from 'child_process';
 import commonjs from '@rollup/plugin-commonjs';
 import nodeResolve from '@rollup/plugin-node-resolve';
 import typescript from '@rollup/plugin-typescript';
 import analyzer from 'rollup-plugin-analyzer';
 import dts from 'rollup-plugin-dts';
 import css from 'rollup-plugin-import-css';
+import { removeFiles } from './rollup-plugins/removeFiles';
+import { replaceCode } from './rollup-plugins/replaceCode';
+import { writeBuildMetadata } from './rollup-plugins/writeBuildMetadata';
+
+/** @typedef {import("type-fest").PackageJson} PackageJson */
 
 // https://yarnpkg.com/advanced/lifecycle-scripts#environment-variables
 const rootDir = process.env.PROJECT_CWD;
 
+/**
+ * @param {PackageJson} pkg
+ */
 const getExternalModules = (pkg) => [
   ...Object.keys(pkg.dependencies ?? {}),
   ...Object.keys(pkg.peerDependencies ?? {}),
   ...('lodash' in pkg.dependencies ? ['lodash-es'] : []),
 ];
 
+/**
+ * @param {PackageJson} pkg
+ */
 const getExternalModuleRegExps = (pkg) => {
   const externalModules = getExternalModules(pkg);
   return externalModules.map((module) => new RegExp(`^${module}(\\/.+)*$`));
 };
 
-const getBanner = (pkg) => {
-  const text = `
-  OpenShift Dynamic Plugin SDK
-  ${pkg.repository.url.replace(/\.git$/, '')}
-
-  ${pkg.name} version ${pkg.version}
-  build date ${new Date().toLocaleString('en-US', { dateStyle: 'long', timeStyle: 'long' })}
-  git commit ${execSync('git rev-parse HEAD').toString().trim()}
-  `.trim();
-
-  return `/*\n  ${text}\n */\n`;
-};
-
 /**
- * @returns {import('rollup').Plugin}
+ * @param {string} file
  */
-const replaceLodashEsRequire = () => ({
-  name: 'replace-lodash-es-require',
-
-  renderChunk(code, chunk) {
-    return {
-      code: code.replace("require('lodash-es')", "require('lodash')"),
-      map: chunk?.map,
-    };
-  },
-});
+const replaceLodashEsRequire = (file) =>
+  replaceCode({
+    file,
+    replacements: {
+      "require('lodash-es')": "require('lodash')",
+    },
+  });
 
 /**
  * Rollup configuration for generating the library `.js` bundle.
  *
+ * @param {PackageJson} pkg
  * @param {string} inputFile
  * @param {'esm' | 'cjs'} format
  * @returns {import('rollup').RollupOptions}
@@ -59,7 +54,6 @@ export const tsLibConfig = (pkg, inputFile, format = 'esm') => ({
   output: {
     file: 'dist/index.js',
     format,
-    banner: getBanner(pkg),
   },
   external: getExternalModuleRegExps(pkg),
   plugins: [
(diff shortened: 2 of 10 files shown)


packages/common/rollup-configs.js
```diff
@@ -100,5 +100,8 @@ export const dtsLibConfig = (pkg, inputFile) => ({
       summaryOnly: true,
       root: rootDir,
     }),
+    removeFiles({
+      files: ['dist/types'],
```
vojtechszocs, line 104: `dist/types` directory contains a `.d.ts` file per each module (following the package source file structure), which gets bundled into a single `dist/index.d.ts` file. To avoid output artifact clutter, we remove this directory once we're done processing its content.

packages/lib-core/package.json
```diff
@@ -10,16 +10,26 @@
...
     "dist/index.js",
-    "dist/index.d.ts"
+    "dist/index.d.ts",
+    "dist/index.ts38.d.ts",
+    "dist/build-meta.json"
   ],
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "typesVersions": {
+    "<4.0": {
```
vojtechszocs, line 20: @florkbr Let me know if you prefer `>=3.8 <4.0` instead of simply `<4.0`.

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions

TypeScript compiler looks at `typesVersions` first and when no matching entry is found, falls back to `types`.

Since we provide a single `.d.ts` bundle, we resolve `index.d.ts` into a single replacement file (instead of using a pattern).

packages/lib-utils/package.json
```diff
@@ -10,7 +10,9 @@
   },
   "files": [
     "dist/index.js",
-    "dist/index.d.ts"
+    "dist/index.d.ts",
+    "dist/index.css",
```
vojtechszocs, line 14: @florkbr @fhlavac This should fix the issue of missing CSS :smiley: 

It's generated, but it wasn't part of `files` and therefore wasn't published.

package.json
```diff
@@ -70,7 +71,9 @@
     "eslint-plugin-promise": "^5.1.1",
     "eslint-plugin-react": "^7.27.0",
     "eslint-plugin-react-hooks": "^4.3.0",
+    "fs-extra": "^10.1.0",
     "jest": "^27.5.1",
+    "lodash": "^4.17.21",
```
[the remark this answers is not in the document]
florkbr: @vojtechszocs I thought we had a webpack rule that converted all imports for non `lodash-es` imports to `lodash-es`?  

**Edit** never mind - I see we have it below `const replaceLodashEsRequire = (file) =>`

package.json
```diff
@@ -90,6 +93,7 @@
     "ts-jest": "^27.1.4",
     "ts-node": "^10.7.0",
     "tslib": "^2.3.1",
+    "type-fest": "^2.18.0",
```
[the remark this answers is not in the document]
florkbr: I would prefer to use the built in typescript types, but it does look like there is a need for this package (100,000,000+ downloads a week): https://www.npmjs.com/package/type-fest 👍

packages/common/rollup-configs.js
```diff
@@ -59,7 +55,6 @@ export const tsLibConfig = (pkg, inputFile, format = 'esm') => ({
   output: {
     file: 'dist/index.js',
     format,
-    banner: getBanner(pkg),
```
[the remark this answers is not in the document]
florkbr: @vojtechszocs I don't think we should remove the comment from the JS bundle. Can we do both? If someone doesn't know to look for `build-meta.json` it may be confusing.

packages/common/rollup-configs.js
florkbr: @vojtechszocs a few things we may want to consider:
1. Should we log anything during the build process about which files are being removed in case we run into a bug where some unintended files are being removed? Not sure if rollup does this for us with a `debug` flag or similar.
2. What is the risk of leaving the duplicate/nested `.d.ts` files? Is there a risk those will be used instead of the `dist/index.d.ts`?

packages/lib-core/package.json
```diff
@@ -10,16 +10,26 @@
...
+  },
   "scripts": {
     "prepack": "yarn build",
     "prepublishOnly": "yarn test",
-    "build": "rm -rf dist && yarn run -T rollup -c",
+    "build": "rm -rf dist && yarn run -T rollup -c && yarn process-dts",
     "lint": "yarn run -T eslint $INIT_CWD",
-    "test": "yarn run -T test $INIT_CWD --passWithNoTests"
+    "test": "yarn run -T test $INIT_CWD --passWithNoTests",
+    "process-dts": "yarn run -T downlevel-dts dist/index.d.ts dist/index.ts38.d.ts --to=3.8"
```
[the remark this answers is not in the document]
florkbr: Agreed. I'm wondering if it's possible to discourage the use of the 3.8 types somehow (deprecated warning perhaps?). I know if you're consuming the SDK you won't know the 3.8 types are there unless you look (specifically named file).

packages/lib-core/package.json
florkbr: I think since this fix is specifically for OCP until they can upgrade to TS 4, I'd prefer to make the version range as specific as possible (IE `>=3.8 <4.0`). We don't want to encourage use of the legacy types IMO. If the need arises we can always switch back to `<4.0`.

packages/common/rollup-plugins/writeBuildMetadata.js
```diff
@@ -0,0 +1,35 @@
...
+    const now = new Date();
+
+    fs.writeJSONSync(
+      path.resolve(cwd, outputDir, 'build-meta.json'),
+      {
+        name: pkg.name,
+        version: pkg.version,
+        buildDate: now.toLocaleString('en-US', { dateStyle: 'long' }),
+        buildTime: now.toLocaleString('en-US', { timeStyle: 'long' }),
+        gitCommit: execSync('git rev-parse HEAD').toString().trim(),
```
florkbr, line 27: 👍  really useful info. I know we have "version" published in the plugin manifests, but I'd love to see something like this for plugins and apps (would help for debugging issues).

florkbr (commented):
A few comments. Looking good @vojtechszocs! CC @jhadvig

package.json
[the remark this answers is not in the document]
vojtechszocs: I'd also prefer to use the built-in types, but TypeScript devs seem to be very picky when it comes to including additional types into the base library.

Using type-only packages has zero overhead on the generated JavaScript, it only impacts the generated `d.ts` type declarations that may include such types.

packages/common/rollup-configs.js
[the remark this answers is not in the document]
vojtechszocs: Our consumers should use webpack to process their host apps and/or plugins into deployable assets, i.e. plugin SDK libraries are not meant to be loaded directly in the browser.

While I think that having a dedicated metadata file is better, we can keep the comment block in the `.js` bundle.

Note that we don't do that for the `.d.ts` bundle, maybe we should?

packages/common/rollup-configs.js
vojtechszocs: When we build a library, we use two Rollup configs:
- `tsLibConfig` which performs TS to JS transpilation and bundling into `dist/index.js`
  - as a side effect of `"declaration": true` setting, this also generates the `dist/types` content
- `dtsLibConfig` which performs the bundling of `dist/index.d.ts` from ^^ `dist/types` content
  - once this is complete, the `dist/types` content is technically no longer needed

Maybe I'm over-engineering this. I'll remove this code, it may look too confusing.

vojtechszocs: > What is the risk of leaving the duplicate/nested `.d.ts` files? Is there a risk those will be used instead of the `dist/index.d.ts`?

There's no risk, files that we publish to npmjs are https://docs.npmjs.com/cli/v8/commands/npm-publish#files-included-in-package basically `package.json`, `README.md` and files specified in `files` array in `package.json`.

We should also include a per-package `LICENSE` file but that's a follow-up improvement.

packages/lib-core/package.json
vojtechszocs: Agreed, I'll change this to be `>=3.8 <4.0`.

packages/lib-core/package.json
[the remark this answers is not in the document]
vojtechszocs: > I'm wondering if it's possible to discourage the use of the 3.8 types somehow (deprecated warning perhaps?).

Not possible AFAIK :smiley: 

> I know if you're consuming the SDK you won't know the 3.8 types are there unless you look (specifically named file).

TypeScript compiler looks at `typesVersions` before falling back to `types` when determining types for the given package. So if you're using TS 3.8 your project should link to the right type declarations automatically.

packages/lib-core/package.json
florkbr: @vojtechszocs I do not see this change in the diff. For now I'll mark this LGTM. Could we please follow up with the change? Thanks!

vojtechszocs:
Follow-up review comments addressed by #143

PR #138, opened 2022-08-11
Pass options to wsAppSettings when using watch

karelhala:
### Description

When using `useK8sWatchResource` and `useK8sWatchResources` consumers has no chance of setting their own host for the websocket. This PR fixes such problem by allowing to pass websocket config to these hooks, when the WS connection is created this config is later passed down to the config function that operates over it.

In this config we have to allow WSConfig and also FetchConfig because before connecting to websockets we'll pull data from regular GET request. We'll also have to support any extra config passed from the plugin down to host app since the host app knows what options to consume.

### JIRA

https://issues.redhat.com/browse/HAC-2026


florkbr:
Chatted with Karel offline. This change is necessary as we will point hooks to a different URL than what the app is currently configured to operate against. For example, hac-infra configuring the useK8sWatchResource hook to point to KCP, however all other APIs will continue to go to hac-core/akamai. 

For now we've made the types added to the hooks strict. It's possible to use generics in the future to allow more customization, but for now we're keeping it simple.

PR #140, opened 2022-08-22
[HAC-2061] Horizontal navigation for details view

vidyanambiar:
Test with `yarn storybook`.

image

[HAC-2061](https://issues.redhat.com/browse/HAC-2061): This PR is for providing an initial basic horizontal navigation to be used in building the common Details UI component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,8 @@
     "eslint-plugin-promise": "^5.1.1",
     "eslint-plugin-react": "^7.27.0",
     "eslint-plugin-react-hooks": "^4.3.0",
+    "fs-extra": "^10.1.0",
+    "identity-obj-proxy": "^3.0.0",
     "jest": "^27.5.1",
     "lodash": "^4.17.21",
     "lodash-es": "^4.17.21",
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "eslint-plugin-promise": "^5.1.1",
     "eslint-plugin-react": "^7.27.0",
     "eslint-plugin-react-hooks": "^4.3.0",
-    "fs-extra": "^10.1.0",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^27.5.1",
     "lodash": "^4.17.21",
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "rollup-plugin-dts": "^4.0.1",
     "rollup-plugin-import-css": "^3.0.3",
     "start-server-and-test": "^1.14.0",
+    "storybook-addon-react-router-v6": "^0.1.14",
     "ts-jest": "^27.1.4",
     "ts-node": "^10.7.0",
     "tslib": "^2.3.1",
--- a/packages/common/test-configs/jest-config-base.ts
+++ b/packages/common/test-configs/jest-config-base.ts
@@ -10,6 +10,7 @@ const config: InitialOptionsTsJest = {
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json'],
   moduleNameMapper: {
     'lodash-es': 'lodash',
+    '\\.(css|scss)$': 'identity-obj-proxy',
   },
 
   clearMocks: true,
--- a/packages/lib-utils/src/components/horizontal-nav/HorizontalNav.tsx
+++ b/packages/lib-utils/src/components/horizontal-nav/HorizontalNav.tsx
@@ -1,0 +1,52 @@
+import { Tabs, Tab, TabTitleText } from '@patternfly/react-core';
+import React from 'react';
+
+export type Tab = {
+  /** Key for individual tab */
+  key: string | number;
+  /** Title for individual tab */
+  title: string;
+  /** Content for individual tab (provided as a React component) */
+  content: React.ReactElement;
+  /** aria-label for individual tab */
+  ariaLabel: string;
+};
+
+export type HorizontalNavProps = {
+  /** aria-label for all tabs */
+  ariaLabel?: string;
+  /** Properties for tabs */
+  tabs: Tab[];
+};
+
+const HorizontalNav: React.FC<HorizontalNavProps> = ({ ariaLabel, tabs }) => {
+  const activeTab = tabs && tabs[0] ? tabs[0].key : 0;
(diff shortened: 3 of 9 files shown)


packages/lib-utils/src/components/horizontal-nav/HorizontalNav.tsx
```diff
@@ -0,0 +1,52 @@
+import { Tabs, Tab, TabTitleText } from '@patternfly/react-core';
+import React from 'react';
+
+export type Tab = {
```
spadgett, line 4: It would be good to have a URL path so that users can link to specific tabs (and we'd need this if we wanted to swap out the tab implementation for HAC core's in OpenShift console).

package.json
```diff
@@ -70,6 +70,7 @@
     "eslint-plugin-promise": "^5.1.1",
     "eslint-plugin-react": "^7.27.0",
     "eslint-plugin-react-hooks": "^4.3.0",
+    "identity-obj-proxy": "^3.0.0",
```
florkbr, line 73: @vidyanambiar would you mind explaining what this new dependency is for? Thanks!

florkbr:
@vidyanambiar merge conflict :(

package.json
vidyanambiar: @florkbr This dependency helps us mock css modules for Jest using an ES6 proxy. I got this idea from Filip's PR for unit testing the list view (#116). Without this change, I noticed the HorizontalNav tests failing on an error in the `backdrop.css` file within patternfly react-styles. 
```
 FAIL   lib-utils  packages/lib-utils/src/components/horizontal-nav/HorizontalNav.test.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/vidyanambiar/Documents/hac-stuff/dynamic-plugin-sdk/node_modules/@patternfly/react-styles/css/components/Backdrop/backdrop.css:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){.pf-c-backdrop {
                                                                                      ^

    SyntaxError: Unexpected token '.'

      at Runtime.createScriptFromCode (../../node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (../../node_modules/@patternfly/react-styles/css/components/Backdrop/backdrop.js:3:1)
```

vidyanambiar:
@spadgett @florkbr I've pushed up a commit that allows the tabs to be rendered based on the routing path. I'm not sure if this is the approach you were getting at Sam. I think one reason I couldn't replicate adding new routes for the tabs was that the tabs and their content are being rendered from Patternfly Tab component unlike console where the NavBar and page/content rendering is separate and controlled by us. 
Storybook view with routing as you click different tabs:
Screen Shot 2022-08-24 at 10 22 03 AM
I also tested this with hac-core using [this branch](https://github.com/openshift/hac-core/compare/main...vidyanambiar:hac-core:testHorizontalNav?expand=1).

Let me know if this solution looks fine or if you prefer that I swap out using the patternfly tabs for a different implementation more similar to console with possibly more control over the routing.  Thank you!

package.json
```diff
@@ -71,6 +71,8 @@
     "eslint-plugin-promise": "^5.1.1",
     "eslint-plugin-react": "^7.27.0",
     "eslint-plugin-react-hooks": "^4.3.0",
+    "fs-extra": "^10.1.0",
```
vojtechszocs, line 74: Hm, why do we need this dependency?

vidyanambiar: @vojtechszocs Sorry, maybe that was pushed up by accident. It's not needed and I've removed it. Thanks!

package.json
florkbr: @vidyanambiar thanks for explaining.

For now this change seems fine. We may want to consider _correctly_ loading these CSS files in the future (versus using `identity-obj-proxy`) in our unit tests. I'm mainly thinking if we ever introduce conditional style logic or if we'd like to protect ourselves from broken style sheets on upgrades in the future (would our build/tests catch it?). We don't necessarily want to test Patternfly's styles however, nor would we really do that from a unit test perspective - perhaps visual snapshot testing with a tool such as https://loki.js.org/.

packages/common/test-configs/jest-config-base.ts
```diff
@@ -10,6 +10,7 @@ const config: InitialOptionsTsJest = {
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json'],
   moduleNameMapper: {
     'lodash-es': 'lodash',
+    '\\.(css|scss)$': 'identity-obj-proxy',
```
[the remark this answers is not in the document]
vidyanambiar: @florkbr From what I can tell this would apply across the board and not just to node_modules.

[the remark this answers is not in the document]
florkbr: @vidyanambiar alright - I'm not sure if this will have unintended consequences so I'd like to follow up on this in the future. Can you open a ticket? Specifically to either:
1. Scope down this rule somehow
2. Actually enable our Jest config to properly load css/scss

florkbr:
For now I think this is a good start for the nav. There will be additional changes required in the future.

vidyanambiar:
Thank you for the review, @florkbr. I've opened this follow-up story to evaluate mocking of css modules for tests - https://issues.redhat.com/browse/HAC-2168

PR #145, opened 2022-08-26
Add missing row key

fhlavac:
This should fix the warning thrown by `VirtualizedTable` saying that the row key is missing.

@vidyanambiar

### DIFF
--- a/packages/lib-utils/src/components/table/VirtualizedTable.tsx
+++ b/packages/lib-utils/src/components/table/VirtualizedTable.tsx
@@ -244,7 +244,7 @@ const VirtualizedTable: React.FC<VirtualizedTableProps<AnyObject>> = ({
             ))) || (
             <Tbody>
               {data.map((item, index) => (
-                <Tr>
+                <Tr key={`row-${item.id}`}>
                   {onSelect && (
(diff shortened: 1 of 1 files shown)


packages/lib-utils/src/components/table/VirtualizedTable.tsx
```diff
@@ -244,7 +244,8 @@ const VirtualizedTable: React.FC<VirtualizedTableProps<AnyObject>> = ({
             ))) || (
             <Tbody>
               {data.map((item, index) => (
-                <Tr>
+                // eslint-disable-next-line react/no-array-index-key
+                <Tr key={`row-${index}`}>
```
florkbr, line 248: @fhlavac we really shouldn't be using indexes for `key`. See https://reactjs.org/docs/lists-and-keys.html#keys.

It would be best to use ID's from the data if possible.

florkbr:
Chatted with @fhlavac offline - he is going to investigate adding a fallback ID if `data.id` is not present (using a RFC UUID tool and much earlier in the table instantiation). Otherwise we could get `row-undefined` for some data types. :shipit:

PR #148, opened 2022-08-30
Add uuid to table rows

fhlavac:
https://issues.redhat.com/browse/HAC-2127

this adds a uuid to each row data item if not defined

@florkbr

### DIFF
--- a/packages/lib-utils/package.json
+++ b/packages/lib-utils/package.json
@@ -39,6 +39,7 @@
     "immutable": "^3.8.2",
     "lodash-es": "^4.17.21",
     "pluralize": "^8.0.0",
-    "typesafe-actions": "^4.4.2"
+    "typesafe-actions": "^4.4.2",
+    "uuid": "^8.3.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "@types/react-router": "^5.1.17",
     "@types/react-router-dom": "^5.3.3",
     "@types/react-virtualized": "^9.21.21",
+    "@types/uuid": "^8.3.4",
     "@typescript-eslint/eslint-plugin": "^5.3.1",
     "@typescript-eslint/parser": "^5.3.1",
     "babel-loader": "^8.2.5",
@@ -94,6 +95,7 @@
     "tslib": "^2.3.1",
     "type-fest": "^2.18.0",
     "typescript": "~4.4.4",
+    "uuid": "^8.3.2",
     "webpack": "^5.69.1"
   }
 }
--- a/packages/lib-utils/src/components/table/VirtualizedTable.tsx
+++ b/packages/lib-utils/src/components/table/VirtualizedTable.tsx
@@ -1,3 +1,4 @@
+import { v4 as uuidv4 } from 'uuid';
 import type { AnyObject } from '@monorepo/common';
 import type { IAction } from '@patternfly/react-table';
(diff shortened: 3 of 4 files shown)


packages/lib-utils/package.json
```diff
@@ -28,12 +28,14 @@
     "@patternfly/react-core": "^4.202.16",
     "@patternfly/react-table": "^4.71.16",
     "@patternfly/react-virtualized-extension": "^4.53.2",
+    "@types/uuid": "^8.3.4",
```
vojtechszocs, line 31: We only need `uuid` in `peerDependencies`, no need for `@types/uuid` here.

Type only dependencies are typically development dependencies. We bundle any referenced vendor types in our own `d.ts` bundles.

fhlavac: without the types the build is failing:
```
➤ YN0000: [!] (plugin typescript) Error: @rollup/plugin-typescript TS7016: Could not find a declaration file for module 'uuid'. '/go/src/github.com/openshift/dynamic-plugin-sdk/node_modules/uuid/dist/index.js' implicitly has an 'any' type.
➤ YN0000:   Try `npm i --save-dev @types/uuid` if it exists or add a new declaration (.d.ts) file containing `declare module 'uuid';`
➤ YN0000: src/components/table/VirtualizedTable.tsx (1:30)
➤ YN0000: 
➤ YN0000: 1 import { v4 as uuidv4 } from 'uuid';
➤ YN0000:                                ~~~~~~
```
do you think there is another way to fix it?

vojtechszocs: @fhlavac AFAIK, we're using the `uuid` library as a helper to generate a fallback ID in `VirtualizedTable` component.

I'd put it in the same category as `lodash` or `immutable` i.e. put it into `dependencies`.

```diff
diff --git a/packages/lib-utils/package.json b/packages/lib-utils/package.json
index 6b217bd..a703f4b 100644
--- a/packages/lib-utils/package.json
+++ b/packages/lib-utils/package.json
@@ -28,19 +28,18 @@
     "@patternfly/react-core": "^4.202.16",
     "@patternfly/react-table": "^4.71.16",
     "@patternfly/react-virtualized-extension": "^4.53.2",
-    "@types/uuid": "^8.3.4",
     "react": "^17.0.2",
     "react-redux": "^7.2.2",
     "react-router-dom": "^6.3.0",
     "react-virtualized": "^9.22.3",
     "redux": "^4.1.2",
-    "redux-thunk": "^2.4.1",
-    "uuid": "^8.3.2"
+    "redux-thunk": "^2.4.1"
   },
   "dependencies": {
     "immutable": "^3.8.2",
     "lodash-es": "^4.17.21",
     "pluralize": "^8.0.0",
-    "typesafe-actions": "^4.4.2"
+    "typesafe-actions": "^4.4.2",
+    "uuid": "^8.3.2"
   }
 }
```

packages/lib-utils/package.json
```diff
@@ -36,9 +36,11 @@
     "redux-thunk": "^2.4.1"
   },
   "dependencies": {
+    "@types/uuid": "^8.3.4",
```
vojtechszocs, line 39: No need for type related dependencies here, any `uuid` related types we use in our library are automatically included in our library's `d.ts` bundle.

vojtechszocs (changes requested):

PR #150, opened 2022-08-31
Investigate CI hangup on #116

vojtechszocs:


vojtechszocs:
With #165 and openshift/release#32497 merged, `ci-operator` builds should now use a builder image that comes with Node.js `v16.17.1` with increased heap limit via `--max-old-space-size=4096` option.

PR #151, opened 2022-09-06
[HAC-2113] LabelList utility component

vidyanambiar:
Jira issue: [HAC-2113](https://issues.redhat.com/browse/HAC-2113)

This PR is for providing a LabelList utility that can be used to construct the details page for a resource.

Test with `yarn storybook`


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
+  // Use IntelliSense to learn about possible attributes.
```
florkbr, line 2: @vidyanambiar did you intend to check in this vscode config?

vidyanambiar: @florkbr Yes - I thought it would make it easy to debug in the future. If it isn't helpful, I can pull it out. LMK

PR #152, opened 2022-09-12
Pass correct props to rowActions onClick

fhlavac:
This adds passing row data as row actions' `onClick` function params.

Thanks @kdoberst for reporting that!


kdoberst:
Looks good to me - this should pass the info that I need.

kdoberst:
@vojtechszocs  the label is actually
/approve

PR #153, opened 2022-09-12
Use FEC toolbar in ListView

fhlavac:
https://issues.redhat.com/browse/HAC-1714

Replaced custom toolbar used in ListView with one from https://github.com/RedHatInsights/frontend-components to make it easier to be maintained in the future. Also added a possibility to define global toolbar actions using `globalActions` property

actions

https://github.com/RedHatInsights/frontend-components/tree/master/packages/components/src/PrimaryToolbar


packages/lib-utils/src/components/list-view/ListView.tsx
```diff
@@ -1,38 +1,26 @@
...
-  ToolbarContent,
-  ToolbarItem,
-  ToolbarItemVariant,
-} from '@patternfly/react-core';
-import { FilterIcon } from '@patternfly/react-icons';
+import { Pagination, PaginationVariant } from '@patternfly/react-core';
+import type {
+  ConditionalFilterItem,
+  FilterChipGroup,
+} from '@redhat-cloud-services/frontend-components';
```
vojtechszocs, line 6: If we only intend to consume a couple of React components and types from ^^ library, wouldn't it make more sense to mark this library as non-external? This way, the code we import from it would become part of the generated SDK code (basically treating it as a dev dependency).

vojtechszocs (changes requested):

packages/lib-utils/package.json
```diff
@@ -36,6 +36,8 @@
     "redux-thunk": "^2.4.1"
   },
   "dependencies": {
+    "@redhat-cloud-services/frontend-components": "^3.9.10",
+    "classnames": "^2.3.1",
```
vojtechszocs, line 40: Do we really use this in our code? If not, we can remove this explicit dependency.

fhlavac:
Will prepare a migration PR for HAC infra after https://github.com/openshift/hac-infra/pull/27 is merged and stable

fhlavac:
Closing this until we have a new component repository which should resolve the CI failures

PR #162, opened 2022-09-27
Tune ci-operator build root Dockerfile

vojtechszocs:
Fedora 36 [comes with OpenSSL version 3.0](https://packages.fedoraproject.org/pkgs/openssl/openssl/) but Node.js v16 (current active LTS) [requires OpenSSL version 1.1.1](https://nodejs.org/en/blog/announcements/nodejs16-eol/).

[This patch](https://github.com/nodejs/node/commit/e903cd19edd27eb6375b4decbcfeff045e5e7df0) backports `--openssl-legacy-provider` option to Node.js v16 for interop with older OpenSSL version, but its usage is not recommended due to security concerns.

Given that Node.js v16 will not change the supported OpenSSL version, this PR updates the CI build root Dockerfile to install Node.js v18 via [Node Version Manager](https://github.com/nvm-sh/nvm), providing full control over the Node.js runtime and npm package manager.

### DIFF
--- a/test-prow-e2e.sh
+++ b/test-prow-e2e.sh
@@ -18,7 +18,7 @@ yarn build-libs
 yarn build-samples
 
 # Create a virtual X11 display via Xvfb for use with Cypress E2E testing
-Xvfb :99 -screen 0 1920x1080x24 &
+Xvfb :99 -screen 0 1920x1080x24 2>&1 > /dev/null &
 export DISPLAY=':99.0'
 
 # Start servers for sample app and sample plugin, run E2E tests and shut down the servers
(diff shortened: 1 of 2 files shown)


test-prow-e2e.sh
```diff
@@ -18,7 +18,7 @@ yarn build-libs
 yarn build-samples
 
 # Create a virtual X11 display via Xvfb for use with Cypress E2E testing
-Xvfb :99 -screen 0 1920x1080x24 &
+Xvfb :99 -screen 0 1920x1080x24 2>&1 > /dev/null &
```
vojtechszocs, line 21: This redirects both stdout and stderr to `/dev/null` to avoid log pollution with non-actionable warnings like
```
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Could not resolve keysym XF86EmojiPicker
Errors from xkbcomp are not fatal to the X server
```

PR #163, opened 2022-09-28
Add GettingStartedCard

rhamilto:
This is a port of `GettingStartedCard` from https://github.com/openshift/hac-dev/pull/141 with the addition of props `isDismissable` for cases where the card should not be dismissable (e.g., Workspaces list page) and `cardClassName` so additional styling can be applied to the card if needed (e.g., to add margin to the bottom).

localhost_6006__path=_story_sdk-utility-components-gettingstartedcard--primary args=cardClassName_


rhamilto:
cc: @rottencandy

packages/lib-utils/src/components/getting-started-card/GettingStartedCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const keys = storageKeys && typeof storageKeys === 'object' ? storageKeys : {};
+  const isDismissed = keys[localStorageKey];
+
+  return !isDismissed ? (
+    <Card className={cardClassName}>
+      <Split>
+        {imgSrc && (
+          <SplitItem className={imgClassName}>
+            <img src={imgSrc} alt={imgAlt} />
```
rottencandy, line 50: We ended up adding a default size for the img in hac-dev because it was the most common size

https://github.com/openshift/hac-dev/blob/342c26553fdcc9840fc1099c763f1349892f7627/src/components/GettingStartedCard/GettingStartedCard.scss#L2-L5

But maybe it's not needed here?

packages/lib-utils/src/components/getting-started-card/GettingStartedCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+          </SplitItem>
+        )}
+        <SplitItem isFilled>
+          <CardHeader>
+            <CardTitle>{title}</CardTitle>
+            {isDismissable && (
+              <CardActions>
+                <Button
+                  variant="plain"
+                  aria-label="close"
```
rottencandy, line 60: nit:
```suggestion
                  aria-label="Hide card"
```

rottencandy (commented):
Added some comments based on a bugfix we had later https://github.com/openshift/hac-dev/pull/151

packages/lib-utils/src/components/getting-started-card/GettingStartedCard.tsx
rhamilto: If the SVG has `width` and `height` attributes within, a size is not needed as it is embedded in the SVG.  I noticed the SVG used in hac-dev is missing these attributes.

rottencandy: Did not know that, thanks.
Those svgs are just temporary placeholders so I think they're fine for now.

packages/lib-utils/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { useEventListener } from './useEventListener';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const tryJSONParse = <T = any>(data: string): string | T => {
```
florkbr, line 5: @rhamilto I think the way in which we are calling this function means we do not need the generic to have a default of `any` here.

So we'd just have:
```ts
const tryJSONParse = <T>(data: string): string | T => {
  try {
    return JSON.parse(data);
  } catch {
    return data;
  }
};
```

... since below useLocalStorage expects `T` to extend `object`

packages/lib-utils/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,36 @@
...
+  try {
+    return JSON.parse(data);
+  } catch {
+    return data;
```
florkbr, line 9: This seems a bit odd that we would return the original string value we were trying to parse without any indication in the hook that something went wrong trying to parse the localStorage. I'd be curious how the gettingStartedCard would behave in that scenario. Would it continue to re-render every page refresh since we cannot correctly parse locally storage to indicate it was dismissed already? 

 For now this is ok, but we may need to revisit it - perhaps in the component here: https://github.com/openshift/dynamic-plugin-sdk/pull/163/files#diff-de59ca7b3bb227b3f0fe8cb00f30a4e840c36509c06bc9a49d7761ab250160ebR42 to handle if the parse failed.

rhamilto: I'll have to defer to @rottencandy as he's the actual author of this code.

packages/lib-utils/src/hooks/useLocalStorage.ts
rhamilto: Updated.

packages/lib-utils/src/hooks/useLocalStorage.ts
rottencandy: I was originally hoping that we'd want to be able to store both objects and strings in local storage. But perhaps it makes sense to restrict it to one type of data.

rottencandy: But yeah in this case the card would render again on refresh if the local storage lookup fails.

florkbr: Gotcha @rottencandy. I think since hac-dev is already behaving in this manner we can merge this and open a follow up ticket to revisit the `tryJSONParse` behavior when parsing JSON fails (and how the card responds to it).

florkbr: https://issues.redhat.com/browse/HAC-2285

PR #165, opened 2022-09-30
Use CentOS 8 Stream based CI build root image

vojtechszocs:
- [x] modify build root image `Dockerfile.ci-operator-buildroot`
  - now based on `centos:stream8` which comes with OpenSSL 1.x
  - now using Node.js `v16.17.1` which requires OpenSSL 1.x
  - added missing test related dependencies
- [x] add build root test image `Dockerfile.ci-operator-buildroot-test`
  - used to verify that test scripts work in CI env. (`test.sh`, `test-prow-e2e.sh`)
- [x] modify basic test script `test.sh`
  - added `--max-old-space-size=4096` Node.js option
  - make sure `prow-codecov.sh` runs only in CI env.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 set -exuo pipefail
 
-print_error() { printf "%s\n" "$*" >&2; }
+# Setup environment
+export NODE_OPTIONS="--max-old-space-size=4096"
 
 # Print system information
 echo "node $(node -v)"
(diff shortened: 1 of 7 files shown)


test.sh
```diff
@@ -20,8 +21,7 @@ yarn lint
...
 yarn test
 
-# Run end-to-end tests
-# TODO
-
-# Upload code coverage
-./prow-codecov.sh 2>/dev/null
+# Upload code coverage (CI environment only)
+if [[ -n "${CI:=}" ]]; then
+  ./prow-codecov.sh 2>/dev/null
```
florkbr, line 26: @vojtechszocs I realized after we talked that this script only runs for non-local jobs. See the check here: https://github.com/openshift/dynamic-plugin-sdk/blob/main/prow-codecov.sh#L14

So this new check is redundant...

PR #166, opened 2022-10-05
Add lib-extensions package

vojtechszocs:
- Add `lib-extensions` package
  - provides core extension types for dynamic plugins
  - can be used to provide other extension types if needed
- Updated package dependencies
  - `lib-extensions` and `lib-utils` has peer dependency on `lib-core`
  - Peer dependencies are not transitive by default [1][2]
  - Due to above, `lib-extensions` and `lib-utils` now include peer dependencies of `lib-core`
  - `lib-extensions` and `lib-utils` now use `"@openshift/dynamic-plugin-sdk": ">=1.0.0-alpha"` which explicitly targets the `alpha` pre-release of `lib-core` package

[1] https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0
[2] https://github.com/yarnpkg/berry/issues/3#issuecomment-872535638

> my final conclusion is that implicit inheritance of peer dependencies is an antipattern that only provides some convenience at the cost of soundness and maintainability (as in, maintenance of a package manager)


packages/common/test-configs/jest-config-base.ts
```diff
@@ -13,6 +13,7 @@ const config: InitialOptionsTsJest = {
     '\\.(css|scss)$': 'identity-obj-proxy',
   },
 
+  passWithNoTests: true,
```
florkbr, line 16: FWIW I originally added this to our package.json as the publish step usually failed for lib-core etc since there were no tests. We probably eventually should change this to only be packages we expect there to be no tests for, versus blanket allowing it across the board.

packages/lib-core/package.json
```diff
@@ -28,24 +28,14 @@
...
-    "react": "^17.0.2",
-    "react-router": "^5.2.1",
-    "redux": "^4.1.2"
+    "react": "^17.0.2"
   },
   "dependencies": {
     "lodash-es": "^4.17.21",
     "yup": "^0.32.11"
-  },
-  "peerDependenciesMeta": {
```
florkbr, line 43: Did we not actually use react-router/redux in the core package? Not sure how these got in here.

packages/lib-core/src/index.ts
```diff
@@ -7,14 +6,12 @@ export { useExtensions } from './store/useExtensions';
...
-  LoadedExtension,
   EncodedExtension,
+  LoadedExtension,
   ResolvedExtension,
 } from './types/extension';
 export { PluginEventType, PluginInfoEntry, PluginConsumer, PluginManager } from './types/store';
 export { PluginManifest, PluginRuntimeMetadata, LoadedPlugin } from './types/plugin';
-
-// Core extension types, to be exported via lib-extensions package
-export * from './extensions';
```
florkbr, line 20: @vojtechszocs we originally went through the ebb and flow of removing this `export *`, then adding it back. I believe @karelhala added this to prevent missing extensions... I want to confirm - since the only thing remaining is the core extensions, the concern of missing some is less prevalent (as lib-extensions has the `export *` now)? If so - I think this is ok.

packages/lib-core/package.json
florkbr: Ah - I think I get it. These were for the extensions we moved. Nvm.

packages/lib-extensions/src/extensions/core/actions.ts
```diff
@@ -1,6 +1,6 @@
+import type { CodeRef, Extension } from '@openshift/dynamic-plugin-sdk';
```
florkbr, line 1: This looks good - do not change it. It will prevent us from breaking consumers of the SDK today 👍 

 Just a longer term question to pick your brain @vojtechszocs  - do you eventually think we'd switch to pull the extensions directly through the `@openshift/dynamic-plugin-sdk-extensions` package and not `lib-core`?

packages/lib-utils/package.json
```diff
@@ -24,12 +24,13 @@
     "test": "yarn run -T test $INIT_CWD"
   },
   "peerDependencies": {
-    "@openshift/dynamic-plugin-sdk": "^1.0.0",
+    "@openshift/dynamic-plugin-sdk": ">=1.0.0-alpha",
```
florkbr, line 27: Does this help us get around the entire alpha versioning issue (where carat does not correctly pick up `alpha1` vs `alpha2` automatically)? If so, we should probably update hac-core so we don't need to change this manually each time we merge/publish the SDK.

florkbr (commented):
Overall looks good to me. I'd like to quickly test this build in hac-core/hac-dev and if all continues to work I'll merge and publish with the other SDK packages (we have some fixes we need published in lib-utils). 

@vojtechszocs we may also want to consider announcing this in the dynamic plugin sync call!

packages/common/test-configs/jest-config-base.ts
vojtechszocs: Well, if a package has no tests, we can either fail the build, or let it pass.

I think letting it pass is more sensible in terms of defaults. The default imposed by Jest is more along the lines of "you have a project with test infrastructure set up, not having any tests doesn't make much sense, therefore it's treated as an error".

Right now, `lib-core` and `lib-webpack` are missing tests but we will add them later on. As for the upcoming `lib-extensions` whose scope is very narrow, tests may not even be necessary for that package, unless we add some logic or we want to get really fancy and start "unit testing" the TS types 😃 also consider the fact that we use the same test infrastructure across multiple packages within a monorepo.

👍 on changing it later on, only for packages where we don't expect any tests.

packages/lib-core/package.json
vojtechszocs: Correct, these were only used by extension types, which got moved into a separate package.

They were marked as optional via `peerDependenciesMeta` because you could use the core package in your host application or plugins without using any/all of the extension types.

packages/lib-core/src/index.ts
vojtechszocs: This "export all extension types" was just moved to `lib-extensions` package, becoming its main scope and purpose.

Since the scope of `lib-extensions` is very narrow, having `export *` in its main entry point is OK, i.e. we generally want to expose everything from that package to consumers.

However, the scope of `lib-core` is much broader - basically it should provide essential APIs to
- **[host apps]** set up runtime plugin infrastructure via `PluginStore` and `PluginStoreProvider`
- **[host apps and/or plugins]** allow consuming extensions via React hooks like `useExtensions`

I generally avoid `export *` in a library's entry point because it exposes everything down to all consumers. I think that API surface should be as small as possible, which (among other things) minimizes the amount of crazy/unexpected ways people use the library which then sparks more issues and maintenance efforts.

Having full control over what we export (via named exports) is basically a way to minimize the API surface. I agree that it's less convenient, but I think it's better in the long term.

packages/lib-extensions/src/extensions/core/actions.ts
vojtechszocs: If you mean the `CodeRef` and `Extension` types, these are integral to the core library, which "owns" these types.

I'd keep them there and have `lib-extensions` peer depend on `lib-core` as proposed in this PR.

An alternative would be to bundle (copy) any types from `lib-core` in the extension package's `d.ts` bundle, not sure if that's the right approach as it would effectively duplicate the type definitions.

vojtechszocs: To clarify, by integral I mean core APIs like `useExtensions` and `useResolvedExtensions.ts` depend on these types.

packages/lib-utils/package.json
vojtechszocs: Yes, I think this should help us with the issue where pre-release versions are not picked up by regular caret `^` or tilde `~` ranges 😃 

As per SemVer rules, version `1.0.0-alpha1` comes "before" version `1.0.0` because it's a pre-release version. Therefore, version `1.0.0-alpha1` may not match the constraint `^1.0.0`.

In version range `>=1.0.0-alpha`, the explicit `-alpha` should cause it to pick up alpha pre-releases only. According to SemVer rules, pre-release identifiers are compared alpha numerically.

florkbr: I think ideally I'd like to remove the `alpha` postfix all-together. We can do that as a follow up.

vojtechszocs:
Note to self: need to update the README to link to the newly added package.

PR #168, opened 2022-10-11
HAC-2193: Create resource summary components

kdoberst:
Creates numerous components to help display a resource summary (left column below) and an OverviewPage which will enforce a two-column layout.

Screen Shot 2022-10-11 at 12 47 46 PM

Note: Not all fields are available yet, just the ones needed for the workspace details page in the HAC-Infra project.  

This PR also includes the following changes:
  - Removes storybook files (.stories files) from test coverage.  This speeds up the time to run the tests
  - Adds jest-axe so accessibility can be checked in unit tests
  - Add DetailsPage to index.ts so it can be used outside the SDK


kdoberst:
Note that there is one failing test that is unrelated to this PR.  It is also located on the main branch

kdoberst:
Question:  The resource summary is available outside the SDK.  In Hac Infra there isn't a need to call that component directly.  Instead, Hac Infra calls the OverviewPage component - which is just a wrapper component for resource summary but enforces a 2-column layout.

rhamilto (commented):
I haven't reviewed at each individual file, but did notice it seems like *.tsx filenames in `resource-summary` should all use camelcase for consistency.

packages/lib-utils/src/components/resource-summary/resource-summary.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+        uid: '8f5ce882-4a20-47da-ad6a-16d0c4846406',
+      },
+    ],
+  },
+};
+
+describe('ResourceSummary', () => {
+  it('is accessible', async () => {
+    const { container } = render(<ResourceSummary resource={k8Object} />);
+    const results = await axe(container);
```
florkbr, line 41: @kdoberst out of curiosity - are you suggesting having separate specs for axe testing, or should we incorporate these assertions into our normal tests? I'm wondering how this will scale when many props/permutations are being tested in larger components going forward.

packages/lib-utils/src/components/resource-summary/resource-summary.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const k8Object2: K8sResourceCommon = {
+      apiVersion: 'v1beta1',
+      apiGroup: 'tenancy.kcp.dev',
+      kind: 'Workspace',
+      metadata: {
+        name: 'demo-workspace',
+      },
+    };
+
+    const { container } = render(<ResourceSummary resource={k8Object2} loaded={false} />);
```
florkbr, line 96: @kdoberst wondering if we could utilize lodash `omit` on the `ownerReferences` key so we don't need separate objects for this: https://lodash.com/docs/4.17.15#omit

packages/lib-utils/src/components/resource-summary/details-item.tsx
```diff
@@ -0,0 +1,49 @@
...
+  hideEmpty,
+  loaded = true,
+  label,
+  resource,
+  path = '',
+}) => {
+  const objValue = path ? _.get(resource, path) : undefined;
+
+  if (hideEmpty && objValue === undefined && !children) {
+    return null;
```
florkbr, line 34: +1 for returning fast

packages/lib-utils/src/components/label-list/LabelList.tsx
```diff
@@ -29,7 +29,7 @@ export const LabelList: React.FC<LabelListProps> = ({
 }) => {
   const defaultColor = 'purple';
 
-  return _.isEmpty(labels) ? (
+  return _.isEmpty(labels) || !labels ? (
```
florkbr, line 32: @kdoberst - curious - what case was `_.isEmpty` missing here?
image

packages/lib-utils/src/components/resource-summary/resource-summary.test.tsx
kdoberst: I've seen it done both ways.  I tend to like to have a test(s) whose primary function is to check for accessibility.  It also makes it very evident in the list of tests run that accessibility was checked vs. hidden as an assert.

packages/lib-utils/src/components/resource-summary/resource-summary.tsx
```diff
@@ -0,0 +1,61 @@
...
+        path="metadata.creationTimestamp"
+        loaded={loaded}
+      >
+        {/* Leaving raw for now - there will need to be a special datetime formatter component */}
+      </DetailsItem>
+
+      <DetailsItem
+        label="Owner"
+        resource={resource}
+        path="metadata.ownerReferences"
```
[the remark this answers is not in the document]
kdoberst: We don't want them passed in.  The purpose of the resource summary is that the labels are identical regardless of where they are used.  I'm not sure it adds a lot, but I can save them as constants at the top of the file

kdoberst:
Responding to 

> I haven't reviewed at each individual file, but did notice it seems like *.tsx filenames in resource-summary should all use camelcase for consistency.

I've found the file naming inconsistent through out the repo.  Do we have naming standards we can follow?

packages/lib-utils/src/components/resource-summary/resource-summary.tsx
```diff
@@ -0,0 +1,61 @@
+import {} from '@patternfly/react-core';
```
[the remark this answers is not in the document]
kdoberst: good question.  My IDE did a bad job of binging me on empty imports.  I'll fix it and reconfigure my IDE lint configs.

packages/lib-utils/src/components/resource-summary/resource-summary.test.tsx
kdoberst: I'm honestly not a fan lodash because it adds overhead that isn't needed in most cases (unless you are doing full-on functional programming).  I'm also a fan of having completely separate objects for testing vs modifying the data.  Makes the tests easier to read and modify without worrying about changing the data messing up other tests.

I can change it if you like

packages/lib-utils/src/components/resource-summary/details-item.test.tsx
```diff
@@ -0,0 +1,99 @@
+import { render, screen } from '@testing-library/react';
+import { toHaveNoViolations, axe } from 'jest-axe';
+import React from 'react';
+import '@testing-library/jest-dom';
+import type { K8sResourceCommon } from '../../index';
+import { DetailsItem } from './index';
+
+expect.extend(toHaveNoViolations);
```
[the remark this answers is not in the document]
kdoberst: If there is a test set-up file, you can add it there.  Having said that, my experience in doing anything like this is using CRA's `setupTests.js` file.  Do we have something similar in the repo?

packages/lib-utils/src/components/label-list/LabelList.tsx
kdoberst: Down a few lines there is this code:
```
 {Object.entries(labels).map(([key, value]) => (
```

This returns a type error because this line, in theory, would blow up if labels were undefined.

packages/lib-utils/src/components/resource-summary/details-item.tsx
kdoberst: Always!

packages/lib-utils/src/components/resource-summary/details-item.tsx
```diff
@@ -0,0 +1,49 @@
...
+const DetailsItem: React.FC<DetailsItemProps> = ({
+  children,
+  defaultValue = '-',
+  hideEmpty,
+  loaded = true,
+  label,
+  resource,
+  path = '',
+}) => {
+  const objValue = path ? _.get(resource, path) : undefined;
```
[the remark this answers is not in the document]
kdoberst: I think I was being extraordinarily clear that undefined is the expected value if there wasn't a path.   Ties into my general dislike for lodash.  I'll change it

kdoberst:
Note that I have squshed all commits into a single commit

packages/lib-utils/src/components/resource-summary/details-item.test.tsx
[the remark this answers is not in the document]
kdoberst: Now that I'm curious - I am going to explode CRA and see if I can figure out what they are doing.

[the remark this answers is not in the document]
kdoberst: The short answer is that Yes we do have a test set-up file in the repo.  I added the expect to that file.

packages/lib-utils/src/components/resource-summary/resource-summary.test.tsx
florkbr: Chatted with @kdoberst offline - maybe in the future we revisit this to fold these into our main test suites, but for now it's nice seeing the explicit accessibility tests.

packages/lib-utils/src/components/resource-summary/resource-summary.test.tsx
florkbr: Chatted with Kim - not blocking. Mainly a maintenance concern in the future (3 places we may need to change a key/value over one). That could in theory be a good thing. For now this is fine as is.

PR #169, opened 2022-10-13
Fix test performance

vojtechszocs:
- optimize test config for better performance (*)
  - set [`maxWorkers`](https://jestjs.io/docs/27.x/configuration#maxworkers-number--string) to `1` when in CI environment
  - set [`isolatedModules`](https://kulshekhar.github.io/ts-jest/docs/27.1/getting-started/options/isolatedModules) to `true` regardless of the environment

- keep using Jest `27.x` but update `testing-library` dependencies
  - `@testing-library/react` must be `12.x` because we're using React `17.x`
  - `@testing-library/dom` was removed since it's pulled in via `@testing-library/jest-dom`

- remove `import '@testing-library/jest-dom'` in test files
  - already covered via common setup `packages/common/test-configs/setup-react.ts`

- update [`collectCoverageFrom`](https://jestjs.io/docs/27.x/configuration#collectcoveragefrom-array) to exclude tests and Storybook stories

---

According to https://github.com/kulshekhar/ts-jest/issues/259#issuecomment-888978737 Jest workers run independently from each other, therefore each worker runs TypeScript compilation on its own, which results in slower times (due to multiple parallel TS compiler executions) which are offset by workers running in parallel.

Local testing via `rm -rf /tmp/jest_rs ; yarn test` shows that running all tests (includes code coverage collection) with `isolatedModules: true` is ~3.5 times faster. Since we don't really need full type checking when running tests (already covered by our "build" scripts which fail upon any TS type errors), it makes sense to always use `isolatedModules: true`.


jest.config.ts
```diff
@@ -12,8 +10,13 @@ const config: InitialOptionsTsJest = {
 
   collectCoverage: true,
   coverageDirectory: 'coverage',
-  coverageReporters: ['json', 'lcov', 'text'],
+  coverageReporters: ['lcov'],
```
vojtechszocs, line 13: @florkbr I think we only need the `lcov` reporter for integration with the [CodeCov](https://docs.codecov.com/docs) service.

Side note, `patternfly-react` also uses just the `lcov` reporter, see [here](https://github.com/patternfly/patternfly-react/blob/main/jest.config.js).

package.json
```diff
@@ -10,13 +10,14 @@
...
     "lint": "yarn eslint packages",
-    "test": "jest --logHeapUsage -w 1",
+    "test": "jest --logHeapUsage",
     "test-print-config": "jest --showConfig",
     "test-e2e": "yarn workspace @monorepo/e2e-sample-app run test-e2e",
     "test-e2e-open": "yarn workspace @monorepo/e2e-sample-app run test-e2e-open",
     "test-e2e-ci": "start-server-and-test 'yarn run-samples' '9000|9001' 'yarn test-e2e'",
     "eslint": "eslint --max-warnings 0 --ext .js,.jsx,.ts,.tsx",
     "eslint-print-config": "eslint --print-config",
+    "jest": "TZ=UTC jest",
```
vojtechszocs, line 20: I think it's generally better to set env. variables when calling scripts, rather than inside scripts or programs.

Side note, `patternfly-react` also sets the timezone this way, see [here](https://github.com/patternfly/patternfly-react/blob/main/package.json).

vojtechszocs:
Some more updates incoming.

packages/common/test-configs/jest-config-base.ts
```diff
@@ -5,7 +5,7 @@ const config: InitialOptionsTsJest = {
   preset: 'ts-jest/presets/js-with-ts',
 
   testMatch: ['**/*.test.(js|jsx|ts|tsx)'],
-  transformIgnorePatterns: ['/node_modules/(?!(@patternfly)/)'],
+  transformIgnorePatterns: ['/node_modules/'],
```
vojtechszocs, line 8: FYI this was one of the reasons of the slowdowns, since we instructed `ts-jest` to transform all PatternFly packages by default.

jest.config.ts
florkbr: We may eventually need the other reporters, but I think this is fine for now.

PR #170, opened 2022-10-17
Fix pagination experience in ListView

fhlavac:
Fixes https://issues.redhat.com/browse/HAC-2273

- this should fix displaying all items instead of first X items on the first page
- also added a missing offset reset on the pagination limit change causing pagination break

### DIFF
--- a/packages/lib-utils/src/components/table/VirtualizedTable.tsx
+++ b/packages/lib-utils/src/components/table/VirtualizedTable.tsx
@@ -117,7 +117,9 @@ const VirtualizedTable: React.FC<VirtualizedTableProps<AnyObject>> = ({
 
   const paginateData = (allData: AnyObject[]) => {
     const end =
-      pagination?.offset && pagination?.limit ? pagination.offset + pagination.limit : undefined;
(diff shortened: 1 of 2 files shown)


karelhala (approved):
/lgtm

packages/lib-utils/src/components/table/VirtualizedTable.tsx
```diff
@@ -125,7 +125,9 @@ const VirtualizedTable: React.FC<VirtualizedTableProps> = ({
 
   const paginateData = (allData: AnyObject[]) => {
     const end =
-      pagination?.offset && pagination?.limit ? pagination.offset + pagination.limit : undefined;
+      typeof pagination?.offset === 'number' && pagination?.limit
```
vojtechszocs, line 128: Why do we have to check if `pagination.offset` is really a number? Is this because `0` is falsy value?

fhlavac: @vojtechszocs yeah, that's the reason

vojtechszocs (approved):

PR #177, opened 2022-10-24
Improve plugin name and extension type RegExp pattern

vojtechszocs:
Allow uppercase letters and improve the existing RegExp patterns.

The documented examples will be transformed into unit tests as a follow-up.


sjd78 (approved):
LGTM and plugs a nice hole so openshift-console extensions that have mixed case names can be used.

mturley (approved):

PR #178, opened 2022-10-25
Add action buttons to list view

kdoberst:
This PR adds the ability to add call-to-action (cta) buttons to the list view header.  This is to fix a placement issue of the "add worksapce" button in HAC-Infra.

**Screenshot**
Screen Shot 2022-10-24 at 11 20 52 PM

### DIFF
--- a/packages/lib-utils/src/components/details-page-header/utils/ActionButtons.tsx
+++ b/packages/lib-utils/src/components/details-page-header/utils/ActionButtons.tsx
@@ -3,14 +3,41 @@ import * as _ from 'lodash-es';
 import React from 'react';
 
 export type ActionButtonProp = {
-  label: string;
+  id?: string;
+  label?: string;
   callback: (event: React.MouseEvent) => void;
   isDisabled?: boolean;
   tooltip?: string;
+  variant?: 'primary' | 'secondary' | 'tertiary' | 'link';
 };
 
 export type ActionButtonsProps = {
   actionButtons: ActionButtonProp[];
+};
+
+const ActionButton: React.FC<ActionButtonProp> = ({
+  id,
+  children,
+  callback,
+  isDisabled,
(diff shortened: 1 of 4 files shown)


kdoberst:
I rebased this and add tests.  Just a note, my IDE was giving me typescript errors on the ListView component after rebasing on code that was brought in (aka code I hadn't touched).  

Specifically the columns, isRowSelected, onSelect, and Row props.  All tests passed and I could build the libs, so I didn't touch it.  As you are reviewing this, please check to see these appear.

packages/lib-utils/src/components/details-page-header/utils/ActionButtons.tsx
```diff
@@ -3,41 +3,57 @@ import * as _ from 'lodash-es';
 import React from 'react';
 
 export type ActionButtonProp = {
-  label: string;
+  id?: string;
+  label?: string;
   callback: (event: React.MouseEvent) => void;
   isDisabled?: boolean;
   tooltip?: string;
+  variant?: 'primary' | 'secondary' | 'tertiary' | 'link';
```
florkbr, line 11: TODO (not blocking) can we forward the Patternfly values here instead of hard-coding it on our end? It seems you cannot get a list today it seems.

florkbr: Quick test locally - looks like you can import the ButtonVariant type from PF: `import type { ButtonVariant } from '@patternfly/react-core';` 

We should investigate in the future how to dedupe all of these props with PF props - and consider moving the ActionButtons outside of the details-page-header folder (they could theoretically be consumed independently of the header).

florkbr:
Verified with Kim locally that the variant prop does indeed change the display in the storybook (possibly add story in the future) 

image

PR #182, opened 2022-11-10
Enable HCC remote modules to use as plugins

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-21494

The current functionally is preserved.

### Changes
- enable entry script hashing 
- expand DynamicRemotePlugin to accommodate HCC build plugins
- add the `createOrReloadPlugin` method to allow direct remote webpack container loading
- enable `var` library type alongside with `jsonp` (`jsonp` is the default)
- add some examples to the `sample-app` and `sample-plugin` for debugging purposes

### Example of new plugin metadata
```JSON
{
  "name": "sample-plugin",
  "version": "1.2.3",
  "dependencies": {
    "sample-app": "^1.0.0"
  },
  "extensions": [ ],
  "libraryType": "jsonp",
  "entryScript": "sample-plugin.8f3367c94598d7ce108a.js",
  "runtimeChunkScript": "runtime.js" // only if a single runtime chunk is permitted in webpack config
}
```

cc @vojtechszocs @karelhala @florkbr @fhlavac


Hyperkid123:
OK, I can't cheat as I would like to in the sample app. This is going to have to be a dynamic import:
```
'../../../../sample-plugin/dist/plugin-manifest.json'
```

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -134,22 +158,24 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
       );
     }
 
-    if (compiler.options.output.uniqueName) {
+    if (this.libraryType === 'jsonp' && compiler.options.output.uniqueName) {
```
vojtechszocs, line 161: Why does this apply only to `jsonp` option?

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -134,22 +158,24 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
...
       throw new Error(
         'output.uniqueName option will be set automatically, do not set it in your webpack configuration',
       );
     }
 
-    compiler.options.output.uniqueName = containerName;
+    if(this.libraryType === 'jsonp') {
```
vojtechszocs, line 167: Nit: space after `if` (lib-webpack is not currently linted, we will fix that later) here and elsewhere.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -134,22 +158,24 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
...
     new container.ModuleFederationPlugin({
       name: containerName,
       library: {
-        type: 'jsonp',
-        name: this.entryCallbackSettings.name,
+        type: this.libraryType,
+        name: this.libraryType === 'var' ? containerName : this.entryCallbackSettings.name,
       },
-      filename: REMOTE_ENTRY_SCRIPT,
+      filename: `${containerName}.[fullhash].js`,
```
vojtechszocs, line 178: :+1:

vojtechszocs (changes requested):

packages/lib-core/src/types/store.ts
```diff
@@ -87,6 +87,11 @@ export type PluginStoreInterface = {
    */
   loadPlugin: (baseURL: string) => void;
 
+  /**
+   * Directly create or update dynamic plugin from plugin manifest
```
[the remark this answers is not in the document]
vojtechszocs: Alternatively, we can simply update the existing `loadPlugin` method signature, adding an optional `reload` parameter that would default to `false`.

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -304,6 +314,86 @@ export class PluginLoader {
...
+    baseUrl: string,
+    pluginLoadData: PluginLoadData,
+    // onLoad type does not have stand alone definition
+    onLoad?: (ev?: Event) => void,
+    onError?: OnErrorEventHandler,
+    getDocument: () => typeof document = _.constant(document),
+  ) {
+    const pluginName = 'sample-plugin';
+
+    this.plugins.set(pluginName, pluginLoadData);
```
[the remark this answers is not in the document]
Hyperkid123: Right I can also see I have a constant for the plugin name which should be based on the manifest file.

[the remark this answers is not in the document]
Hyperkid123: Where do you think would be a good place? I am asking specifically for the case when we call the `injectPluginScript` method directly in `PluginStore.createOrReloadPlugin`. The Plugin store does not have access to the plugins map. I don't know if it's a good idea to add the plugins map to public interface.

packages/lib-core/src/types/store.ts
```diff
@@ -87,6 +87,11 @@ export type PluginStoreInterface = {
...
+   */
+  createOrReloadPlugin: (baseUrl: string, manifest: PluginManifest) => Promise<void>;
```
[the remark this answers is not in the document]
Hyperkid123: What if you need to add a new plugin at runtime?

In the HCC, we will have the application register as a plugin initially. We will have to create and provide the metadata at runtime. Is there already a way to add new manifests at runtime?

packages/lib-core/src/types/store.ts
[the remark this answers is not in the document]
Hyperkid123: Again, we might not have all the modules in the plugin registry when we spin up HCC. We will need to access this method directly before all our tenants adopt the SDK in their builds.

packages/lib-webpack/src/types/plugin.ts
```diff
@@ -1,5 +1,6 @@
 import type { PluginRuntimeMetadata } from '@openshift/dynamic-plugin-sdk/src/types/plugin';
 
-export type PluginBuildMetadata = PluginRuntimeMetadata & {
+export type PluginBuildMetadata = Omit<PluginRuntimeMetadata, 'entryScript'> & {
```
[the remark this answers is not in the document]
Hyperkid123: I think there were some issues with validators in the webpack plugins. The entry script is generated by the webpack plugin and is not available at the time the webpack plugin options are validated.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -34,6 +34,12 @@ const validateExtensions = (extensions: EncodedExtension[]) => {
   extensionArraySchema.strict(true).validateSync(extensions);
 };
 
+
+
+function validateVarLibrary(pluginMetadata: string | PluginBuildMetadata) {
```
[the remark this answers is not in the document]
Hyperkid123: Ah, I see now. I saw the options.pluginMetadata can be a string and did no notice it is the parsed to JSON. :+1: we don't need this at all.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -84,10 +90,22 @@ export type DynamicRemotePluginOptions = Partial<{
...
+
+  /**
+   * Select module injection type
+   * 
+   * jsonp is consumed via pre-defined callback. By default __load_plugin_entry__(..)
+   * var is assigned to global variable name
+   * 
+   * Default value: jsonp
+   */
+  libraryType: 'jsonp' | 'var'
```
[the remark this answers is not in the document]
Hyperkid123: I do personally tend to adopt dependency terminology. It makes it easier to find the original docs if you'd like to learn more. But I do understand that in this case, it might not make sense at first glance. I think we can rename it, but I will add a link to the webpack docs.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
Hyperkid123: Is is a left over when there was the `plugin-entry.js` default name. With the hash changes, it can be removed.

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -150,6 +158,24 @@ export class PluginLoader {
 
     this.plugins.set(pluginName, data);
 
+    /**
+     * Inject runtime script if it exists in a separate chunk
+     */
+    if (manifest.runtimeChunkScript) {
```
vojtechszocs, line 164: Loading a separate runtime chunk (if present) should happen right before loading the actual plugin entry script, i.e. after resolving the plugin's dependencies.

This can be done as a follow-up improvement.

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -184,6 +210,19 @@ export class PluginLoader {
         errorCause: e,
       });
     }
+
+    const pluginReadynesPromise = new Promise<void>((resolve) => {
```
vojtechszocs, line 214:
```suggestion
    const pluginReadinessPromise = new Promise<void>((resolve) => {
```

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -304,10 +343,146 @@ export class PluginLoader {
...
+      script.async = true;
+
+      script.onload = async (event) => {
+        if (onLoad) {
+          onLoad(event);
+        }
+        if (pluginLoadData.manifest.libraryType === 'var') {
+          await this.initializeVarModule(pluginName);
+        }
+        return resolve();
```
vojtechszocs, line 373: Is this a short-hand for the following?
```ts
resolve();
return;
```

Both `resolve` and `reject` functions return `void`.

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -304,10 +343,146 @@ export class PluginLoader {
...
+   * Required by HCC plugins.
+   * Simulates the jsonp flow used in the other type of plugins.
+   */
+  async initializeVarModule(pluginName: string) {
+    if (!this.plugins.has(pluginName)) {
+      throw new ErrorWithCause(`No plugin data exist for '${pluginName}'`);
+    }
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const data = this.plugins.get(pluginName)!;
+    const moduleScope = data.manifest.name;
```
vojtechszocs, line 403: Plugin name is required and will always be present in a valid plugin manifest.

In other words, `pluginName` should be equal to `data.manifest.name`.

Thus, we should be able to remove the "else" part below.

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -304,10 +343,146 @@ export class PluginLoader {
...
+        window as unknown as { [key: string]: PluginEntryModule }
+      )[moduleScope];
+      try {
+        await Promise.resolve(entryModule.init(this.options.sharedScope));
+      } catch (error) {
+        data.status = 'failed';
+        this.invokeListeners({
+          success: false,
+          pluginName,
+          errorCause: 'Plugin container initialization failed',
```
vojtechszocs, line 416: `errorCause` should be the thrown error object.

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -304,10 +343,146 @@ export class PluginLoader {
...
+
+  setPluginData(data: PluginLoadData) {
+    this.plugins.set(data.manifest.name, data);
+  }
+
+  static isUniqueScript(
+    scriptURL: string,
+    getDocument: () => typeof document = _.constant(document),
+  ) {
+    return getDocument().head.querySelectorAll(`script[src="${scriptURL}"]`).length === 0;
```
vojtechszocs, line 448: If we're looking for the presence of a `<script>` element with the given `src` we could simply use `querySelector` which returns the first matching element, or `null` if not found.

packages/lib-core/src/store/PluginLoader.ts
```diff
@@ -304,10 +343,146 @@ export class PluginLoader {
...
+  }
+
+  static async loadPluginRuntimeScript(
```
vojtechszocs, line 451: I still wonder if this added complexity is really worth it (supporting a separate webpack runtime chunk) but I guess we need it for some plugins.

We should update the sample plugin to emit a separate runtime chunk so this code can be easily exercised.

Martin, in case of plugin reload - not reloading the runtime chunk script is OK, because its content doesn't change across plugin builds, is that correct?

packages/lib-core/src/types/plugin.ts
```diff
@@ -9,6 +9,9 @@ export type PluginRuntimeMetadata = {
 
 export type PluginManifest = PluginRuntimeMetadata & {
   extensions: Extension[];
+  libraryType?: 'jsonp' | 'var';
```
vojtechszocs, line 12:
```suggestion
  registrationMethod?: 'jsonp' | 'var';
```

packages/lib-webpack/src/types/plugin.ts
```diff
@@ -1,5 +1,6 @@
...
+  name?: string;
```
vojtechszocs, line 4: `name` should already be there (?)

vojtechszocs (changes requested):

packages/lib-core/src/store/PluginLoader.ts
Hyperkid123: Yeah, it's just a short hand.

packages/lib-core/src/store/PluginLoader.ts
Hyperkid123: The runtime chunk will not change. It only provides the ability for additional webpack chunk loading. I was doing some testing where a plugin has changed in the sample app and it worked well.

I think we are at a point where adding some e2e tests is a good idea.

packages/sample-app/src/shared-scope.ts
```diff
@@ -15,15 +17,15 @@
  * })
  * ```
  */
-export const initSharedScope = async (name = 'default') => __webpack_init_sharing__(name);
+export const initSharedScope = async () => __webpack_init_sharing__(SHARED_SCOPE_NAME);
```
florkbr, line 20: 👍 for moving this default to a const for clarity.

packages/lib-webpack/package.json
```diff
@@ -24,7 +24,7 @@
     "api-extractor": "yarn run -T api-extractor -c $INIT_CWD/api-extractor.json"
   },
   "peerDependencies": {
```
florkbr, line 26: @vojtechszocs Since this needs to be published ASAP - do you want to bump the versions in package.json to 2.0.0 for the effected packages and publish after this PR merges?

packages/lib-core/src/utils/random.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  return [...lowercase, ...uppercase, ...numbers].includes(char);
+};
+
+/**
+ * Return a string containing random characters generated via web `crypto` API.
+ *
+ * @see https://deepsource.io/blog/dont-use-math-random/
+ */
+export const getRandomString = (length = 16) => {
```
florkbr, line 16: @vojtechszocs couldn't we use a very popular open source package for this such as https://www.npmjs.com/package/uuid?

packages/eslint-plugin-internal/src/rules/all-bases.js
```diff
@@ -17,6 +17,22 @@ module.exports = {
   // Enforce a maximum number of classes per file
   'max-classes-per-file': 'off',
 
+  // Disallow reassigning function parameters
+  'no-param-reassign': [
```
florkbr, line 21: @vojtechszocs my personal opinion would be to leave these rules configured as eslint has them by default:
* `no-param-reassign`: I feel React is heavily leaning towards `props` being treated as immutable. I feel turning this off would cause several anti-patterns. https://eslint.org/docs/latest/rules/no-param-reassign
* `no-plusplus`: I think we should favor loops that avoid the need for us to manage our own iterator unless it is explicitly required. I see in some of the tests below we use the ++ to populate an empty array. We could achieve that using something like `[...Array(20)].map((item, i) => { return { foo: i }});`. https://eslint.org/docs/latest/rules/no-plusplus

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -93,63 +119,77 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
...
   constructor(options: DynamicRemotePluginOptions = {}) {
     const adaptedOptions: Required<DynamicRemotePluginOptions> = {
-      pluginMetadata: options.pluginMetadata ?? 'plugin.json',
-      extensions: options.extensions ?? 'extensions.json',
+      pluginMetadata: options.pluginMetadata ?? 'plugin-metadata.json',
+      extensions: options.extensions ?? 'plugin-extensions.json',
       sharedModules: options.sharedModules ?? {},
+      moduleFederationLibraryType: options.moduleFederationLibraryType ?? 'jsonp',
       entryCallbackSettings: options.entryCallbackSettings ?? {},
+      entryScriptFilename: options.entryScriptFilename ?? 'plugin-entry.js',
```
florkbr, line 137: @vojtechszocs NIT - we may want to consider putting the default options in an object and then spreading the provided options over it to make this a bit easier to read. Not blocking.

packages/lib-core/src/runtime/PluginLoader.ts
```diff
@@ -0,0 +1,553 @@
+import type { AnyObject } from '@monorepo/common';
```
florkbr, line 1: @vojtechszocs Github couldn't diff these files for some reason (this moved from `store/PluginLoader.ts` to here). I diffed the two locally:

```diff
2c2
< import { consoleLogger } from '@monorepo/common';
---
> import { consoleLogger, ErrorWithCause } from '@monorepo/common';
5c5
< import { PLUGIN_MANIFEST, REMOTE_ENTRY_SCRIPT, REMOTE_ENTRY_CALLBACK } from '../constants';
---
> import { DEFAULT_PLUGIN_MANIFEST, DEFAULT_REMOTE_ENTRY_CALLBACK } from '../constants';
9a10,12
> import { settleAllPromises } from '../utils/promise';
> import { getRandomString } from '../utils/random';
> import { injectScriptElement, getScriptElement } from '../utils/scripts';
12a16,21
> declare global {
>   interface Window {
>     [name: string]: PluginEntryCallback;
>   }
> }
> 
14d22
<   entryCallbackFired: boolean;
16a25,26
>   entryCallbackFired?: boolean;
>   entryCallbackModule?: PluginEntryModule;
45,47c55,79
<   /** Control which plugins can be loaded. */
<   canLoadPlugin: (pluginName: string) => boolean;
<   /** Name of the global callback function used by plugin entry scripts. */
---
>   /**
>    * Control which plugins can be loaded.
>    *
>    * The `reload` argument indicates whether an already loaded plugin is
>    * to be reloaded, assuming there was a change in the plugin manifest.
>    *
>    * By default, all plugins are allowed to be loaded.
>    */
>   canLoadPlugin: (manifest: PluginManifest, reload: boolean) => boolean;
> 
>   /**
>    * Control whether the given plugin script can be reloaded when attempting
>    * to reload the associated plugin.
>    *
>    * By default, all plugin scripts are allowed to be reloaded.
>    */
>   canReloadScript: (manifest: PluginManifest, scriptName: string) => boolean;
> 
>   /**
>    * Name of the global callback function used by plugin entry scripts.
>    *
>    * Applies only to plugins using the `callback` registration method.
>    *
>    * Default value: `__load_plugin_entry__`.
>    */
49c81,90
<   /** Custom resource fetch implementation. */
---
> 
>   /**
>    * Custom resource fetch implementation.
>    *
>    * The custom implementation may specify any host application or environment
>    * specific request headers that are necessary to fetch plugin resources over
>    * the network.
>    *
>    * By default, a basic {@link fetch} API based implementation is used.
>    */
51c92,106
<   /** Fixed resolutions used when processing plugin dependencies. */
---
> 
>   /**
>    * Fixed resolutions used when processing plugin dependencies.
>    *
>    * There are two kinds of dependencies:
>    * - Regular dependencies, which refer to actual plugins that must be loaded with
>    *   the right version prior to loading the plugin that depends on them.
>    * - Environment dependencies, which refer to any environment specific constants
>    *   such as the host application's version, runtime platform version, etc.
>    *
>    * This option should be used to satisfy any environment specific dependencies
>    * supported by the host application.
>    *
>    * Default value: empty object.
>    */
53c108,118
<   /** Shared scope object for initializing `PluginEntryModule` containers. */
---
> 
>   /**
>    * Shared scope object for initializing `PluginEntryModule` containers.
>    *
>    * Host applications built with webpack should use dedicated webpack specific APIs
>    * such as `__webpack_init_sharing__` and `__webpack_share_scopes__`.
>    *
>    * Default value: empty object.
>    *
>    * @see https://webpack.js.org/concepts/module-federation/#dynamic-remote-containers
>    */
55,56c120,138
<   /** Post-process the plugin manifest. Can be used as a custom validation hook. */
<   postProcessManifest: (manifest: PluginManifest) => Promise<PluginManifest>;
---
> 
>   /**
>    * Post-process the plugin manifest when fetched over the network.
>    *
>    * By default, no post-processing is performed on the manifest.
>    */
>   postProcessManifest: (manifest: PluginManifest) => PluginManifest;
> 
>   /**
>    * Provide access to the plugin's entry module.
>    *
>    * Applies only to plugins using the `custom` registration method.
>    *
>    * For example, if a plugin was built with `var` library type (i.e. its entry module
>    * is assigned to a variable), you can access the entry module via `window[pluginName]`.
>    *
>    * By default, this function does nothing.
>    */
>   getPluginEntryModule: (manifest: PluginManifest) => PluginEntryModule | void;
74c156
<       entryCallbackName: options.entryCallbackName ?? REMOTE_ENTRY_CALLBACK,
---
>       canReloadScript: options.canReloadScript ?? (() => true),
78c160,162
<       postProcessManifest: options.postProcessManifest ?? (async (manifest) => manifest),
---
>       postProcessManifest: options.postProcessManifest ?? _.identity,
>       entryCallbackName: options.entryCallbackName ?? DEFAULT_REMOTE_ENTRY_CALLBACK,
>       getPluginEntryModule: options.getPluginEntryModule ?? _.noop,
108c192
<    * Start loading a plugin from `baseURL`.
---
>    * Load a plugin from the given URL.
110,111c194,195
<    * This involves the following steps:
<    * - get plugin manifest
---
>    * This involves the following asynchronous operations:
>    * - load plugin manifest (unless provided as an object)
113c197
<    * - load plugin entry script
---
>    * - load plugin scripts
115,116c199,201
<    * The last step delegates control to the plugin entry script, which is supposed to call
<    * the global callback function created via {@link registerPluginEntryCallback} method.
---
>    * Plugins using the `callback` registration method are expected to call the global entry
>    * callback function created via {@link registerPluginEntryCallback} method, passing two
>    * arguments: plugin name and the entry module.
117a203,206
>    * For plugins using the `custom` registration method, the `getPluginEntryModule` function
>    * is expected to return the entry module of the given plugin. If not implemented properly,
>    * plugins using the `custom` registration method will fail to load.
>    *
120c209,212
<   async loadPlugin(baseURL: string) {
---
>   async loadPlugin(
>     baseURL: string,
>     manifestNameOrObject: string | PluginManifest = DEFAULT_PLUGIN_MANIFEST,
>   ) {
124c216,219
<       manifest = await this.getPluginManifest(baseURL);
---
>       manifest =
>         typeof manifestNameOrObject === 'string'
>           ? await this.loadPluginManifest(resolveURL(baseURL, manifestNameOrObject))
>           : manifestNameOrObject;
128c223
<         errorMessage: `Failed to get a valid plugin manifest from ${baseURL}`,
---
>         errorMessage: 'Failed to load plugin manifest',
133a229,240
>     try {
>       manifest = pluginManifestSchema.strict(true).validateSync(manifest, { abortEarly: false });
>     } catch (e) {
>       this.invokeListeners({
>         success: false,
>         pluginName: typeof manifest?.name === 'string' ? manifest.name : undefined,
>         errorMessage: 'Failed to validate plugin manifest',
>         errorCause: e,
>       });
>       return;
>     }
> 
134a242
>     const reload = this.plugins.has(pluginName);
137c245
<       consoleLogger.warn(`Attempt to reload plugin ${pluginName} while being loaded`);
---
>       consoleLogger.warn(`Attempt to reload plugin ${pluginName} which has not finished loading`);
141,142c249,254
<     if (this.plugins.get(pluginName)?.status === 'loaded') {
<       consoleLogger.warn(`Attempt to reload plugin ${pluginName} after being loaded`);
---
>     if (
>       reload &&
>       this.plugins.get(pluginName)?.status === 'loaded' &&
>       _.isEqual(manifest.compilationHash, this.plugins.get(pluginName)?.manifest.compilationHash)
>     ) {
>       consoleLogger.warn(`Attempt to reload plugin ${pluginName} with same compilation hash`);
146,150c258
<     const data: PluginLoadData = {
<       entryCallbackFired: false,
<       status: 'pending',
<       manifest,
<     };
---
>     const data: PluginLoadData = { status: 'pending', manifest };
151a260,263
>     if (manifest.registrationMethod === 'callback') {
>       data.entryCallbackFired = false;
>     }
> 
154c266
<     if (!this.options.canLoadPlugin(pluginName)) {
---
>     if (!this.options.canLoadPlugin(manifest, reload)) {
159c271
<         errorMessage: `Plugin ${pluginName} is not allowed to be loaded`,
---
>         errorMessage: `Plugin ${pluginName} is not allowed to be ${reload ? 'reloaded' : 'loaded'}`,
178c290
<       this.loadPluginEntryScript(baseURL, manifest);
---
>       await this.loadPluginScripts(baseURL, manifest, data);
184c296
<         errorMessage: `Failed to start loading entry script of plugin ${pluginName}`,
---
>         errorMessage: `Failed to load scripts of plugin ${pluginName}`,
191c303
<    * Fetch the plugin manifest from `baseURL` and validate it.
---
>    * Load plugin manifest from the given URL.
193,194c305,307
<   async getPluginManifest(baseURL: string) {
<     const manifestURL = resolveURL(baseURL, PLUGIN_MANIFEST);
---
>   private async loadPluginManifest(manifestURL: string) {
>     const response = await this.options.fetchImpl(manifestURL, { cache: 'no-cache' });
>     const responseText = await response.text();
196c309,310
<     consoleLogger.info(`Loading plugin manifest from ${manifestURL}`);
---
>     return this.options.postProcessManifest(JSON.parse(responseText));
>   }
198,199c312,316
<     const response = await this.options.fetchImpl(manifestURL);
<     const responseText = await response.text();
---
>   /**
>    * Load plugin scripts from the given URL.
>    */
>   private async loadPluginScripts(baseURL: string, manifest: PluginManifest, data: PluginLoadData) {
>     const pluginName = manifest.name;
201,203c318,321
<     let manifest: PluginManifest = await pluginManifestSchema
<       .strict(true)
<       .validate(JSON.parse(responseText));
---
>     const [, rejectedReasons] = await settleAllPromises(
>       manifest.loadScripts.map((scriptName) => {
>         const scriptID = `${pluginName}/${scriptName}`;
>         const scriptElement = getScriptElement(scriptID);
205c323,325
<     manifest = await this.options.postProcessManifest(manifest);
---
>         if (scriptElement && !this.options.canReloadScript(manifest, scriptName)) {
>           return Promise.resolve();
>         }
207c327,382
<     return manifest;
---
>         if (scriptElement) {
>           scriptElement.remove();
>         }
> 
>         return injectScriptElement(
>           resolveURL(baseURL, scriptName, (url) => {
>             url.searchParams.set('cacheBuster', getRandomString());
>             return url;
>           }),
>           scriptID,
>         );
>       }),
>     );
> 
>     if (rejectedReasons.length > 0) {
>       throw new ErrorWithCause(
>         `Detected ${rejectedReasons.length} errors while loading plugin scripts`,
>         rejectedReasons,
>       );
>     }
> 
>     if (manifest.registrationMethod === 'callback' && !data.entryCallbackFired) {
>       data.status = 'failed';
>       this.invokeListeners({
>         success: false,
>         pluginName,
>         errorMessage: `Scripts of plugin ${pluginName} loaded without entry callback`,
>       });
>       return;
>     }
> 
>     if (manifest.registrationMethod === 'callback' && !data.entryCallbackModule) {
>       data.status = 'failed';
>       this.invokeListeners({
>         success: false,
>         pluginName,
>         errorMessage: `Entry callback for plugin ${pluginName} called without entry module`,
>       });
>       return;
>     }
> 
>     const entryModule =
>       manifest.registrationMethod === 'callback'
>         ? data.entryCallbackModule
>         : this.options.getPluginEntryModule(manifest);
> 
>     if (entryModule) {
>       await this.processPlugin(data, entryModule);
>     } else {
>       data.status = 'failed';
>       this.invokeListeners({
>         success: false,
>         pluginName,
>         errorMessage: `Failed to retrieve entry module of plugin ${pluginName}`,
>       });
>     }
210c385,421
<   private getDependencyResolutions() {
---
>   private processPlugin = async (data: PluginLoadData, entryModule: PluginEntryModule) => {
>     const { manifest } = data;
>     const pluginName = manifest.name;
> 
>     if (typeof entryModule.init !== 'function' || typeof entryModule.get !== 'function') {
>       data.status = 'failed';
>       this.invokeListeners({
>         success: false,
>         pluginName,
>         errorMessage: `Entry module of plugin ${pluginName} does not meet expected contract`,
>       });
>       return;
>     }
> 
>     try {
>       await Promise.resolve(entryModule.init(this.options.sharedScope));
>     } catch (e) {
>       data.status = 'failed';
>       this.invokeListeners({
>         success: false,
>         pluginName,
>         errorMessage: `Failed to initialize shared modules of plugin ${pluginName}`,
>         errorCause: e,
>       });
>       return;
>     }
> 
>     data.status = 'loaded';
>     this.invokeListeners({
>       success: true,
>       pluginName,
>       manifest,
>       entryModule,
>     });
>   };
> 
>   private getCurrentDependencyResolutions() {
235c446
<   async resolvePluginDependencies(manifest: PluginManifest) {
---
>   private async resolvePluginDependencies(manifest: PluginManifest) {
257c468
<         const resolutions = this.getDependencyResolutions();
---
>         const resolutions = this.getCurrentDependencyResolutions();
286d496
< 
308,355d517
<   /**
<    * Start loading the plugin entry script from `baseURL`.
<    */
<   loadPluginEntryScript(
<     baseURL: string,
<     manifest: PluginManifest,
<     getDocument: () => typeof document = _.constant(document),
<   ) {
<     const pluginName = manifest.name;
<     const data = this.plugins.get(pluginName);
< 
<     if (data?.status !== 'pending') {
<       consoleLogger.warn(`Attempt to load entry script of non-pending plugin ${pluginName}`);
<       return;
<     }
< 
<     const script = getDocument().createElement('script');
<     const scriptURL = resolveURL(baseURL, REMOTE_ENTRY_SCRIPT);
< 
<     script.src = scriptURL;
<     script.async = true;
< 
<     script.onload = () => {
<       if (!data.entryCallbackFired) {
<         data.status = 'failed';
<         this.invokeListeners({
<           success: false,
<           pluginName,
<           errorMessage: `Entry script of plugin ${pluginName} loaded without callback`,
<         });
<       }
<     };
< 
<     script.onerror = (event) => {
<       data.status = 'failed';
<       this.invokeListeners({
<         success: false,
<         pluginName,
<         errorMessage: `Failed to load entry script of plugin ${pluginName}`,
<         errorCause: event,
<       });
<     };
< 
<     consoleLogger.info(`Loading entry script of plugin ${pluginName} from ${scriptURL}`);
< 
<     getDocument().head.appendChild(script);
<   }
< 
359c521
<         consoleLogger.warn(`Received entry callback for unknown plugin ${pluginName}`);
---
>         consoleLogger.warn('Received entry callback for unknown plugin');
372,392c534
< 
<       Promise.resolve(entryModule.init(this.options.sharedScope))
<         // eslint-disable-next-line promise/always-return -- entryModule.init() returns Promise<void>
<         .then(() => {
<           data.status = 'loaded';
<           this.invokeListeners({
<             success: true,
<             pluginName,
<             manifest: data.manifest,
<             entryModule,
<           });
<         })
<         .catch((e) => {
<           data.status = 'failed';
<           this.invokeListeners({
<             success: false,
<             pluginName,
<             errorMessage: `Failed to initialize shared modules of plugin ${pluginName}`,
<             errorCause: e,
<           });
<         });
---
>       data.entryCallbackModule = entryModule;
397a540,541
>    *
>    * This must be called in order to load plugins using the `callback` registration method.
399,400c543
<   registerPluginEntryCallback(getWindow: () => typeof window = _.constant(window)) {
<     const windowGlobal = getWindow() as unknown as AnyObject;
---
>   registerPluginEntryCallback() {
403,404c546,547
<     if (typeof windowGlobal[callbackName] === 'function') {
<       consoleLogger.warn(`Plugin callback ${callbackName} is already registered`);
---
>     if (typeof window[callbackName] === 'function') {
>       consoleLogger.warn(`Plugin entry callback ${callbackName} is already registered`);
408c551
<     windowGlobal[callbackName] = this.createPluginEntryCallback();
---
>     window[callbackName] = this.createPluginEntryCallback();
```

packages/lib-webpack/src/webpack/ValidateCompilationPlugin.ts
```diff
@@ -0,0 +1,29 @@
...
+import { findPluginChunks } from '../utils/plugin-chunks';
+
+export class ValidateCompilationPlugin implements WebpackPluginInstance {
+  constructor(private readonly containerName: string, private readonly jsonpLibraryType: boolean) {}
+
+  apply(compiler: Compiler) {
+    compiler.hooks.afterCompile.tap(ValidateCompilationPlugin.name, (compilation) => {
+      const { runtimeChunk } = findPluginChunks(this.containerName, compilation);
+
+      if (runtimeChunk && this.jsonpLibraryType) {
```
florkbr, line 11: @vojtechszocs it would still be nice to link some sort of Webpack issue or doc link here since in theory this is a webpack issue - but I know we may not be able to find one.

packages/lib-webpack/src/webpack/PatchEntryCallbackPlugin.ts
```diff
@@ -0,0 +1,41 @@
...
+        () => {
+          const { entryChunk } = findPluginChunks(this.containerName, compilation);
+
+          entryChunk.files.forEach((fileName) => {
+            compilation.updateAsset(fileName, (source) => {
+              const newSource = new sources.ReplaceSource(source);
+              const fromIndex = source.source().toString().indexOf(`${this.callbackName}(`);
+
+              if (fromIndex >= 0) {
+                newSource.insert(fromIndex + this.callbackName.length + 1, `'${this.pluginID}', `);
```
florkbr, line 27: @vojtechszocs could you explain what exactly we are doing here? I thought webpack would take care of the naming when using `callback` registration type, or the app would be responsible for this when using `custom`.

packages/lib-webpack/src/webpack/PatchEntryCallbackPlugin.ts
```diff
@@ -0,0 +1,41 @@
+import { WebpackPluginInstance, Compiler, Compilation, sources, WebpackError } from 'webpack';
```
florkbr, line 1: Github destroyed this diff too :(  Local diff:

```diff
1a2
> import { findPluginChunks } from '../utils/plugin-chunks';
3c4
< export class PatchContainerEntryPlugin implements WebpackPluginInstance {
---
> export class PatchEntryCallbackPlugin implements WebpackPluginInstance {
5c6
<     private readonly fileName: string,
---
>     private readonly containerName: string,
11c12
<     compiler.hooks.thisCompilation.tap(PatchContainerEntryPlugin.name, (compilation) => {
---
>     compiler.hooks.thisCompilation.tap(PatchEntryCallbackPlugin.name, (compilation) => {
14c15
<           name: PatchContainerEntryPlugin.name,
---
>           name: PatchEntryCallbackPlugin.name,
18c19,22
<           compilation.updateAsset(this.fileName, (source) => {
---
>           const { entryChunk } = findPluginChunks(this.containerName, compilation);
> 
>           entryChunk.files.forEach((fileName) => {
>             compilation.updateAsset(fileName, (source) => {
25,28c29,30
<               const error = new WebpackError(
<                 `Missing call to ${this.callbackName} in ${this.fileName}`,
<               );
<               error.file = this.fileName;
---
>                 const error = new WebpackError(`Missing call to ${this.callbackName}`);
>                 error.file = fileName;
32a35
>             });
```

packages/lib-webpack/src/webpack/PatchEntryCallbackPlugin.ts
florkbr: Ah - nevermind. I see what's going on with the new diff above. Sorry for the confusion

packages/lib-webpack/src/webpack/GenerateManifestPlugin.ts
```diff
@@ -12,10 +17,28 @@ export class GenerateManifestPlugin implements WebpackPluginInstance {
...
+
           compilation.emitAsset(
-            this.fileName,
-            new sources.RawSource(Buffer.from(JSON.stringify(this.manifest, null, 2))),
+            this.manifestFilename,
+            new sources.RawSource(Buffer.from(JSON.stringify(manifest, null, 2))),
           );
+
+          if (manifest.extensions.length === 0) {
+            const error = new WebpackError('Plugin manifest has no extensions');
```
florkbr, line 38: @vojtechszocs have we always asserted that a plugin must have 1 or more extensions? Why are we adding this stipulation now?

packages/lib-webpack/package.json
vojtechszocs: Let's do the package version bump for `lib-core` and `lib-webpack` in a separate PR and publish once it's merged.

packages/lib-core/src/utils/random.ts
vojtechszocs: While I consider the above `getRandomString` implementation to be relatively small and solid, I agree with your point about using the existing `uuid` package. I've updated the code in `PluginStore` to use the `uuid` v4 function.

packages/eslint-plugin-internal/src/rules/all-bases.js
vojtechszocs: Based on our discussion, I've reverted the custom configs for the above rules, as well as updating the existing `for` loops to use `index += 1` instead of `index++`.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
vojtechszocs: Agreed. One note to add here, the updated code should retain the type safety, i.e. there must be a default value for each declared option within the "defaults" object. We can do this as a follow-up.

packages/lib-core/src/runtime/PluginLoader.ts
vojtechszocs: Sorry, this was my fault, I didn't do `git mv` on this file, so it shows up as removed and added like it was new.

packages/lib-core/src/runtime/PluginLoader.ts
```diff
@@ -0,0 +1,553 @@
...
+      this.invokeListeners({
+        success: false,
+        errorMessage: 'Failed to load plugin manifest',
+        errorCause: e,
+      });
+      return;
+    }
+
+    try {
+      manifest = pluginManifestSchema.strict(true).validateSync(manifest, { abortEarly: false });
```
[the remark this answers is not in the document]
vojtechszocs: We don't validate the generated plugin manifest, because we're the ones generating it here :smiley: 

We originally validated only some `DynamicRemotePlugin` constructor options (`pluginMetadata` and `extensions`). Current code validates the whole "options" object passed to `DynamicRemotePlugin` constructor.

packages/lib-webpack/src/webpack/PatchEntryCallbackPlugin.ts
vojtechszocs: This is another file where I messed up the `git mv` sorry about that.

packages/lib-webpack/src/webpack/GenerateManifestPlugin.ts
vojtechszocs: This plugin SDK is built on the assumption that individual plugins will utilize extensions to extend host applications.

We didn't have this check before because ^^ assumption was implicit, now we're just making it more explicit. If a plugin doesn't provide any extensions (checked during plugin manifest generation), we emit a warning to the webpack compilation.

There may be existing plugins (like the ones for ConsoleDot) which communicate with their host application only via exposed modules. In order to support such plugins, we added new `PluginStore.getExposedModule` API which is basically an equivalent to [Scalprum `getCachedModule` function](https://github.com/scalprum/scaffloding/blob/main/packages/core/src/index.ts).

florkbr:
Re-reviewed with the latest commit and things look good to me. @vojtechszocs to publish this as version 2.0.0 as there are major breaking changes in this PR.

PR #184, opened 2022-11-28
RHCLOUD-23014: Use API Extractor on distributable packages

vojtechszocs:
Fixes [RHCLOUD-23014](https://issues.redhat.com/browse/RHCLOUD-23014)

This PR adds [API Extractor](https://api-extractor.com/) and [API Documenter](https://api-extractor.com/pages/commands/api-documenter_markdown/) tools, with their workflow applied on all distributable `lib` packages.

## API Extractor workflow
Executed as part of the package `build` script, API Extractor does the following:
- generate `index.d.ts` bundle, e.g. `packages/lib-core/dist/index.d.ts`
- generate Markdown API report, e.g. `packages/lib-core/dist/api/lib-core.api.md`
- generate JSON API doc model, e.g. `packages/lib-core/dist/api/lib-core.api.json`

_Note_ :memo: [`rollup-plugin-dts`](https://github.com/Swatinem/rollup-plugin-dts) is dropped in favor of the `d.ts` bundling feature of API Extractor. This means that Rollup is only used for generating the `index.js` bundle.

### API reports overview

> API reports are meant to improve the visibility of public APIs exposed by the given package and track public API changes through code review process. These reports are aimed at plugin SDK maintainers to better understand the impact of changes on public APIs.

API Extractor compares the currently generated API report file with one currently committed at `reports` directory.

If the files don't match, API Extractor fails the build and instructs the developer to copy the currently generated file to `reports` directory. This way, changes in API reports become part of the code review.

_Note_ :memo: `tsdoc-metadata.json` is not generated since we don't need it, see [here](https://github.com/Azure/azure-sdk-for-js/issues/8662#issuecomment-765088241) and [here](https://github.com/microsoft/rushstack/pull/1628#issuecomment-553665782) for details.

### API docs overview

> API docs are meant to document public APIs of the given package (various objects, functions, React components, etc.) from plugin SDK consumer perspective. In other words, these docs are aimed at developers of dynamic plugins as well as developers of host applications that support such plugins.

This PR adds `generate-api-docs.sh` script that copies all the currently generated API doc model files to `docs/generated` directory and then invokes API Documenter to generate Markdown API docs at `docs/generated/api`.

_Note_ :memo: `docs/generated` directory is git-ignored which means that the `docs/generated/api` content isn't currently tracked in the git repo.

As a follow-up, we should create a new monorepo package with all developer oriented docs (API docs, READMEs etc.) collected and processed into static web pages ready for deployment on services like [Surge.sh](https://surge.sh/).

## Changes due to API Extractor package analysis

- internal methods in `PluginLoader` and `PluginStore` are marked as `private`
  - this causes such methods to be excluded from the generated `index.d.ts` bundle
  - as for `private` method testability, I generally agree with points described [here](https://softwareengineering.stackexchange.com/a/380296) - code should be tested in a way similar to how the client would use it, which is through the public interface
- new API exports (***lots*** of them) to ensure zero [`ae-forgotten-export`](https://api-extractor.com/pages/messages/ae-forgotten-export/) warnings
- added [`@packageDocumentation`](https://api-extractor.com/pages/tsdoc/tag_packagedocumentation/) to index files of all `lib` packages
- fixed malformed or malformatted TSDoc comments

## Other changes and improvements

- new [runtime ESLint rule](https://eslint.org/docs/latest/developer-guide/working-with-rules#runtime-rules) `lib-restricted-external-imports` which enforces imported external modules to be declared in `dependencies` or `peerDependencies` within the closest parent `package.json`
  - this effectively ensures that Rollup generated `index.js` bundle doesn't include any external module code
- [`import/no-extraneous-dependencies`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md) replaced by `lib-restricted-external-imports`
- removed `replaceCode` custom Rollup plugin for Lodash-specific code manipulation
  - not needed since `lib-restricted-external-imports` guards the import vs. manifest dependency consistency
- [`downlevel-dts`](https://github.com/sandersn/downlevel-dts) dropped for now since we don't need it (yet)
  - can be re-added once we need TypeScript 3.x compatible typings for our `lib` packages
- `dist/build-meta.json` renamed to `dist/build-metadata.json`


reports/lib-utils.api.md
```diff
@@ -0,0 +1,837 @@
...
+// @public (undocumented)
+export type FilterValue = Partial<{
+    selected: string[];
+    all: string[];
+}>;
+
+// @public (undocumented)
+export type GenericHandler<T = unknown> = (data: T) => void;
+
+// @public (undocumented)
```
florkbr, line 221: @vojtechszocs I'm assuming we need to take todos to document these undocumented types?

package.json
```diff
@@ -60,7 +62,6 @@
     "@typescript-eslint/parser": "^5.3.1",
     "babel-loader": "^8.2.5",
     "cypress": "^9.7.0",
-    "downlevel-dts": "^0.10.0",
```
florkbr, line 63: Just for sanity/clarity - we will need to add this back in a later PR for the 4.13 OCP console dynamic plugins epic: https://issues.redhat.com/browse/CONSOLE-3304

packages/eslint-plugin-internal/src/runtime-rules/lib-restricted-external-imports.js
```diff
@@ -0,0 +1,101 @@
...
+const findUp = require('find-up');
+const minimatch = require('minimatch');
+
+/** @typedef {{ includeFiles: string, excludeFiles: string, excludeModules: string[] }} RuleSetting */
+
+/**
+ * @param {import('eslint').Rule.RuleContext} context
+ * @param {import('estree').ImportDeclaration} node
+ */
+const checkImport = (context, node) => {
```
florkbr, line 11: @vojtechszocs I'm a bit concerned we are overriding the default eslint behavior here. Can you give a brief summary of why this override of the default eslint `import` behavior is required? Wondering if it's possible for us to fix the root issue.

packages/common/rollup-configs.js
```diff
@@ -113,9 +84,8 @@ export const tsLibConfig = (pkg, inputFile, format = 'esm') => {
         include: ['src/**/*', '../common/src/**/*'],
         noEmitOnError: true,
       }),
-      ...(format === 'cjs' ? [replaceLodashEsRequire('index.js')] : []),
       writeJSONFile({
-        fileName: 'build-meta.json',
+        fileName: 'build-metadata.json',
```
florkbr, line 88: @vojtechszocs can we move the filename to the top of the file in a constant so it's easier to find and change in the future?

package.json
florkbr: Also @vojtechszocs we may want to consider splitting up unrelated changes in the future. It will reduce the diff (already very large) and make it faster for approval.

packages/lib-extensions/src/index.ts
```diff
@@ -1 +1,12 @@
+/**
+ * Extension types that can be interpreted at runtime.
+ *
+ * @remarks
+ * This package provides extension types and corresponding type guard functions
+ * used to interpret these extensions at runtime.
```
florkbr, line 6: NIT @vojtechszocs maybe instead of `these` extensions we say `SDK` extensions?

packages/lib-utils/src/components/details-page-header/DetailsPageHeader.tsx
```diff
@@ -7,7 +7,7 @@ import {
   Label,
   DropdownPosition,
 } from '@patternfly/react-core';
-import * as _ from 'lodash';
+import * as _ from 'lodash-es';
```
florkbr, line 10: @vojtechszocs I see in some other locations we removed the `-es` and here we are adding it. Can we come up with a consistent losdah import across all of our packages?

florkbr (commented):
Testing these changes still in hac-dev and hac-core now.

florkbr:
@vojtechszocs I actually ran into some issues testing this in hac-core as https://github.com/openshift/hac-core/pull/108/f has not been merged yet 🤦 

I'll update that PR tomorrow if you're willing to review it.

test.sh
```diff
@@ -21,5 +21,11 @@ yarn lint
 # Run unit tests
 yarn test
 
+# Ensure that Git repo is still clean
+if [[ -n "$(git status --porcelain)" ]]; then
+  echo "Git repository is not clean, make sure that all changes are committed"
```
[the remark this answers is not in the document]
vojtechszocs: Generally speaking, if a git repo is clean before building the project, it should remain clean after the build as well. This was the intended purpose of the check above. Any artifacts generated during the build should be either committed or git-ignored.

Originally, I wanted to commit the generated API docs to the git repo, but then decided to git-ignore them for now, given the above mentioned follow-up that puts them to actual use.

> I know several folks run this script locally and this will be very annoying with uncommitted changes during development.

Stuff which is user specific, like TODO files for example, can be git-ignored, e.g. I have this in my `~/.gitignore`
```
**/vszocs-TODO*.txt
```
which allows me to put e.g. `vszocs-TODO-items.txt` file in any git repo I work on.

That said, I guess you're right that it may cause troubles for some folks, so I'll remove it.

packages/eslint-plugin-internal/src/runtime-rules/lib-restricted-external-imports.js
```diff
@@ -0,0 +1,101 @@
+const path = require('path');
+const findUp = require('find-up');
+const minimatch = require('minimatch');
+
+/** @typedef {{ includeFiles: string, excludeFiles: string, excludeModules: string[] }} RuleSetting */
+
+/**
+ * @param {import('eslint').Rule.RuleContext} context
+ * @param {import('estree').ImportDeclaration} node
```
vojtechszocs, line 9: @florkbr Forgot to mention, this rule doesn't check TS specific `export =` and `import = require()` syntax as documented [here](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require), since the main purpose of that syntax is interop with legacy CommonJS or AMD workflow.

packages/common/rollup-configs.js
vojtechszocs: Chatted with @florkbr and since we're not planning to change this string in future, we can keep it here.

generate-api-docs.sh
```diff
@@ -6,7 +6,8 @@ yarn build-libs
 
 mkdir -p docs/generated && rm -rf docs/generated/*
 
-shopt -s globstar
-cp packages/lib-*/dist/api/lib-*.api.json docs/generated
+for dir in packages/lib-*; do
+  cp $dir/dist/api/$(basename $dir).api.json docs/generated
```
florkbr, line 10: 👍

florkbr:
Latest commit looks good to me and tested in hac-core with hac-dev:
Screenshot 2022-12-07 at 3 00 19 PM

PR #186, opened 2022-11-29
Add option to pass static API models

karelhala:
### Description

API Discovery is firing too many requests when trying to map entire API. This PR adds option to pass already known api models. The consuming app will pass an object where keys will be edpoint to keep static and value will be the response itself.

### DIFF
--- a/packages/lib-utils/src/app/AppInitSDK.tsx
+++ b/packages/lib-utils/src/app/AppInitSDK.tsx
@@ -12,7 +12,7 @@ export type AppInitSDKProps = {
   configurations: {
     apiDiscovery?: InitAPIDiscovery;
     apiPriorityList?: string[];
-    staticApiModels?: Record<string, APIResourceList>;
+    staticAPIModels?: Record<string, APIResourceList>;
     appFetch: UtilsConfig['appFetch'];
     pluginStore: PluginStore;
     wsAppSettings: UtilsConfig['wsAppSettings'];
@@ -45,7 +45,7 @@ const AppInitSDK: React.FC<AppInitSDKProps> = ({ children, configurations }) => 
     wsAppSettings,
     apiDiscovery = initAPIDiscovery,
     apiPriorityList,
-    staticApiModels,
+    staticAPIModels,
   } = configurations;
 
   React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -5,14 +5,15 @@ import * as React from 'react';
...
-import type { InitAPIDiscovery } from '../types/api-discovery';
+import type { APIResourceList, InitAPIDiscovery } from '../types/api-discovery';
 import { initAPIDiscovery } from './api-discovery';
 import { useReduxStore } from './redux';
 
 type AppInitSDKProps = {
   configurations: {
     apiDiscovery?: InitAPIDiscovery;
     apiPriorityList?: string[];
+    staticApiModels?: Record<string, APIResourceList>;
```
florkbr, line 16: Nit: should we capitalize `API`? Not blocking.

packages/lib-utils/src/app/api-discovery/index.ts
```diff
@@ -67,11 +67,11 @@ type APIResourceData = {
...
 
 const batchResourcesRequest = (
-  batch: string[],
+  batch: (string | APIResourceList)[],
   groupVersionMap: AnyObject,
 ): Promise<DiscoveryResources>[] => {
-  return batch.map<Promise<DiscoveryResources>>(async (p: string) => {
-    const resourceList = await commonFetchJSON<APIResourceList>(p);
+  return batch.map<Promise<DiscoveryResources>>(async (p: string | APIResourceList) => {
+    const resourceList = typeof p === 'string' ? await commonFetchJSON<APIResourceList>(p) : p;
```
florkbr, line 74: Ideally I'd like a cleaner method of determining if we should fetch or not, but for now this should be fine (rather then fetching if the parameter is a string). When I was looking at this code I was wondering if we could split the actual fetch out to another function and have the processing be in it's own function that we could invoke directly with the static resource list.

Just a thought and not blocking.

florkbr (commented):
Overall looks good to me. @karelhala is there a good way to test this in hac-core?

packages/lib-utils/src/app/AppInitSDK.tsx
karelhala: Good point!

packages/lib-utils/src/app/api-discovery/index.ts
karelhala: Yeah, I wanted to split the logic as well, but it was too complex to work with. Let's roll with this approach and we can refactor in current sprint as it's a tech debt sprint.

karelhala:
@florkbr yeah! I'll open a PR there for testing.

florkbr:
@vojtechszocs I think something may be wrong with the API extractor return code in CI... I think it's failing the build here

PR #187, opened 2022-12-23
Add option to allow APIs app is interested in

karelhala:
### Description

Another API discovery improvement which allows application to list api it is interested in. API discovery will watch only for these API endpoints and would further limit the number of requests.

### Caveats

Once this is merged and hac-core app starts to send in the list apps might not be able to watch for some resources, unless they specifically list them. We should communicate this change to apps and prepare them for this change (adding the apiAllowed to their config) before we roll this feature fully.

### Example

Usage of this new feature can be seen at https://github.com/openshift/hac-core/pull/119

### DIFF
--- a/packages/lib-core/src/types/plugin.ts
+++ b/packages/lib-core/src/types/plugin.ts
@@ -4,6 +4,7 @@ import type { PluginEntryModule } from './runtime';
 export type PluginRuntimeMetadata = {
   name: string;
   version: string;
+  apiWhitelist?: Readonly<(string | undefined)[] | undefined>;
   dependencies?: Record<string, string>;
 };
 
--- a/packages/lib-core/src/types/plugin.ts
+++ b/packages/lib-core/src/types/plugin.ts
@@ -4,7 +4,7 @@ import type { PluginEntryModule } from './runtime';
 export type PluginRuntimeMetadata = {
   name: string;
   version: string;
-  apiWhitelist?: Readonly<(string | undefined)[] | undefined>;
+  apiAllowed?: Readonly<(string | undefined)[] | undefined>;
   dependencies?: Record<string, string>;
 };
 
--- a/packages/lib-core/src/yup-schemas.ts
+++ b/packages/lib-core/src/yup-schemas.ts
@@ -88,14 +88,18 @@ export const extensionArraySchema = yup.array().of(extensionSchema).required();
 /**
  * Schema for `PluginRuntimeMetadata` objects.
  */
-export const pluginRuntimeMetadataSchema = yup.object().required().shape({
-  name: pluginNameSchema,
-  version: semverStringSchema,
-  // TODO(vojtech): Yup lacks native support for map-like structures with arbitrary keys
-  // TODO(vojtech): suppress false positive https://github.com/jsx-eslint/eslint-plugin-react/pull/3326
-  // eslint-disable-next-line react/forbid-prop-types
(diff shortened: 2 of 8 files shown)


jerolimov (changes requested):
Hey @karelhala, we should try to avoid a nomenclature like `whitelist` as people might feel uncomfortable with this and we try to have an inclusive culture. Maybe you can use `apiAllowed` instead? I don't check the exact details here.

You might know this already and just overseen this here, but if you want to read some more about this you find some info here:

1. https://source.redhat.com/groups/public/word-nerds/red_hat_word_nerds_wiki/corporate_style_guide
2. https://source.redhat.com/groups/public/ospo/open_source_program_office_wiki/resources_for_avoiding_problematic_language
3. https://lethargy.org/~jesus/writes/a-guide-to-nomenclature-selection/

jerolimov:
Thanks! For the general review I would expect that someone else will do it, but thanks for the naming change! :+1:

jerolimov:
Thanks! For the general review I would expect that someone else will do it, but thanks for the naming change! :+1: 

/uncc

florkbr:
@karelhala this PR build isfailing due to missing generated API report docs for lib-core and lib-utils. When you run a local build you should see a warning with instructions to copy the modified API doc into version control under the `reports` dir.

I've create a commit with those changes if you rather cherry pick it: https://github.com/openshift/dynamic-plugin-sdk/commit/1d79ec6207ad66b24a04900519c234076f1ce44a

karelhala:
@florkbr aaah, good point! I didn't read the warnings. Updated the PR with those changes.

florkbr:
@karelhala now prettier is angry :(

```
openshift/dynamic-plugin-sdk/packages/lib-utils/src/app/api-discovery/index.ts
  149:7  error  Insert `··`  prettier/prettier
```

Do we still need this PR with your PR to move api-discovery to hac-core?

karelhala:
@florkbr I didn't even tuched that line. I don't think so. I'm keeping it open just in case we decide to merge it in dynamic plugin SDK as well.

packages/lib-core/src/types/plugin.ts
```diff
@@ -4,6 +4,7 @@ import type { PluginEntryModule } from './runtime';
 export type PluginRuntimeMetadata = {
   name: string;
   version: string;
+  apiAllowed?: Readonly<(string | undefined)[] | undefined>;
```
vojtechszocs, line 7: @karelhala @florkbr 

Is this still up to date?

If so, I'd be very careful about adding new fields into the plugin manifest.

Can you please describe the purpose of this new field?

If our intent is to have individual plugins pass additional information to host applications like HAC, we should introduce some standard API to access such information.

For example, exposing the manifest via `PluginStore.getPluginInfo` method and corresponding React `usePluginInfo` hook.

vojtechszocs (changes requested):

packages/lib-core/src/types/plugin.ts
karelhala: @vojtechszocs sorry for such a late reply, we've had to rearrange things slightly. Good poin on the custom field to allow any config from plugin to host app. I'll opene a new PR with this change (will probably close this 187 PR)

karelhala:
Closing in favor of https://github.com/openshift/dynamic-plugin-sdk/pull/204

PR #190, opened 2023-01-25
Allow postProcessManifest on manifest object

Hyperkid123:
It is beneficial to allow the `postProcessManifest` function to run even if a full manifest was passed to the `loadPlugin` function. Sometimes you might need some exception on manifest files, even if they have the correct format.

One exception might be not including the runtime chunk of a module that has disruptive code (react-dom render for example) in that chunk.

cc @vojtechszocs @florkbr

### DIFF
--- a/packages/lib-core/src/runtime/PluginLoader.ts
+++ b/packages/lib-core/src/runtime/PluginLoader.ts
@@ -119,7 +119,7 @@ export type PluginLoaderOptions = Partial<{
   sharedScope: AnyObject;
 
   /**
-   * Post-process the plugin manifest when fetched over the network.
+   * Post-process the plugin manifest.
    *
    * By default, no post-processing is performed on the manifest.
    */
@@ -226,6 +226,8 @@ export class PluginLoader {
       return;
     }
 
+    manifest = this.options.postProcessManifest(manifest);
+
     try {
       manifest = pluginManifestSchema.strict(true).validateSync(manifest, { abortEarly: false });
     } catch (e) {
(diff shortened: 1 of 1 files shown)


florkbr:
I think this looks fine. I would like @vojtechszocs to have final approval

packages/lib-core/src/runtime/PluginLoader.ts
```diff
@@ -216,7 +216,7 @@ export class PluginLoader {
       manifest =
         typeof manifestNameOrObject === 'string'
           ? await this.loadPluginManifest(resolveURL(baseURL, manifestNameOrObject))
-          : manifestNameOrObject;
+          : this.options.postProcessManifest(manifestNameOrObject);
```
vojtechszocs, line 219: @Hyperkid123 @florkbr 

We already call `this.options.postProcessManifest` inside `this.loadPluginManifest` and I think it's better to just call it on one place.

I'd suggest to change the `loadPluginManifest` method like so:

```diff
   const response = await this.options.fetchImpl(manifestURL, { cache: 'no-cache' });
   const responseText = await response.text();
 
-  return this.options.postProcessManifest(JSON.parse(responseText));
+  return JSON.parse(responseText);
```

and then add the following line right before the validation try/catch statement:

```ts
manifest = this.options.postProcessManifest(manifest);
```

Also please update the doc for `postProcessManifest` option:

```diff
- * Post-process the plugin manifest when fetched over the network.
+ * Post-process the plugin manifest.
  *
  * By default, no post-processing is performed on the manifest.
```

vojtechszocs (changes requested):

packages/lib-core/src/runtime/PluginLoader.ts
Hyperkid123: @vojtechszocs done

PR #191, opened 2023-01-25
Expose public API functions for api discovery

karelhala:
### Description

Since apps can use custom api discovery we have to expose functions to dispatch actions to redux store. We can even later down the line change the data passed from application to check if they are sending correct data.

### DIFF
--- a/packages/lib-utils/src/app/api-discovery/index.ts
+++ b/packages/lib-utils/src/app/api-discovery/index.ts
@@ -5,6 +5,7 @@ import { plural } from 'pluralize';
 import type { Dispatch } from 'redux';
 import { kindToAbbr } from '../../k8s/k8s-resource';
 import type {
+  APIActions,
   APIResourceList,
   DiscoveryResources,
   InitAPIDiscovery,
@@ -17,6 +18,12 @@ import { cacheResources, getCachedResources } from './discovery-cache';
 
 const API_DISCOVERY_INIT_DELAY = 5_000;
 const API_DISCOVERY_REQUEST_BATCH_SIZE = 5;
+
+export const createAPIActions = (dispatch: Dispatch): APIActions => ({
+  setResourcesInFlight: (isInFlight) => dispatch(setResourcesInFlight(isInFlight)),
+  setBatchesInFlight: (isInFlight) => dispatch(setBatchesInFlight(isInFlight)),
+  receivedResources: (resource) => dispatch(receivedResources(resource)),
+});
 
 const pluralizeKind = (kind: string): string => {
   // Use startCase to separate words so the last can be pluralized but remove spaces so as not to humanize
--- a/packages/lib-utils/src/types/api-discovery.ts
+++ b/packages/lib-utils/src/types/api-discovery.ts
@@ -36,3 +36,9 @@ export type DiscoveryResources = {
     };
   };
 };
+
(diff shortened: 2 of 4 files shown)


karelhala:
@florkbr ping

packages/lib-utils/src/app/AppInitSDK.tsx
```diff
@@ -18,6 +20,12 @@ export type AppInitSDKProps = {
   };
 };
 
+export const createActions = (dispatch: Dispatch): ApiActions => ({
```
florkbr, line 23: @karelhala
1. I'm wondering if we should rename this function to something specific to API discovery. I'm not sure we would want to encourage exposing all of our actions. It appears all of the actions below are only utilized by API discovery:
  * https://github.com/openshift/dynamic-plugin-sdk/search?q=setResourcesInFlight
  * https://github.com/openshift/dynamic-plugin-sdk/search?q=setBatchesInFlight
  * https://github.com/openshift/dynamic-plugin-sdk/search?q=receivedResources
2. Should we move this export into the API discovery implementation versus `AppInitSDK`?

packages/lib-utils/src/app/api-discovery/index.ts
```diff
@@ -2,17 +2,15 @@ import type { AnyObject } from '@openshift/dynamic-plugin-sdk';
 import { consoleLogger } from '@openshift/dynamic-plugin-sdk';
 import * as _ from 'lodash-es';
 import { plural } from 'pluralize';
-import type { Dispatch } from 'redux';
 import { kindToAbbr } from '../../k8s/k8s-resource';
 import type {
+  ApiActions,
```
vojtechszocs, line 7: Nit: abbreviations should be uppercase in type names, e.g. `APIActions` instead of `ApiActions`.

vojtechszocs (changes requested):
Small nitpick otherwise LGTM.

Since this is lib-utils, deferring approval to @florkbr

packages/lib-utils/src/types/api-discovery.ts
```diff
@@ -1,11 +1,12 @@
...
-  store: Store<unknown, Action<AnyAction>>,
-  preferenceList?: string[],
-) => void;
+export type ApiActions = {
+  setResourcesInFlight: (isInFlight: boolean) => void;
+  setBatchesInFlight: (isInFlight: boolean) => void;
+  receivedResources: (resource: DiscoveryResources) => void;
+};
+
+export type InitAPIDiscovery = (apiActions: ApiActions, preferenceList?: string[]) => void;
```
florkbr, line 9: @karelhala do we really need to modify the signature of `InitAPIDiscovery`? I was thinking if an app chooses to use our baked-in API discovery they would just provide the `store`. If an app chooses not to use the built in API discovery we will have action creators exposed to pipe data into the SDK redux.

karelhala: Well we are calling the `apiDiscovery` which defaults to `initAPIDiscovery`. We can expose action creator and app would pass in the `store` to it and get these functions, which could probably be more straight forward approach. I was just thing of how to provide the proper store. But when we call the `apiDiscovery` we were passing the store so that could be used for that. Let me change it slightly.

florkbr (approved):

PR #195, opened 2023-01-27
Export common functions for lib utils version 1

karelhala:


florkbr:
I think we should favor this PR in lieu of adding back the exports to lib-core: https://github.com/openshift/dynamic-plugin-sdk/pull/196 (update lib-utils to use the monorepo common package)

florkbr:
#196 is merged. Closing this for now.

PR #196, opened 2023-01-27
Update lib utils to use monorepo common

florkbr:
@vojtechszocs @karelhala FYI. We ran into this issue trying to use `lib-utils` version 1.0.2 in HAC Core. It looks like we missed updated lib-utils to pull the common functions from `@monorepo/common` versus `lib-core`.


karelhala (approved):
/lgtm

florkbr:
Spoke with @vojtechszocs - we will need to revert this PR once HAC Core and HAC Dev are updated to `lib-core@2.0.0`

PR #199, opened 2023-01-31
Allow building plugins without any exposed modules

vojtechszocs:
This PR fixes an issue described by [CONSOLE-3431](https://issues.redhat.com/browse/CONSOLE-3431).

If a plugin has no exposed modules, e.g. its `exposedModules` object in the plugin metadata is either empty or missing, we should still be able to build the plugin and load it in the host application.

There are two reasons to support this use case:
- plugins which only use extensions without code references
- plugins which provide no extensions (for testing purposes)

This PR also adds an option in `DynamicRemotePlugin` to specify the name of webpack share scope object (using `default` as fallback value).

### Breaking changes in `DynamicRemotePlugin`

- removed option `moduleFederationLibraryType`
- added option `moduleFederationSettings` which contains `libraryType` and `sharedScope`

cc @gnunn1

### DIFF
--- a/packages/lib-core/src/runtime/PluginLoader.ts
+++ b/packages/lib-core/src/runtime/PluginLoader.ts
@@ -107,10 +107,11 @@ export type PluginLoaderOptions = Partial<{
   fixedPluginDependencyResolutions: Record<string, string>;
 
   /**
-   * Shared scope object for initializing `PluginEntryModule` containers.
+   * webpack share scope object for initializing `PluginEntryModule` containers.
    *
    * Host applications built with webpack should use dedicated webpack specific APIs
-   * such as `__webpack_init_sharing__` and `__webpack_share_scopes__`.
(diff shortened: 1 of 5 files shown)


packages/lib-core/src/runtime/PluginLoader.ts
```diff
@@ -107,10 +107,11 @@ export type PluginLoaderOptions = Partial<{
   fixedPluginDependencyResolutions: Record<string, string>;
 
   /**
-   * Shared scope object for initializing `PluginEntryModule` containers.
+   * webpack share scope object for initializing `PluginEntryModule` containers.
```
florkbr, line 110: @vojtechszocs nit - can we capitalize the `W` in webpack?

florkbr: Synced with Vojtech offline. Webpack has taken the stance on using lowercase `w` when referring to webpack.

PR #200, opened 2023-01-31
Add public sdkVersion property to PluginStore

vojtechszocs:
This PR gives runtime consumers of `PluginStore` the ability to determine the version of `@openshift/dynamic-plugin-sdk` package that was used in their build.

For example, host applications can print the plugin SDK version at startup for better diagnostics.

### DIFF
--- a/packages/lib-core/src/runtime/PluginStore.ts
+++ b/packages/lib-core/src/runtime/PluginStore.ts
@@ -2,6 +2,7 @@ import { v4 as uuidv4 } from 'uuid';
 import type { AnyObject } from '@monorepo/common';
 import { consoleLogger } from '@monorepo/common';
 import * as _ from 'lodash-es';
+import { version as sdkVersion } from '../../package.json';
 import type { Extension, LoadedExtension } from '../types/extension';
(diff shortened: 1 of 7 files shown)


packages/lib-core/src/runtime/PluginStore.ts
```diff
@@ -54,6 +55,8 @@ export class PluginStore implements PluginStoreInterface {
   /** Feature flags used to determine the availability of extensions. */
   private featureFlags: FeatureFlags = {};
 
+  readonly sdkVersion = sdkVersion;
```
florkbr, line 58: @vojtechszocs I am wondering if we need to go through all the effort of exposing the version or if we should just log it to the console ourselves. We already log info to the console about which plugins are loading etc.

vojtechszocs: Sorry for responding late. I think it's best to leave this to the host application, with us providing the actual version via the `sdkVersion` property. Plugin runtime infra init code is a responsibility of the host application, logging the version should be part of that code.

PR #201, opened 2023-01-31
[RHCLOUD-21699] Add workspace support

BlakeHolifield:
* Adds a hook called `useWorkspace` to get and set the workspace
* Adds helper functions to get and set workspace outside the hook
* Adds logic to create workspace based paths if workspaces are set

### DIFF
--- a/packages/lib-utils/src/hooks/useWorkspace.test.ts
+++ b/packages/lib-utils/src/hooks/useWorkspace.test.ts
@@ -21,7 +21,7 @@ describe('useWorkspace', () => {
     jest.clearAllMocks();
   });
 
-  test('unset workspace should return null', () => {
+  test('an unset workspace should return null', () => {
     const { result } = renderHook(() => useWorkspace());
     const [data, setter] = result.current;
 
@@ -29,7 +29,7 @@ describe('useWorkspace', () => {
     expect(setter).toBeDefined();
   });
 
-  test('set workspace should return the activeWorkspace', () => {
+  test('a set workspace should return the activeWorkspace', () => {
     localStorage.setItem(WORKSPACE_KEY, 'platform-experience');
     const { result } = renderHook(() => useWorkspace());
     const [data, setter] = result.current;
@@ -49,5 +49,18 @@ describe('useWorkspace', () => {
       setter('openshift');
     });
     expect(localStorage.getItem(WORKSPACE_KEY)).toBe('openshift');
+  });
+
+  test('clearing localStorage should return null', () => {
+    localStorage.setItem(WORKSPACE_KEY, 'platform-experience');
+    const { result } = renderHook(() => useWorkspace());
+    const [data, setter] = result.current;
+
+    expect(data).toBe('platform-experience');
+    expect(setter).toBeDefined();
+
+    localStorage.clear();
+
+    expect(localStorage.getItem(WORKSPACE_KEY)).toBeNull();
   });
 });
--- a/packages/lib-utils/src/hooks/useWorkspace.test.ts
+++ b/packages/lib-utils/src/hooks/useWorkspace.test.ts
@@ -1,0 +1,52 @@
+import { renderHook } from '@testing-library/react-hooks/native';
+// import { act } from 'react-dom/test-utils';
+import { useWorkspace } from './useWorkspace';
+import TestRenderer from 'react-test-renderer';
+const { act } = TestRenderer;
+
+jest.mock('react-redux', () => ({
+  useDispatch: jest.fn(() => jest.fn()),
+  useSelector: jest.fn(),
+}));
+
+const WORKSPACE_KEY = 'sdk/active-workspace';
+
+describe('useWorkspace', () => {
+  beforeEach(() => {
+    jest.resetModules();
+    localStorage.clear();
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('unset workspace should return null', () => {
+    const { result } = renderHook(() => useWorkspace());
+    const [data, _] = result.current;
+
+    expect(data).toBeNull();
+  });
+
+  test('set workspace should return the activeWorkspace', () => {
+    localStorage.setItem(WORKSPACE_KEY, 'platform-experience');
+    const { result } = renderHook(() => useWorkspace());
+    const [data, setter] = result.current;
+
+    expect(data).toBe('platform-experience');
+    expect(setter).toBeDefined();
+  });
+
+  test('updating workspace should return the activeWorkspace', () => {
+    localStorage.setItem(WORKSPACE_KEY, 'platform-experience');
+    const { result } = renderHook(() => useWorkspace());
+    const [data, setter] = result.current;
+
+    expect(data).toBe('platform-experience');
+
+    act(() => {
+      setter('openshift');
+    });
+    expect(localStorage.getItem(WORKSPACE_KEY)).toBe('openshift');
+  });
+});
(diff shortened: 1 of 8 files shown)


packages/lib-utils/src/hooks/useWorkspace.test.ts
```diff
@@ -0,0 +1,53 @@
...
+const { act } = TestRenderer;
+
+jest.mock('react-redux', () => ({
+  useDispatch: jest.fn(() => jest.fn()),
+  useSelector: jest.fn(),
+}));
+
+const WORKSPACE_KEY = 'sdk/active-workspace';
+
+describe('useWorkspace', () => {
```
florkbr, line 14: 👍 for tests

packages/lib-utils/src/hooks/useWorkspace.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  test('set workspace should return the activeWorkspace', () => {
+    localStorage.setItem(WORKSPACE_KEY, 'platform-experience');
+    const { result } = renderHook(() => useWorkspace());
+    const [data, setter] = result.current;
+
+    expect(data).toBe('platform-experience');
+    expect(setter).toBeDefined();
+  });
+
+  test('updating workspace should return the activeWorkspace', () => {
```
florkbr, line 41: @BlakeHolifield I would like to see one additional test case around clearing the workspace after it had already been set and confirm null is returned.

BlakeHolifield: Added a test after this to confirm that. Good catch!

packages/lib-utils/src/index.ts
```diff
@@ -28,7 +29,7 @@ export {
   K8sResourceListOptions,
   K8sResourceListResult,
 } from './k8s/k8s-resource';
-export { getK8sResourceURL } from './k8s/k8s-utils';
+export { getK8sResourceURL, getActiveWorkspace, setActiveWorkspace } from './k8s/k8s-utils';
```
florkbr, line 32: I'm not entirely convinced `workspaces` are a K8s concept, but I do think this is the best place for now to put these utils/hooks. We can revisit in the future 👍

packages/lib-utils/src/hooks/useWorkspace.ts
```diff
@@ -0,0 +1,51 @@
+import { consoleLogger } from '@openshift/dynamic-plugin-sdk';
+import { useState } from 'react';
+import { getActiveWorkspace, setActiveWorkspace } from '../k8s/k8s-utils';
+
+/**
+ * Hook that retrieves the active workspace from localStorage. The key for the active workspace is
+ * always `sdk/active-workspace`
```
florkbr, line 7: For whom it may concern - `sdk/` is consistent with the localstorage prefix we use in API discovery as well. We may want to eventually consolidate the prefix to one const that all location can pull from in case we decide to change it in the future.

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -15,14 +15,36 @@ import type {
...
+}
+
 const getQueryString = (queryParams: QueryParams) =>
   Object.entries(queryParams)
     .map(([key, value = '']) => `${encodeURIComponent(key)}=${encodeURIComponent(value)}`)
     .join('&');
 
 const getK8sAPIPath = ({ apiGroup = 'core', apiVersion }: K8sModelCommon) => {
+  let path = '';
+  const activeWorkspace = getActiveWorkspace();
```
spadgett, line 41: We might need to eventually allow overriding namespace for individual requests (so clients can get resources from different names), but not worry about that now and unblock the team. We can consider as a follow up.

Will anything trigger an update to watch hooks when the workspace changes?

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -43,10 +65,16 @@ export const getK8sResourceURL = (
   const { ns, name, path, queryParams } = queryOptions;
   let resourcePath = getK8sAPIPath(model);
 
+  const activeWorkspace = getActiveWorkspace();
+  if (activeWorkspace) {
+    resourcePath += `/${activeWorkspace}`;
```
florkbr, line 70: @BlakeHolifield this is duplicated with what we already do in `getK8sAPIPath` above

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -43,10 +65,16 @@ export const getK8sResourceURL = (
...
+  }
   if (resource?.metadata?.namespace) {
-    resourcePath += `/namespaces/${resource.metadata.namespace}`;
+    resourcePath += activeWorkspace
+      ? `/namespaces/${activeWorkspace}/${resource.metadata.namespace}`
```
florkbr, line 74: @BlakeHolifield I'm not sure this is the URL we expect based on our discussion with @rohitkrai03:
https://issues.redhat.com/browse/RHCLOUD-21699?focusedCommentId=21573215&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21573215

`/workspaces/<workspace>/api/namespaces/<namespace>/applications`

I don't think we should inject `activeWorkspace` after `namespaces`

packages/lib-utils/src/hooks/useWorkspace.ts
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      // get the activeWorkspace under that key if set, otherwise return null.
+      return getActiveWorkspace();
+    } catch (error) {
+      consoleLogger.error(`Failed to get activeWorkspace due to: ${error}`);
+      return null;
+    }
+  });
+
+  const setActive = (ws: string) => {
```
florkbr, line 38: We may want to consider using another param name than `ws`. When I see `ws` I think websocket

packages/lib-utils/src/k8s/k8s-utils.ts
spadgett: @florkbr Sorry, I just edited my comment. I meant _workspaces_, not namespaces. We might have a use case in the future where clients need to get resources from both workspace-a and workspace-b so there's not a single context. We'd need some prop in the request to specify the workspace and fall back to the active workspace when not present.

Also we'll want to make sure the k8s hooks trigger a re-render when the workspace context is updated.

florkbr: @spadgett ah - I see your point now. We actually started down that path already via `pathPrefix`. See: https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts#L36

The idea would be a plugin could use the hooks with different `pathPrefix` set to achieve fetching resources from multiple workspaces at once. However there are a few challenges:
1. There is a bug with `pathPrefix` (more specifically passing any options object to the hook) that causes an infinite rendering loop
2. The hook expects API discovery to load the models. We have not yet updated API discovery to fetch workspace specific resources. We could however update the hook to accept a static model (versus relying on API discovery), but that would push the onus onto the plugin to provide the model. I think we mentioned the workspace model may not change much, so maybe that is ok? 

Finally - we will follow up in another PR to update the hooks to re-render when the active workspace changes (and indicate loading is occurring).

spadgett: @florkbr Sounds good. When using dev sandbox, the resources will be the same regardless of workspace, so we don't need to trigger discovery again.

florkbr:
I think this is a good first step. We will need to revisit in a follow up on Sam's comment here: https://github.com/openshift/dynamic-plugin-sdk/pull/201#discussion_r1095068536 as this implementation won't scale to fetching resources from multiple workspaces at the same time.

PR #202, opened 2023-02-06
[RHCLOUD-21699] Update watches on workspace change

BlakeHolifield:
In order to refresh the watches on each resource when a workspace changes, we've added a Provider and publish/subscribe model to the workspace switcher. On a workspace change, the `watchK8sResources` hooks will refire to update the websocket connections to the new URIs.

We will need an additional PR to hac-core to actually initialize the provider.

If this approach gets the full 👍 , I'll update the tests, docs, and remove it from Draft.

### DIFF
--- a/packages/lib-utils/src/hooks/useWorkspace.ts
+++ b/packages/lib-utils/src/hooks/useWorkspace.ts
@@ -1,6 +1,8 @@
 import { consoleLogger } from '@openshift/dynamic-plugin-sdk';
-import { useState } from 'react';
-import { getActiveWorkspace, setActiveWorkspace } from '../k8s/k8s-utils';
+import { useContext, useEffect, useReducer } from 'react';
+import { setActiveWorkspaceLocalStorage } from '../k8s/k8s-utils';
+import WorkspaceContext from '../utils/WorkspaceContext';
+import { UpdateEvents } from '../utils/workspaceState';
 
 /**
  * Hook that retrieves the active workspace from localStorage. The key for the active workspace is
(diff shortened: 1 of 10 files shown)


packages/lib-utils/src/hooks/useWorkspace.ts
```diff
@@ -18,34 +20,31 @@ import { getActiveWorkspace, setActiveWorkspace } from '../k8s/k8s-utils';
  * ```
  */
 export const useWorkspace = () => {
+  const { subscribe, unsubscribe, getState, setWorkspaceContext } = useContext(WorkspaceContext);
+  const [, forceRender] = useReducer((count) => count + 1, 0);
```
florkbr, line 24: @Hyperkid123 I'm a bit worried this is an anti-pattern WRT hooks in React. See their doc here: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate

> Normally, you shouldn’t mutate local state in React. However, as an escape hatch, you can use an incrementing counter to force a re-render even if the state has not changed
>
> Try to avoid this pattern if possible.

While this will work, it breaks the flow of updates and immutable data. Can we use something else to cause a re-render? Maybe some ID associated with the workspace?

Hyperkid123: The issue is you want a mutable data structure (localStorage) to trigger react render cycle that depends purely on immutable data. So you have to force it. If you take a deep dive into some state management libraries that are framework agnostic, you will find that their react bindings do something very similar if not exactly the same thing.

The only two things that can trigger react render cycle are either prop updates or state updates. We got the state here to help us.

reports/lib-utils.api.md
```diff
@@ -667,7 +670,7 @@ export type Selector = Partial<{
 }>;
 
 // @public (undocumented)
-export function setActiveWorkspace(workspace: string): void;
+export function setActiveWorkspaceLocalStorage(workspace: string): void;
```
florkbr, line 673: FYI We will publish a new version of lib-utils post merging this PR. We will also update the major version to indicate breaking changes.

packages/lib-utils/src/hooks/useWorkspace.ts
florkbr: I synced with Martin and Blake offline. For now we will leave this in with a comment above explaining the purpose. It should pose little performance risk as of now.

packages/lib-utils/src/utils/WorkspaceProvider.tsx
florkbr, line 1: @rohitkrai03 FYI. I know we're currently using the `ContextProvider` extension to manage the workspace in hac-dev, however I suspect in the future the SDK will need access to the Workspace context so it makes sense to move it here.

florkbr:
Tested locally and hacked into HAC Core and seems fine. :shipit:

packages/lib-utils/src/hooks/useWorkspace.ts
vojtechszocs: FYI, OpenShift Console has a dedicated [`useForceRender`](https://github.com/openshift/console/blob/master/frontend/packages/console-plugin-shared/src/hooks/useForceRender.ts) hook for this purpose.

Forcing a re-render of the given React component due to a change in certain conditions that happen outside React control flow is not necessarily an anti-pattern.

PR #204, opened 2023-02-07
Add support for custom plugin properties

vojtechszocs:
This PR allows plugins to pass custom properties to host applications.

For example, [OpenShift Console dynamic plugins](https://github.com/openshift/console/tree/master/frontend/packages/console-dynamic-plugin-sdk) can specify additional properties such as `displayName`, `description` and `disableStaticPlugins` within their plugin metadata.

This PR adds new `customProperties` field to contain all such application or environment specific properties.

At runtime, these properties can be accessed via loaded plugin info entries, for example:
```ts
const infoEntries = usePluginInfo();

infoEntries.forEach((entry) => {
  if (entry.status === 'loaded') {
    // do something with entry.metadata.customProperties
  }
});
```

### Other code changes
- update `PluginInfoTable` component - add action dropdown to log custom properties
- simplify `FeatureFlagTable` component - remove unnecessary `React.useState` call
- remove `AnyFunction` common type since it seems to be unused

### DIFF
--- a/packages/common/src/types/common.ts
+++ b/packages/common/src/types/common.ts
@@ -8,16 +8,6 @@
 export type AnyObject = Record<string, unknown>;
 
 /**
- * The type `Function` doesn't provide enough type safety.
- *
- * Use the `AnyFunction` type for arbitrary functions whose signature is unknown.
- *
- * @see https://github.com/typescript-eslint/typescript-eslint/issues/1896
- */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type AnyFunction = (...args: any) => any;
-
-/**
  * Replace existing direct properties of `T` with ones declared in `R`.
  */
 export type ReplaceProperties<T, R> = {
--- a/packages/lib-core/src/runtime/PluginStore.ts
+++ b/packages/lib-core/src/runtime/PluginStore.ts
@@ -155,8 +155,8 @@ export class PluginStore implements PluginStoreInterface {
 
     Array.from(this.loadedPlugins.entries()).forEach(([pluginName, plugin]) => {
       entries.push({
-        pluginName,
         status: 'loaded',
+        pluginName,
         metadata: plugin.metadata,
         enabled: plugin.enabled,
         disableReason: plugin.disableReason,
@@ -165,8 +165,8 @@ export class PluginStore implements PluginStoreInterface {
 
     Array.from(this.failedPlugins.entries()).forEach(([pluginName, plugin]) => {
       entries.push({
-        pluginName,
         status: 'failed',
+        pluginName,
         errorMessage: plugin.errorMessage,
         errorCause: plugin.errorCause,
       });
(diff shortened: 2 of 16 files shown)


packages/lib-core/src/runtime/PluginStore.ts
```diff
@@ -155,18 +150,19 @@ export class PluginStore implements PluginStoreInterface {
 
     Array.from(this.loadedPlugins.entries()).forEach(([pluginName, plugin]) => {
       entries.push({
-        pluginName,
         status: 'loaded',
-        metadata: plugin.metadata,
+        pluginName,
+        metadata: _.pick(plugin.manifest, ['name', 'version', 'dependencies']),
+        customProperties: plugin.manifest.customProperties,
```
vojtechszocs, line 156: @florkbr Right now, we're just passing `manifest.customProperties` through to `entry.customProperties`.

Would it be better to add `?? {}` fallback here to avoid `entry.customProperties` being `undefined` ?

karelhala:
@vojtechszocs oooh! I really like this approach. Didn't knew about it when I've opened #205 (gotta close is in favor of this one). This one coveres more ground and adds a couple more information - I like the example part!

florkbr:
Tested locally and looks good to me (also modified and added/removed customProperties). 
image

packages/common/src/types/common.ts
```diff
@@ -7,16 +7,6 @@
...
 
-/**
- * The type `Function` doesn't provide enough type safety.
- *
- * Use the `AnyFunction` type for arbitrary functions whose signature is unknown.
- *
- * @see https://github.com/typescript-eslint/typescript-eslint/issues/1896
- */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type AnyFunction = (...args: any) => any;
```
florkbr, line 18: 👍 for removing. Not used anywhere as far as I can tell.

packages/lib-core/src/types/extension.ts
```diff
@@ -15,7 +15,6 @@ export type Extension<TType extends string = string, TProperties extends AnyObje
 export type LoadedExtension<TExtension extends Extension = Extension> = TExtension & {
   pluginName: string;
   uid: string;
-  [customProperty: string]: unknown;
```
florkbr, line 18: I don't think HAC Dev or HAC Core rely on this logic AFAIK. @vojtechszocs does this have any implications to OCP Console plugins?

vojtechszocs: @florkbr The `Extension` type already has the "allow additional properties whose value is unknown" declaration:

```ts
export type Extension<TType extends string = string, TProperties extends AnyObject = AnyObject> = {
  type: TType;
  properties: TProperties;
  flags?: ExtensionFlags;
  [customProperty: string]: unknown;
};
```

The `LoadedExtension` type is the (augmented) runtime version of the base `Extension` type, it inherits from it via `&` which is why the `[customProperty: string]: unknown` is superfluous in `LoadedExtension`. This is why I removed it.

```ts
export type LoadedExtension<TExtension extends Extension = Extension> = TExtension & {
  pluginName: string;
  uid: string;
};
```

There should be no change at all. You may still add custom properties onto `LoadedExtension` instances.

florkbr: Perfect. I missed that declaration. Thanks!

PR #205, opened 2023-02-15
Add plugin data object to webpack plugin metadata

karelhala:
### Description

We want to consume custom values from plugin in any hosting app (currently needed for apiAllowedList). Let's add custom prop `pluginData` so hosting app can consume these values.

### JIRA

https://issues.redhat.com/browse/RHCLOUD-23122


karelhala:
Closing in favor of https://github.com/openshift/dynamic-plugin-sdk/pull/204

PR #206, opened 2023-02-22
RHCLOUD-24456: Add base URL to plugin manifest

vojtechszocs:
- [x] add new mandatory field `baseURL` to plugin manifest, computed from [webpack `output.publicPath`](https://webpack.js.org/configuration/output/#outputpublicpath) option
- [x] `baseURL` should have a trailing slash as per common webpack recommendation (if not, emit a build warning)
- [x] expose processed manifest object to consumers at runtime via `LoadedPluginInfoEntry` type - contains everything from raw manifest, minus extensions, plus fallbacks for optional properties missing in raw manifest
- [x] `sample-plugin/plugin-metadata.json` :arrow_right: `sample-plugin/plugin-metadata.ts` for better type checking


vojtechszocs:
> make `semver` related imports more specific for better module tree shaking

Using [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) shows that
```ts
import { valid as semverValid, satisfies as semverSatisfies } from 'semver';
import type { RangeOptions as SemverRangeOptions } from 'semver';
```
vs.
```ts
import * as semver from 'semver';
```
has no impact on which `semver` code gets pulled into the webpack compilation of the sample application.

vojtechszocs:
/hold

Some small improvements incoming.

PR #207, opened 2023-02-27
Clone manifest extensions

vojtechszocs:
`PluginStore` will use `_.cloneDeep(manifest.extensions)` to compute processed extensions (via `decodeCodeRefs`).

This allows us to safely expose the original raw manifest object including the original `extensions` array.

### Other changes
- lib-core
  - `PluginStore.postProcessExtensions` option removed in favor of `PluginLoader.postProcessManifest`
- lib-webpack
  - `DynamicRemotePlugin` options `pluginMetadata` and `extensions` are now required, encouraging developers to use TypeScript instead of plain JSON files
  - Removed `glob` dependency since it's not needed anymore


karelhala (commented):
I don't have any objections to this. If nobody else reviews I'll mark as lgtm tomorrow

packages/lib-core/src/runtime/PluginStore.ts
```diff
@@ -23,13 +18,6 @@ export type PluginStoreOptions = Partial<{
    * Default value: `true`.
    */
   autoEnableLoadedPlugins: boolean;
-
-  /**
-   * Post-process the plugin's extension objects.
-   *
-   * By default, no post-processing is performed on the extension objects.
-   */
-  postProcessExtensions: (extensions: LoadedExtension[]) => LoadedExtension[];
```
florkbr, line 32: Chatted with @vojtechszocs and this post process of extensions was not utilized and didn't really make sense to expose with postProcessManifest.

packages/lib-core/src/runtime/PluginStore.ts
```diff
@@ -281,30 +268,24 @@ export class PluginStore implements PluginStoreInterface {
...
-      registrationMethod: manifest.registrationMethod,
-      buildHash: manifest.buildHash ?? uuidv4(),
-    };
-
-    const processedExtensions = this.processExtensions(
-      pluginName,
-      processedManifest.buildHash,
-      manifest.extensions,
-      entryModule,
+    const loadedExtensions = _.cloneDeep(manifest.extensions).map<LoadedExtension>((e, index) =>
```
florkbr, line 274: Confirmed with @vojtechszocs this clone operation only occurs once when the plugin is added to the store and should have minimal performance impact. We may need to revisit this in the future if we have an overwhelming amount of extensions or plugins themselves.

PR #208, opened 2023-02-27
Add technical compatibility with React 18

vojtechszocs:
React `18.0.0` was published on 2022-03-29 and some projects might have already upgraded from React 17 to 18.

This PR adds technical compatibility with React 18 in terms of package peer dependencies.

- Plugin SDK packages will continue to be built with React 17
- Sample plugin and host app will continue to use React 17

Once available and stable, we will switch from PatternFly 4 to 5 (and using React 18) within the dynamic plugin SDK repo.


karelhala (approved):
/lgtm

PR #209, opened 2023-02-28
Add webpack bundle analyzer to sample application

vojtechszocs:
Running `yarn analyze` in `packages/sample-app` directory will start a webpack production build with [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) plugin applied to build configuration.

The bundle report is generated to a single `packages/sample-app/dist/bundle-report.html` file. This can be useful when looking into which modules make up the resulting application and the level of module optimization per each used package.


karelhala (approved):
/lgtm

PR #212, opened 2023-03-15
Improve PluginLoader and PluginStore

vojtechszocs:
This PR refactors `PluginLoader` and `PluginStore` in order to simplify and improve existing APIs.

## Change overview

We can now track plugins which are currently being loaded (status `pending`).

Plugin manifest object is now available regardless of the plugin's current status.

`PluginLoader.loadPlugin` method now accepts either a manifest object or its URL address.

```diff
-  loadPlugin: (baseURL: string, manifestNameOrObject?: string | PluginManifest) => Promise<void>;
+  loadPlugin: (manifest: PluginManifest | string, forceReload?: boolean) => Promise<void>;
```

Sample application now shows an error message tooltip for plugins that failed to load properly.

### Add `loaderOptions` to `PluginStore`

`PluginLoader` class is treated as an implementation detail. It is now configured via `PluginStore` options, for example:

```ts
const pluginStore = new PluginStore({
  loaderOptions: {
    sharedScope: getSharedScope(),
  },
});
```

### Refactor `PluginLoader`

- `loadPlugin` accepts `PluginManifest` parameter and returns `Promise<PluginLoadResult>`
- `PluginManifest` loading and processing moved out to separate methods
- `subscribe` removed due to `loadPlugin` returning `Promise<PluginLoadResult>`
- `loadPluginScripts` uses `manifest.baseURL` when computing plugin script URLs
- `resolvePluginDependencies` uses internal `loadListeners` structure instead of `subscribe`

### Changes in `PluginLoaderOptions`

- Remove `entryCallbackName` in favor of `entryCallbackSettings.name`
- Add `entryCallbackSettings.autoRegisterCallback` to control global callback auto-registration

### Refactor `PluginStore`

- `setLoader` removed, use `loaderOptions` instead
- If the plugin is still loading, `loadPlugin` returns the same Promise which handles the load operation
- If the plugin is loaded and `forceReload` is not `true`, `loadPlugin` returns `Promise.resolve()`
- Once the plugin load operation completes, the associated Promise reference is cleared

### Changes in `PluginStoreOptions`

- Add `loaderOptions` to be passed to `PluginLoader` instance managed by the `PluginStore`

### DIFF
--- a/packages/lib-core/src/runtime/PluginLoader.ts
+++ b/packages/lib-core/src/runtime/PluginLoader.ts
@@ -71,11 +71,11 @@ export type PluginLoaderOptions = Partial<{
    */
   entryCallbackSettings: Partial<{
     /**
-     * Control whether to automatically register the callback function.
+     * Control whether to register the callback function.
      *
      * Default value: `true`.
      */
-    autoRegisterCallback: boolean;
+    registerCallback: boolean;
 
     /**
      * Name of the callback function.
@@ -167,9 +167,7 @@ export class PluginLoader {
       getPluginEntryModule: options.getPluginEntryModule ?? _.noop,
     };
 
-    if (this.options.entryCallbackSettings.autoRegisterCallback ?? true) {
-      this.registerPluginEntryCallback();
-    }
+    this.registerPluginEntryCallback();
   }
 
   private invokeLoadListeners() {
@@ -383,7 +381,7 @@ export class PluginLoader {
       let isResolutionComplete = false;
       let listener: VoidFunction;
 
-      const resolutionComplete = () => {
+      const setResolutionComplete = () => {
         isResolutionComplete = true;
         this.loadListeners.delete(listener);
       };
@@ -418,7 +416,7 @@ export class PluginLoader {
 
         if (resolutionErrors.length > 0) {
           const errorTitle = `Detected ${resolutionErrors.length} resolution errors with ${pendingDepInfo}`;
-          resolutionComplete();
+          setResolutionComplete();
           reject(new Error(`${errorTitle}:\n\n${resolutionErrors.join('\n')}`));
           return;
         }
@@ -426,7 +424,7 @@ export class PluginLoader {
         consoleLogger.info(`Plugin ${pluginName} has ${pendingDepInfo}`);
 
         if (pendingDepNames.length === 0) {
-          resolutionComplete();
+          setResolutionComplete();
           resolve();
         }
       };
@@ -466,7 +464,13 @@ export class PluginLoader {
    * This must be called in order to load plugins using the `callback` registration method.
    */
   registerPluginEntryCallback() {
+    const registerCallback = this.options.entryCallbackSettings.registerCallback ?? true;
     const callbackName = this.options.entryCallbackSettings.name ?? DEFAULT_REMOTE_ENTRY_CALLBACK;
+
+    if (!registerCallback) {
+      consoleLogger.info(`Plugin entry callback ${callbackName} will not be registered`);
+      return;
+    }
 
     if (typeof window[callbackName] === 'function') {
       consoleLogger.warn(`Plugin entry callback ${callbackName} is already registered`);
--- a/packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
+++ b/packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
@@ -10,6 +10,9 @@ import { GenerateManifestPlugin } from './GenerateManifestPlugin';
 import { PatchEntryCallbackPlugin } from './PatchEntryCallbackPlugin';
 import { ValidateCompilationPlugin } from './ValidateCompilationPlugin';
 
+const DEFAULT_MANIFEST = 'plugin-manifest.json';
+const DEFAULT_ENTRY_SCRIPT = 'plugin-entry.js';
+
 /**
  * Settings for the webpack `ModuleFederationPlugin`.
  */
@@ -128,8 +131,8 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
       sharedModules: options.sharedModules ?? {},
       moduleFederationSettings: options.moduleFederationSettings ?? {},
       entryCallbackSettings: options.entryCallbackSettings ?? {},
-      entryScriptFilename: options.entryScriptFilename ?? 'plugin-entry.js',
-      pluginManifestFilename: options.pluginManifestFilename ?? 'plugin-manifest.json',
+      entryScriptFilename: options.entryScriptFilename ?? DEFAULT_ENTRY_SCRIPT,
+      pluginManifestFilename: options.pluginManifestFilename ?? DEFAULT_MANIFEST,
     };
 
     try {
--- a/reports/lib-core.api.md
+++ b/reports/lib-core.api.md
@@ -151,7 +151,7 @@ export type PluginLoaderOptions = Partial<{
     canLoadPlugin: (manifest: PluginManifest, reload: boolean) => boolean;
     canReloadScript: (manifest: PluginManifest, scriptName: string) => boolean;
     entryCallbackSettings: Partial<{
-        autoRegisterCallback: boolean;
+        registerCallback: boolean;
         name: string;
     }>;
     fetchImpl: ResourceFetch;
--- a/packages/lib-core/src/types/store.ts
+++ b/packages/lib-core/src/types/store.ts
@@ -102,25 +102,24 @@ export type PluginStoreInterface = {
   /**
    * Start loading a plugin from the given manifest.
    *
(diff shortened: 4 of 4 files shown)


packages/sample-app/src/components/common/LabelWithTooltipIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+import * as React from 'react';
+
+type LabelWithTooltipIconProps = {
+  label: React.ReactNode;
+  tooltipContent?: React.ReactNode;
+};
+
+const LabelWithTooltipIcon: React.FC<LabelWithTooltipIconProps> = ({ label, tooltipContent }) => {
+  if (!tooltipContent) {
+    return label as React.ReactElement;
```
vojtechszocs, line 14: This cast was added to work around a TypeScript typing issue.

One solution would be a fragment `<>{label}</>` but this triggers an ESLint warning about avoiding single-element fragments.

On the other hand, typing `label` as `ReactElement` would restrict users from passing e.g. `undefined` or `boolean` or React fragment etc.

This is basically a `ReactNode` input :arrow_right: `React.FC` output passthrough issue.

@florkbr @Hyperkid123 Is there some obvious simple solution to this that I'm missing?

Hyperkid123: Yeah, this is fine. I've encountered this many times. I don't think there is a nice solution to this. Technically `undefined` or `boolean` is not a valid ReactNode. Bot If it was not supposed to be permitted then react should throw a runtime error. I think the type cast is OK.

vojtechszocs: @Hyperkid123 Thanks :+1: I've found some discussions around this issue, for example:

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18051#issuecomment-645447374

Some devs even advocate not using `React.FC` type at all:

https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript

The technically correct solution (type wise) would be:

```tsx
// rule suppression below can be avoided with "allowExpressions: true" config
// eslint-disable-next-line react/jsx-no-useless-fragment
return React.isValidElement(label) ? label : <>{label}</>;
```

but that's just too complex for common use, so I'll keep the type cast.

vojtechszocs: Hopefully React 18 has addressed this already :smiley:

packages/sample-app/src/components/common/LoadPluginModal.tsx
```diff
@@ -1,38 +1,49 @@
...
+      [setForceReload],
+    );
+
+    const closeModal = React.useCallback(() => {
       setModalOpen(false);
-    }, [baseURL, pluginStore, setModalOpen]);
+    }, [setModalOpen]);
+
+    const loadPlugin = React.useCallback(() => {
+      pluginStore.loadPlugin(manifestURL, forceReload);
```
florkbr, line 44: @vojtechszocs would it be possible for us to add another scenario where we provide the manifest itself (over the URL) to test that code-path as well?

florkbr (commented):
Tested the sample app/plugin locally and looking good. @vojtechszocs posting a few requests. Let me know what you think! 
image

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -54,7 +51,7 @@ export type PluginModuleFederationSettings = Partial<{
  */
 export type PluginEntryCallbackSettings = Partial<{
   /**
-   * Name of the function to call.
+   * Name of the callback function.
```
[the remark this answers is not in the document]
vojtechszocs: Well, each method is a function, but not each function is a method :smiley: 

In this case, the global callback function is just a global function, although technically it's a method that belongs to the global `window` object.

:+1: on staying consistent with the naming.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -132,7 +129,7 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
       moduleFederationSettings: options.moduleFederationSettings ?? {},
       entryCallbackSettings: options.entryCallbackSettings ?? {},
       entryScriptFilename: options.entryScriptFilename ?? 'plugin-entry.js',
-      pluginManifestFilename: options.pluginManifestFilename ?? DEFAULT_PLUGIN_MANIFEST,
+      pluginManifestFilename: options.pluginManifestFilename ?? 'plugin-manifest.json',
```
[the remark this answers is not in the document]
vojtechszocs: Agreed, I'll move these two strings to local constants.

packages/sample-app/src/components/common/LoadPluginModal.tsx
vojtechszocs: Sounds good to me, the only question is what would be the best way to represent the manifest object input?

Aside from doing this via UI, we could always expose `pluginStore` via the global `window` object, so developers can interact with it directly, for example:
```ts
pluginStore.loadPlugin('http://localhost:9001/plugin-manifest.json');
pluginStore.getPluginInfo();
```

FYI, OpenShift Console does this for non-production builds ([link](https://github.com/openshift/console/blob/7500dbd9fe588c7e83e92f79d9ba32bbd61dc8e3/frontend/public/plugins.ts#L38)).

This could make local testing via the sample application a bit easier.

packages/sample-app/src/components/common/PageLayout.tsx
```diff
@@ -1,9 +1,9 @@
 import { Page, PageSection } from '@patternfly/react-core';
 import * as React from 'react';
 
-type PageLayoutProps = {
+type PageLayoutProps = React.PropsWithChildren<{
```
vojtechszocs, line 4: FYI, the reason I've used `PropsWithChildren` here is to better communicate the fact that this component is supposed to have children.

In React 17, `React.FC` type automatically uses `PropsWithChildren` which adds `children?: ReactNode | undefined` even if the component is _not_ supposed to have children by design (so React is basically imposing a hard assumption).

This is one of the downsides of using `React.FC` as mentioned in https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript - for example, `<Foo>something</Foo>` expression is valid even if the `Foo` component is not supposed to have children :smiley: 

This issue should be fixed in React 18.
- https://solverfox.dev/writing/no-implicit-children/
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210 (see "Removal of implicit children")

reports/lib-core.api.md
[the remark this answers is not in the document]
vojtechszocs: Yes, we'll bump the major version of lib-core due to breaking API changes in this package.

packages/lib-core/src/runtime/PluginLoader.ts
```diff
@@ -148,54 +153,54 @@ export class PluginLoader {
...
+      entryCallbackSettings: options.entryCallbackSettings ?? {},
       fetchImpl: options.fetchImpl ?? basicFetch,
       fixedPluginDependencyResolutions: options.fixedPluginDependencyResolutions ?? {},
       sharedScope: options.sharedScope ?? {},
       postProcessManifest: options.postProcessManifest ?? _.identity,
-      entryCallbackName: options.entryCallbackName ?? DEFAULT_REMOTE_ENTRY_CALLBACK,
       getPluginEntryModule: options.getPluginEntryModule ?? _.noop,
     };
+
+    if (this.options.entryCallbackSettings.autoRegisterCallback ?? true) {
```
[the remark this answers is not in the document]
vojtechszocs: Personally I'd prefer to perform this kind of check outside the function, since
```ts
if (doNothingCondition) {
  return;
}
```
adds complexity without much gain.

I agree that the `if` statement expression in
```ts
if (this.options.entryCallbackSettings.autoRegisterCallback ?? true) {
  this.registerPluginEntryCallback();
}
```
looks a bit weird. I'll try to improve that :smiley:

packages/sample-app/src/components/common/LoadPluginModal.tsx
florkbr: It might be nice to have a UI option to enter the object (for discoverability purposes). Perhaps code editor (https://www.patternfly.org/v4/components/code-editor) or multi line inputs (https://www.patternfly.org/v4/components/text-input) with the prefilled JSON? I do think it's nice to expose the pluginStore for debugging non-prod builds too. 

We can follow up in another PR for this IMO.

florkbr: Chatted with Vojtech and we will follow up in another PR with exposing the plugin loader/store in the sample app.

PR #217, opened 2023-04-10
RHCLOUD-23119: Allow for on demand API discovery through hooks

florkbr:
### Success case: demonstration showing fetch is only called once per model
https://user-images.githubusercontent.com/21317056/231302312-a288291d-5c8f-4815-9c22-c25f04967adf.mp4

### Failure case: demonstration showing fetch will be limited to only 3 calls for a given model
Screenshot 2023-04-11 at 6 09 28 PM

https://issues.redhat.com/browse/RHCLOUD-23119


packages/lib-utils/src/k8s/hooks/useK8sWatchResource.ts
```diff
@@ -76,6 +83,10 @@ export const useK8sWatchResource = <R extends K8sResourceCommon | K8sResourceCom
     }
     if (!resourceK8s) {
       const data = resource.isList ? [] : {};
+      if (!k8sModel && !inFlight && !batchesInFlight && retries < 3) {
+        fetchModel(resource, dispatch);
+        setRetries(retries + 1);
+      }
       return modelsLoaded && !k8sModel && !batchesInFlight
         ? [data, true, new NoModelError()]
```
florkbr, line 91: We may need to prevent this return value from returning while we're still trying to fetch the model on demand above. While I was testing locally this didn't _seem_ to be an issue, but I'll continue looking.

packages/lib-utils/src/k8s/k8s-utils.ts
```diff
@@ -235,6 +236,35 @@ export const getGroupVersionKindForModel: GetGroupVersionKindForModel = ({
   kind,
 });
 
+/**
+ * Provides the K8sResourceIdentifier for the given K8sGroupVersionKind or K8sResourceKindReference.
```
florkbr, line 240: FWIW this is not exposed to SDK consumers, just doc'ed internally for SDK devs.

karelhala (commented):
Looking good! I wouldn't include `onlyPreferenceList` in this PR. We should work on allowing plugins listing models to watch for and any other model can be on demand trough these new hooks.

florkbr:
@karelhala I'm not sure I follow. At some point won't we need to fallback on fetching the model if it's not in the list plugins provide?

karelhala:
Yes, that's the aim of this PR isn't it? The idea is that plugins will provide default models we'll watch for. If extension would want to watch for another model they'll use these new hooks.

PR #219, opened 2023-04-13
RHCLOUD-21652: Don't suppress failed call status errors

florkbr:
https://issues.redhat.com/browse/RHCLOUD-21652

### DIFF
--- a/packages/lib-utils/src/k8s/k8s-errors.ts
+++ b/packages/lib-utils/src/k8s/k8s-errors.ts
@@ -1,0 +1,14 @@
+import { CustomError } from '@openshift/dynamic-plugin-sdk';
+import type { K8sStatus } from '../types/k8s';
+
+/**
+ * Error class used when Kubernetes cannot handle a request.
+ */
+export class K8sStatusError extends CustomError {
+  public statusObject: K8sStatus;
+
+  constructor(statusObject: K8sStatus) {
+    super(statusObject.message);
+    this.statusObject = statusObject;
+  }
+}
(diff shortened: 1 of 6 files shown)


reports/lib-utils.api.md
```diff
@@ -10,6 +10,7 @@ import type { ActionType as ActionType_2 } from 'typesafe-actions';
 import type { AnyAction } from 'redux';
 import type { AnyObject } from '@openshift/dynamic-plugin-sdk';
 import { Context } from 'react';
+import { CustomError } from '@openshift/dynamic-plugin-sdk';
```
florkbr, line 13: FWIW the CustomError class is exported through the lib-core package, not here.

andrewballantyne:
It should be noted this is a backwards breaking change -- where once you didn't crash and burn, and now this will cause your thread to break -- which could lead to error boundaries (if you have them) or to whitescreening your app using the SDK.

packages/lib-utils/src/k8s/k8s-errors.ts
```diff
@@ -0,0 +1,11 @@
+import { CustomError } from '@openshift/dynamic-plugin-sdk';
+import type { K8sStatus } from '../types/k8s';
+
+/**
+ * Error class used when Kubernetes cannot handle a request.
+ */
+export class K8sStatusError extends CustomError {
+  constructor(readonly statusObject: K8sStatus) {
```
vojtechszocs, line 8: Nit: I'd omit the "Object" suffix for simplicity, since `K8sStatus` refers to object type definition anyway.

```suggestion
  constructor(readonly status: K8sStatus) {
```

packages/lib-utils/src/utils/common-fetch.ts
```diff
@@ -65,10 +67,14 @@ export const commonFetchJSON = async <TResult>(
     timeout,
     isK8sAPIRequest,
   );
+  const responseText: string = await response.text();
```
vojtechszocs, line 70: This will be implicitly typed as `string` since `response.text()` returns `Promise<string>`.

```suggestion
  const responseText = await response.text();
```

vojtechszocs (changes requested):

PR #220, opened 2023-04-13
Prepare for next release of dist packages

vojtechszocs:
- lib-core :arrow_right: `4.0.0`
- lib-extensions :arrow_right: `1.3.0`
- lib-utils :arrow_right: `2.1.0`
- lib-webpack :arrow_right: `3.0.1`

### DIFF
--- a/packages/lib-extensions/package.json
+++ b/packages/lib-extensions/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openshift/dynamic-plugin-sdk-extensions",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "description": "Provides extension types for dynamic plugins",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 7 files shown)


packages/lib-extensions/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@openshift/dynamic-plugin-sdk-extensions",
-  "version": "1.2.0",
+  "version": "1.3.0",
```
vojtechszocs, line 3: By convention, we bump the minor semver number due to change in package dependencies.

packages/lib-webpack/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@openshift/dynamic-plugin-sdk-webpack",
-  "version": "3.0.0",
+  "version": "3.0.1",
```
vojtechszocs, line 3: We bump the bugfix semver number since there were no changes in package dependencies and the changelog doesn't warrant a minor version bump either.

PR #224, opened 2023-06-02
RHCLOUD-21655: k8sCreate (POST) API behavior deviation from K8s server

florkbr:
See https://issues.redhat.com/browse/RHCLOUD-21655

### DIFF
--- a/packages/lib-utils/src/k8s/k8s-resource.ts
+++ b/packages/lib-utils/src/k8s/k8s-resource.ts
@@ -62,7 +62,7 @@ export const k8sCreateResource = <
   fetchOptions = {},
 }: K8sResourceUpdateOptions<TResource>): Promise<TCreatedResource> =>
   commonFetchJSON.post<TCreatedResource>(
-    getK8sResourceURL(model, resource, queryOptions),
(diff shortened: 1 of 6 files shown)


packages/lib-utils/src/k8s/k8s-resource.ts
```diff
@@ -62,7 +63,7 @@ export const k8sCreateResource = <
   fetchOptions = {},
 }: K8sResourceUpdateOptions<TResource>): Promise<TCreatedResource> =>
   commonFetchJSON.post<TCreatedResource>(
-    getK8sResourceURL(model, resource, queryOptions),
+    getK8sResourceURL(model, resource, omit(queryOptions, 'queryParams')),
```
andrewballantyne, line 66: What is the URL out the other end -- because wouldn't the `resource` still make use of the name?

florkbr: Yeah. I jumped the gun on opening a PR. I am working on addressing the name and workspace path params.

florkbr:
Trying one more time to see if gateway timeouts resolve

PR #228, opened 2023-08-02
Improve error handling

vojtechszocs:
This PR allows the host application to handle both load-time and runtime errors related to plugins.

- new `PluginStore` option `pluginErrorHandler`
  - if `errorDetails.loadError` is `true`, the error has occurred while loading the given plugin
  - if `errorDetails.loadError` is `false`, the error was reported for an already loaded plugin (see below)
```ts
pluginErrorHandler: (manifest: PluginManifest, errorDetails: PluginErrorDetails) => void;
```

- new `PluginStore` interface method `reportPluginError`
  - if the given plugin is currently loaded, pass provided error details to `pluginErrorHandler`
  - if the given plugin is not currently loaded, issue a console warning and do nothing
```ts
reportPluginError: (
  pluginName: string,
  reportedBy: string,
  errorMessage: string,
  errorCause?: unknown,
) => void;
```

- `useResolvedExtensions` hook reports code reference resolution errors via `reportPluginError` API

- `useResolvedExtensions` hook _**does not**_ disable problematic plugins by default anymore
  - aligns with openshift/console#12347
  - if necessary, problematic plugins can be disabled via the application provided `pluginErrorHandler`
```ts
const pluginStore = new PluginStore({
  // ... other options ...
  pluginErrorHandler: (manifest, errorDetails) => {
    if (!errorDetails.loadError && errorDetails.reportedBy === 'useResolvedExtensions') {
      pluginStore.disablePlugins([manifest.name], 'Errors while resolving code references');
    }
  },
});
```

### DIFF
--- a/packages/lib-core/src/runtime/PluginStore.ts
+++ b/packages/lib-core/src/runtime/PluginStore.ts
@@ -12,6 +12,16 @@ import { decodeCodeRefs, getPluginModule } from './coderefs';
 import { PluginLoader } from './PluginLoader';
 import type { PluginLoaderOptions } from './PluginLoader';
 
+export type PluginErrorDetails = { message: string; cause?: unknown } & (
+  | {
+      loadError: true;
+    }
+  | {
+      loadError: false;
+      reportedBy: string;
+    }
+);
+
 export type PluginStoreOptions = Partial<{
   /**
    * Options passed to `PluginLoader` instance managed by the `PluginStore`.
@@ -26,6 +36,19 @@ export type PluginStoreOptions = Partial<{
    * Default value: `true`.
    */
   autoEnableLoadedPlugins: boolean;
+
+  /**
+   * Custom plugin error handler.
+   *
+   * The `errorDetails.loadError` property indicates the type of the error:
+   * - if `true`, the error has occurred while loading the given plugin
+   * - if `false`, the error was reported for an already loaded plugin
+   *
+   * By default, this function does nothing.
+   *
+   * @see {@link PluginStoreInterface.reportPluginError}
+   */
+  pluginErrorHandler: (manifest: PluginManifest, errorDetails: PluginErrorDetails) => void;
 }>;
 
 /**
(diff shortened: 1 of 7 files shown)


packages/lib-core/src/runtime/PluginStore.ts
```diff
@@ -12,6 +12,16 @@ import { decodeCodeRefs, getPluginModule } from './coderefs';
...
 import type { PluginLoaderOptions } from './PluginLoader';
 
+export type PluginErrorDetails = { message: string; cause?: unknown } & (
+  | {
+      loadError: true;
+    }
+  | {
+      loadError: false;
+      reportedBy: string;
+    }
```
florkbr, line 22: I see what we're trying to do here (`reportedBy` is not needed when `loadError` is true as an error occurs loading the plugin). What are you're thoughts on this alternative approach?
```js
export type PluginLifecycle = 'load' | 'reload' | 'run' | 'destroy';

export type PluginErrorDetails = {
  message: string;
  lifecycle: PluginLifecycle;
  cause?: unknown;
  reportedBy?: string;
};
```

packages/lib-core/src/runtime/coderefs.ts
```diff
@@ -126,18 +116,22 @@ export const decodeCodeRefs = (extension: LoadedExtension, entryModule: PluginEn
...
  *
- * This is an asynchronous operation that completes when all of the associated
- * Promises are either resolved or rejected. If there were no resolution errors,
- * the resulting Promise resolves with the updated extension.
+ * The resulting Promise resolves with the updated extension; its `properties` object
+ * is cloned in order to preserve the original structure of decoded extensions.
  *
- * The updated extension is a new object; its `properties` are cloned in order
- * to preserve the original structure of decoded extensions.
+ * The resulting Promise never rejects. Use the `onResolutionErrors` callback to handle
```
florkbr, line 129: @vojtechszocs assuming this will need a major version bump as it is a breaking change?

packages/lib-core/src/runtime/PluginStore.ts
```diff
@@ -26,6 +36,19 @@ export type PluginStoreOptions = Partial<{
...
+   *
+   * The `errorDetails.loadError` property indicates the type of the error:
+   * - if `true`, the error has occurred while loading the given plugin
+   * - if `false`, the error was reported for an already loaded plugin
+   *
+   * By default, this function does nothing.
+   *
+   * @see {@link PluginStoreInterface.reportPluginError}
+   */
+  pluginErrorHandler: (manifest: PluginManifest, errorDetails: PluginErrorDetails) => void;
```
florkbr, line 51: Might be worth having an example of this in the sample-app/plugin.

vojtechszocs:
Closing in favor of #241

PR #232, opened 2023-08-25
RHCLOUD-27198: Improve Cypress testing of sample plugin host application

vojtechszocs:
### Summary of changes:
- [x] bump Cypress to version 12 and scope all Cypress tests to sample app package
- [x] improve existing Cypress E2E test for sample app (load sample plugin and assert plugin runtime information)
- [x] add Cypress component tests using `TestPluginStore` for React components that consume plugin information
- [x] it is now possible to pass custom `PluginLoader` implementation to `PluginStore` via the `PluginLoaderInterface`
- [x] add "Extensions" column to `PluginInfoTable` showing the number of extensions brought in by the given plugin

### Notes on Cypress component testing
We recommend using a single `TestPluginStore` instance for all your tests. This can be done via the corresponding [Cypress support file](https://docs.cypress.io/guides/component-testing/styling-components#Component-Support-File) which is typically used for one-time preparation of your testing environment.
```tsx
let pluginStore: TestPluginStore;

beforeEach(() => {
  pluginStore = new TestPluginStore();
});

Cypress.Commands.add('mount', (component, options = {}) => {
  return mount(<PluginStoreProvider store={pluginStore}>{component}</PluginStoreProvider>, options);
});

Cypress.Commands.add('getPluginStore', () => {
  return cy.wrap(pluginStore);
});
```

In your component test, you can then just `cy.mount` the given component and use `cy.getPluginStore` command to access the `TestPluginStore` instance.

```tsx
describe('FooComponent', () => {

  it('should work', () => {
    // renders inside <PluginStoreProvider> automatically
    cy.mount(<FooComponent />);

    // add plugins, enable them, etc.
    cy.getPluginStore().then((pluginStore) => {
      pluginStore.addLoadedPlugin(mockPluginManifest({ name: 'test' }), mockPluginEntryModule());
      pluginStore.enablePlugins(['test']);
    });

    // assertions go here
    cy.get('[data-test-id="foo"]').should('contain.text', 'Foo Text');

    // you can also use "cy.getPluginStore" to assert the current state of TestPluginStore
  });

});
```

Check out [this article](https://www.awesome-testing.com/2023/03/understanding-cypress) to understand the sync / async duality of Cypress and how it transforms and runs your tests.


packages/lib-core/src/index.ts
```diff
@@ -37,8 +37,12 @@ export {
   useResolvedExtensions,
   UseResolvedExtensionsResult,
 } from './runtime/useResolvedExtensions';
-export { useFeatureFlag, UseFeatureFlagResult } from './runtime/useFeatureFlag';
 export { usePluginInfo } from './runtime/usePluginInfo';
+export { useFeatureFlag, UseFeatureFlagResult } from './runtime/useFeatureFlag';
+
+// Testing utilities
+export { TestPluginStore } from './testing/TestPluginStore';
+export { mockPluginManifest, mockPluginEntryModule } from './testing/mocks';
```
florkbr, line 45: NIT - maybe we could rename this directory `test-fixtures`?

packages/lib-core/src/testing/mocks.ts
```diff
@@ -0,0 +1,31 @@
+import type { AnyObject } from '@monorepo/common';
+import { noop } from 'lodash';
+import type { PluginManifest } from '../types/plugin';
+import type { PluginEntryModule } from '../types/runtime';
+
+export const mockPluginManifest = ({
```
florkbr, line 6: Are these 2 utilities something we really want to expose/maintain? I'd imagine most apps/plugins consuming the SDK and writing tests will be providing their own test manifest/entry module.

packages/sample-app/cypress.config.ts
```diff
@@ -0,0 +1,24 @@
...
+  video: false,
+
+  component: {
+    devServer: {
+      framework: 'react',
+      bundler: 'webpack',
+      webpackConfig,
+    },
+    specPattern: 'src/components/**/*.cy.{js,jsx,ts,tsx}',
+    indexHtmlFile: 'src/cypress/component-index.html',
```
florkbr, line 15: NIT: Wondering if it makes sense to put the `cypress` and `e2e` directories in the `src` folder or if we should move them out to top level directories.

packages/lib-core/src/testing/mocks.ts
vojtechszocs: That's a good question.

Mocking a plugin manifest (like in `mockPluginManifest`) is very simple so this could be offloaded to consumers.

Mocking an entry module (like in `mockPluginEntryModule`) is a bit more complex, but still most typically a simple transformation from static `pluginModules: { [moduleRequest: string]: AnyObject }` declaration to SDK expected `get` function `(moduleRequest: string) => Promise<() => AnyObject>` - I don't think it makes sense to test asynchronicity within that function, so the resulting Promise would most typically be `Promise.resolve(() => pluginModules[moduleRequest]` or similar.

Overall, I think you're right here. We don't have to expose these functions.

packages/sample-app/cypress.config.ts
vojtechszocs: We currently put all the source code in `src` directory per each package.

I don't see much value in creating sibling source directories scoped to specific tools or purposes - having all the sources under one directory seems simpler to me. (Test code is source code too, but oftentimes get treated as second class citizen in the codebase.)

AFAK there is no special treatment necessary for `src/cypress` directory - its code is used in the Cypress component testing webpack build, so the usual context applies (React code to be part of the resulting web application).

I've placed `component-index.html` into the `cypress` directory to co-locate the HTML template with associated setup code. This is similar to `src/app-index.html.ejs` and `src/app-styles.css` sitting next to `src/app.tsx`.

Note to self: `app.css` could be inlined to `app-styles.css` and the latter renamed to `app.css`.

As for `src/e2e` - this was meant to contain all E2E tests. If you think that e.g. `src/app.e2e.cy.ts` is better, I can change that.

florkbr (approved):
I've checked these changes out locally and ran the sample app/plugin, builds, and tests. Things appear to be functioning as expected. :shipit:

PR #241, opened 2023-10-19
Improve core logic and useResolvedExtensions hook

vojtechszocs:
Notable changes
- update `resolveCodeRefValues` function to never reject and use `onResolutionErrors` callback instead
- add options to `useResolvedExtensions` hook for better control over code reference resolution errors
- remove "disable associated plugins on code reference resolution errors" logic in `useResolvedExtensions` hook


florkbr (approved):

florkbr:
Overall looks good to me and should allow applications to decide if a plugin should be disabled :shipit:

PR #246, opened 2023-11-03
Revert base image to name:tag in CI buildroot dockerfile

vojtechszocs:
It seems that Quay repo doesn't support referencing older container image builds via `name@sha256:digest` so we switch back to using `name:tag` again.


jhadvig (approved):
/lgtm

PR #247, opened 2023-11-09
Remove UI components and PF dependency from lib-utils

florkbr:
This PR will remove the UI components exposed by the SDK lib-utils package.

Also removes global storybook config/install as it is no longer needed.

See https://github.com/patternfly/react-component-groups.

After discussion we have decided apps and plugins can pull UI components from the PF react-component-groups repo (and decide how to handle versioning or de-duping of the dependency). Several of the UI components removed in this PR already exist in the new location and will continue to be built upon.


package.json
```diff
@@ -37,15 +36,7 @@
...
-    "@storybook/addon-actions": "^6.5.9",
-    "@storybook/addon-docs": "^6.5.9",
-    "@storybook/addon-essentials": "^6.5.9",
-    "@storybook/addon-interactions": "^6.5.9",
-    "@storybook/addon-links": "^6.5.9",
-    "@storybook/builder-webpack5": "^6.5.9",
-    "@storybook/manager-webpack5": "^6.5.9",
-    "@storybook/react": "^6.5.9",
-    "@storybook/testing-library": "^0.0.13",
+    "@testing-library/dom": "^9.3.3",
```
florkbr, line 39: @vojtechszocs FYI yarn was complaining about this missing peer dependency (being requested by `@testing-library/user-event` from one of the other `@testing-library` libs below).

packages/lib-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@openshift/dynamic-plugin-sdk-utils",
-  "version": "4.1.0",
+  "version": "5.0.0",
```
florkbr, line 3: @vojtechszocs While none of the UI components being removed seem to be used in production environments as far as we can tell - I am proposing we bump a major version of the lib-utils package in case we missed any locations.

vojtechszocs: Sounds good, we can also publish that new version to npmjs repository once the PR gets merged.

package.json
vojtechszocs: AFAIK, we only used React testing library due to lib-utils having React components for which we had unit tests.

I'm not sure we still need it, will check.

PR #250, opened 2023-11-23
Fix asset hash related bug in GenerateManifestPlugin

vojtechszocs:
This PR fixes a bug that occurs when using `[contenthash]` placeholder within the generated entry script filename.

webpack compilation hook [`processAssets`](https://webpack.js.org/api/compilation-hooks/#processassets) stage `SUMMARIZE` runs before asset hash optimizations, therefore we need to use a later stage like `ANALYSE`.

Also, when running webpack in production mode, the generated plugin manifest JSON is no longer pretty formatted.


Hyperkid123 (approved):
/lgtm

PR #253, opened 2024-01-08
Always transform plugin manifest before loading the associated plugin

vojtechszocs:
This PR ensures that `transformPluginManifest` function is always called within `PluginStore.loadPlugin` in order to allow pre-processing of all manifest objects.


florkbr:
@vojtechszocs sorry for the delay. Looks good!

PR #254, opened 2024-01-10
Remove loadError state if modify action is fired

rottencandy:
Fixes https://issues.redhat.com/browse/RHCLOUD-26143

Always update the state to clear loadError if an update action is fired but loadError is not empty in state.

### DIFF
--- a/packages/lib-utils/src/app/redux/reducers/k8s/k8s.ts
+++ b/packages/lib-utils/src/app/redux/reducers/k8s/k8s.ts
@@ -204,6 +204,15 @@ export const sdkK8sReducer = (state: K8sState, action: K8sAction): K8sState => {
       });
     }
 
+    case ActionType.ClearError: {
+      const { k8sObjects, id } = action.payload;
+      return state.mergeIn([id], {
+        loadError: '',
+        loaded: true,
+        data: k8sObjects,
+      });
+    }
+
     case ActionType.StopWatchK8s:
       return state.delete(action.payload.id);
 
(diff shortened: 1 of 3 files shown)


rottencandy:
@rohitkrai03 ^

packages/lib-utils/src/app/redux/reducers/k8s/k8s.ts
```diff
@@ -192,7 +192,8 @@ export const sdkK8sReducer = (state: K8sState, action: K8sAction): K8sState => {
       if (currentJS.toJSON) {
         currentJS = currentJS.toJSON();
         currentJS.metadata.resourceVersion = k8sObjects?.metadata?.resourceVersion;
-        if (isEqual(currentJS, k8sObjects)) {
+        const loadError = state.getIn([action.payload.id, 'loadError']);
```
florkbr, line 195: @rottencandy we should use `id` in lieu of `action.payload.id` (already destructured here: https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/lib-utils/src/app/redux/reducers/k8s/k8s.ts#L189)

Stepping back, this seems odd that we are indirectly using this reducer to modify the `loadError` state. This reducer is primarily concerned with handling modified `K8sObjects`.

Would it make more sense to add a new reducer to clear the error? Something like `ActionType.ClearError`? If our goal here is to prevent overriding the state when an error is present, I think we should find where `ActionType.ModifyObject` is dispatched and update to check if there is a `loadError` present before dispatching?

rottencandy: Thanks, having a separate reducer to clear the error sounds much better. Updated.

rohitkrai03:
@florkbr Can we get this PR merged if everything looks good?

rohitkrai03:
@vojtechszocs @florkbr can someone take a look at this PR? Its been stuck for several months now.

PR #259, opened 2024-04-22
Allow overriding core webpack plugins used by DynamicRemotePlugin

vojtechszocs:
This PR adds new `moduleFederationSettings.pluginOverride` option to webpack `DynamicRemotePlugin`.

This allows us to use libraries like [`@module-federation/enhanced`](https://github.com/module-federation/universe/tree/main/packages/enhanced) for better interoperability with tools like [Rspack](https://www.rspack.dev/).

For example:

```ts
import { ModuleFederationPlugin, ContainerPlugin } from '@module-federation/enhanced';

// in your webpack plugins declaration
new DynamicRemotePlugin({
  pluginMetadata,
  extensions,
  moduleFederationSettings: {
    // use webpack plugins from @module-federation/enhanced package (as imported above)
    pluginOverride: { ModuleFederationPlugin, ContainerPlugin }
  }
})
```

Related to https://github.com/scalprum/scaffolding/pull/127 - cc @Hyperkid123

### DIFF
--- a/packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
+++ b/packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
@@ -51,6 +51,27 @@ export type PluginModuleFederationSettings = Partial<{
    * @see https://webpack.js.org/plugins/module-federation-plugin/#sharescope
    */
   sharedScopeName: string;
+
+  /**
+   * Override the implementation of core webpack module federation plugin(s) to use
+   * for the webpack build.
+   *
+   * If not specified, standard webpack provided implementation of the given plugin
+   * will be used.
+   *
+   * @see https://github.com/module-federation/universe/tree/main/packages/enhanced
+   */
+  pluginOverride: Partial<{
+    /**
+     * Custom `ModuleFederationPlugin` implementation.
+     */
+    ModuleFederationPlugin: typeof container.ModuleFederationPlugin;
+
+    /**
+     * Custom `ContainerPlugin` implementation.
+     */
+    ContainerPlugin: typeof container.ContainerPlugin;
+  }>;
 }>;
 
 /**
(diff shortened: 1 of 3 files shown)


Hyperkid123 (approved):

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -195,6 +216,11 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
     const moduleFederationLibraryType = moduleFederationSettings.libraryType ?? 'jsonp';
     const moduleFederationSharedScope = moduleFederationSettings.sharedScopeName ?? 'default';
 
+    const {
+      ModuleFederationPlugin = container.ModuleFederationPlugin,
+      ContainerPlugin = container.ContainerPlugin,
+    } = moduleFederationSettings.pluginOverride ?? {};
```
florkbr, line 222: @vojtechszocs thoughts about writing this as:
```js
const {
  ModuleFederationPlugin,
  ContainerPlugin
} = moduleFederationSettings.pluginOverride || container;
```
?

vojtechszocs: @florkbr I've checked the above code - it would result in both `ModuleFederationPlugin` and `ContainerPlugin` variables being possibly `undefined` so we'd still need the default assignments within that destructure statement, which then beats the `|| container` part :smiley: 

The reason for destructuring `moduleFederationSettings.pluginOverride ?? {}` instead of just `moduleFederationSettings.pluginOverride` is that the `pluginOverride` object itself can be `undefined` - TypeScript strict mode detects this as well, raising type errors if the `?? {}` part is missing.

TL;DR - I'd keep the original code.

JS destructuring assignment works with a single object getting picked from - imagine if `ModuleFederationPlugin` option was defined within the `pluginOverride` object but `ContainerPlugin` was not. In such case, the `|| container` part would have no effect.

florkbr: Chatted with @vojtechszocs offline. My main concern here was readability - though we do have some limitations with object destructuring in JS. The alternative approach is not much clearer than what we have here - so I'm fine keeping as is for now.

PR #266, opened 2025-04-23
update dependencies and remove stale ones

christianvogt:
With the removal of PatternFly, there are several more dependencies that can be cleaned up from the `utils` package.

`react-redux` `v8` is [backwards compatible](https://github.com/reduxjs/react-redux/releases/tag/v8.0.0) with `v7`. Therefore I updated the `peerDependencies` to `^7 || ^8`


florkbr:
Prow is still failing due to mirror issues. We will eventually need to follow up on the pipeline. For now I am running all relevant commands locally.

✅ Pipeline passes locally!

[dynamic-plugin-sdk-pr-266-local.log](https://github.com/user-attachments/files/19871231/dynamic-plugin-sdk-pr-266-local.log)

PR #273, opened 2025-09-05
CONSOLE-4621: Add support for optional dependencies

vojtechszocs:
This PR adapts openshift/console#15183 changes to support optional dependencies.

## Summary

Dynamic plugins may depend on
- other dynamic plugins - **regular dependencies**
- environment specific constants - **environment dependencies** (*)

(*) Provided by host application via `PluginLoader` option `fixedPluginDependencyResolutions`

Every dependency (regular or environment related) may be declared as
- **required** - using `dependencies` object in build metadata
- **optional** - using `optionalDependencies` object in build metadata

### Resolution logic

Failure to resolve all of the plugin's required dependencies will cause that plugin to load with failed status.

Failure to resolve any of the plugin's optional dependencies will not prevent that plugin from loading, i.e. await optional dependency resolution but don't fail if the resolution was not a success.

:memo: If the optional dependency resolution was a success but the resolved version does not match the plugin's expectation, that plugin will fail to load with the following error message:
```
`Dependency ${depName} not met: required range ${versionRange}, resolved version ${res.version}`
```

### Plugin A has optional dependency on plugin B
- plugin B loads with error :arrow_forward: plugin A loads OK :white_check_mark:
- plugin B loads successfully and its version matches the expected range :arrow_forward: plugin A loads OK :white_check_mark:
- plugin B loads successfully and its version does not match the expected range :arrow_forward: plugin A fails to load :x:

### Plugin A has optional dependency on environment constant C
- version of constant C matches the expected range :arrow_forward: plugin A loads OK :white_check_mark:
- version of constant C does not match the expected range :arrow_forward: plugin A fails to load :x:

### DIFF
--- a/packages/lib-core/src/runtime/PluginLoader.ts
+++ b/packages/lib-core/src/runtime/PluginLoader.ts
@@ -351,7 +351,8 @@ export class PluginLoader implements PluginLoaderInterface {
   private resolvePluginDependencies(manifest: PluginManifest) {
     return new Promise<void>((resolve, reject) => {
       const pluginName = manifest.name;
-      const dependencies = manifest.dependencies ?? {};
+      const requiredDependencies = manifest.dependencies ?? {};
+      const optionalDependencies = manifest.optionalDependencies ?? {};
       const semverRangeOptions: semver.RangeOptions = { includePrerelease: true };
 
       let isResolutionComplete = false;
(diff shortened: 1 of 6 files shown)


logonoff:
/retitle CONSOLE-4621: Add support for optional dependencies

logonoff (commented):
maybe we can add an e2e test and/or some unit tests for your test cases in the pr description?

logonoff:
> We should totally add some unit tests at some point

Do we have unit tests that we can easily port over from openshift/console? If not I am okay with addressing this as a follow up

vojtechszocs:
Changes verified locally using the test cases in PR description.

florkbr (approved):
LGTM :shipit: 

@Hyperkid123 FYI this is new functionality we can leverage in the SDK

PR #285, opened 2025-12-16
feat: add rspack support

logonoff:
Turns out rspack was not working for DynamicRemotePlugin due to a bug upstream: https://github.com/web-infra-dev/rspack/issues/14277

The bug is fixed in [2.0.7](https://github.com/web-infra-dev/rspack/releases/tag/v2.0.7). The only remaining compatibility mismatch is that webpack's `ModuleFederationPlugin` is renamed to `ModuleFederationPluginV1` in rspack. 

This PR adds detection for rspack and changes the ModuleFederationPlugin implementation to use `ModuleFederationPluginV1`, as expected


packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -210,19 +210,26 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
     const moduleFederationLibraryType = moduleFederationSettings.libraryType ?? 'jsonp';
     const moduleFederationSharedScope = moduleFederationSettings.sharedScopeName ?? 'default';
 
+    const isRspack = 'rspack' in compiler;
+    const jsonp = moduleFederationLibraryType === 'jsonp';
+
     const {
-      ModuleFederationPlugin = compiler.webpack.container.ModuleFederationPlugin,
+      ModuleFederationPlugin = isRspack
+        ? (compiler as any).rspack.container.ModuleFederationPluginV1
```
vojtechszocs, line 219: This ternary branch value computed from `any` cast will result in `ModuleFederationPlugin | any` value type, correct?

Does it make sense to import Rspack types to avoid the `any` cast?

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -210,19 +210,26 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
...
+        : compiler.webpack.container.ModuleFederationPlugin,
       ContainerPlugin = compiler.webpack.container.ContainerPlugin,
     } = moduleFederationSettings.pluginOverride ?? {};
 
     const entryCallbackName = entryCallbackSettings.name ?? DEFAULT_REMOTE_ENTRY_CALLBACK;
     const entryCallbackPluginID = entryCallbackSettings.pluginID ?? pluginMetadata.name;
 
-    const jsonp = moduleFederationLibraryType === 'jsonp';
+    // Rspack does not support jsonp library type for Module Federation;
```
vojtechszocs: Are we 100% sure on this?

https://rspack.rs/config/output#outputlibrarytype - `jsonp` is listed there

https://rspack.rs/config/output#other-types - `jsonp` is used in the example

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -210,19 +210,26 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
...
+    // use var instead and append the callback invocation via PatchEntryCallbackPlugin
+    const rspackVarName = containerName.replace(/[^a-zA-Z0-9_$]/g, '_');
```
vojtechszocs: Can you trace the actual Rspack code that computes the var name to ensure we're aligned with that?

packages/sample-app/rspack.config.ts
vojtechszocs, line 1: We should generally try to avoid duplication and import and reuse stuff from webpack config if possible.

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
logonoff: Rspack and webpack compiler types are close but incompatible. I did a quick skim of a few rspack-compatible webpack plugins and none of them attempt to use both rspack and webpack types in the same plugin

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
logonoff: This was previously broken, should be fixed upstream in v2.0.7. See https://github.com/web-infra-dev/rspack/issues/14277

packages/lib-webpack/src/webpack/DynamicRemotePlugin.ts
```diff
@@ -210,8 +210,14 @@ export class DynamicRemotePlugin implements WebpackPluginInstance {
...
     const moduleFederationSharedScope = moduleFederationSettings.sharedScopeName ?? 'default';
 
+    // Rspack's renamed ModuleFederationPlugin to ModuleFederationPluginV1.
+    // Their ModuleFederationPlugin implements module federation 1.5+
+    const isRspack = 'rspack' in compiler;
+
     const {
-      ModuleFederationPlugin = compiler.webpack.container.ModuleFederationPlugin,
+      ModuleFederationPlugin = isRspack
+        ? (compiler as any).rspack.container.ModuleFederationPluginV1
```
charlesmulder, line 219: Unsafe cast to `any` - no guard if rspack structure changes. Use optional chaining or check existence.

```ts
const {
    ModuleFederationPlugin = isRspack
      ? compiler.rspack?.container?.ModuleFederationPluginV1
      : compiler.webpack.container.ModuleFederationPlugin,
    ContainerPlugin = compiler.webpack.container.ContainerPlugin,
  } = moduleFederationSettings.pluginOverride ?? {};

  if (!ModuleFederationPlugin) {
    throw new Error(
      `${isRspack ? 'rspack' : 'webpack'}.container.ModuleFederationPlugin${isRspack ? 'V1' : ''} not found`
    );
  }
```

logonoff: The guard is `const isRspack = 'rspack' in compiler;`. We can be pretty certain that api will be stable: https://rspack.rs/plugins/webpack/module-federation-plugin-v1

Any cast is pragmatic - rspack and webpack types are structurally similar but incompatible.

charlesmulder (approved):

TheRealJon (commented):
I pulled this down and ran the test script with no issues, so with the caveat that I'm not very knowledgable when it comes to this repo or rspack...

/lgtm
/verified by @TheRealJon

logonoff:
not needed for merging

/remove-label verified

PR #313, opened 2026-04-10
Modify useResolvedExtensions hook

vojtechszocs:
### lib-core

- Modify `useResolvedExtensions` hook to accept `extensions` array as a parameter
- Expose `visitDeep` utility function

### sample-app

- Add `useReplaceTextExtensions` hook to showcase custom extension transformations


logonoff (approved):
/lgtm

PR #317, opened 2026-05-20
Upgrade TypeScript to latest 5.x and Jest to latest 29.x

vojtechszocs:
- bump TypeScript to 5.9.3
- bump Jest to 29.7.0
  - :memo: can't use Jest 30 due to `ts-jest` issue https://github.com/kulshekhar/ts-jest/discussions/4625
- bump Jest related packages
  - `ts-jest` to 29.4.10
  - `jest-axe` to 10.0.0
- bump `@types/jest` to latest 29.x
  - :memo: we still need `@types/jest` for global test types like `describe`, `it`, `expect`, `jest.fn` etc.
- add `jest-environment-jsdom`
  - :memo: starting with Jest 28 this needs to be an explicit dependency
- update `jest.mocked` calls in tests
  - :memo: using `jest.mocked(fn)` since in Jest 29 shallow mocking is the default setting
- bump API Extractor & API Documenter to latest 7.x
  - :memo: fix top-level `api-extractor` script to refer to root `node_modules/typescript/lib`
- bump `ts-node` to latest 10.x


package.json
```diff
@@ -20,13 +20,13 @@
     "sample-app-install-cypress": "yarn workspace @monorepo/sample-app run cypress install",
     "sample-app-test-component": "yarn workspace @monorepo/sample-app run cypress run --component",
     "sample-app-test-e2e": "yarn workspace @monorepo/sample-app run cypress run --e2e",
-    "api-extractor": "api-extractor run --typescript-compiler-folder ./node_modules/typescript/lib",
```
vojtechszocs, line 23: Note - when running e.g. `(cd packages/lib-core ; yarn api-extractor)` this has resulted in
```
--typescript-compiler-folder packages/lib-core/node_modules/typescript/lib
```
which does not exist, so we're using `$PROJECT_CWD` which always points to workspace root directory.

logonoff (approved):
/lgtm

PR #318, opened 2026-05-27
feat: add `@openshift/dynamic-plugin-sdk-types`

logonoff:
Implements step 1 of https://github.com/openshift/enhancements/blob/master/enhancements/console/generated-resource-types.md

Port of https://github.com/logonoff/openshift-types over to `dynamic-plugin-sdk` repo.

Build scripts are semi-bespoke so there is some duplication when generating build-metadata and such.

Currently OCP console and lib-utils do not have fully described types, this package fixes that by describing these types

`lib-utils` now has this package as a peer dependency and uses it for `K8sResourceCommon` and `OwnerReference` types


packages/lib-types/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "module": "nodenext",
+    "target": "es2024",
```
vojtechszocs, line 4: Note that the current `compilerOptions.target` option for all existing lib packages is `es2021` which gives us some (reasonable) degree of down-leveling in generated JS code for compatibility with existing projects.

I think `es2024` is a bit too new'ish but we can change it anytime so not an issue for this PR.

packages/lib-types/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "sourceMap": true,
+    "noUnusedLocals": true,
+    "strict": true,
+    "skipLibCheck": true,
```
vojtechszocs, line 8: Some of these settings are already provided in our base TSConfig files, that said we can tweak this later on.

packages/lib-types/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Package structure
+
+```
+dist/
+  index.d.ts                     # Entry point — re-exports everything
+  K8sResourceCommon.d.ts          # Base K8s interfaces (K8sResourceCommon, ObjectMetadata, OwnerReference, etc.)
+  build-metadata.json             # Build info (version, date, source commits)
+
+  kubernetes/                     # Types generated from the OpenShift fork of the K8s swagger spec
```
vojtechszocs:
```suggestion
  kubernetes/                     # Types generated from the OpenShift fork of the Kubernetes API swagger spec
```

packages/lib-types/README.md
```diff
@@ -0,0 +1,37 @@
+# `@openshift/dynamic-plugin-sdk-types`
+
+> TypeScript types generated from OpenShift and Kubernetes APIs.
+
+## Package structure
+
+```
+dist/
+  index.d.ts                     # Entry point — re-exports everything
+  K8sResourceCommon.d.ts          # Base K8s interfaces (K8sResourceCommon, ObjectMetadata, OwnerReference, etc.)
```
vojtechszocs: In future, consider naming it `k8s-common.d.ts` or similar since it has more types than just `K8sResourceCommon`.

packages/lib-types/README.md
```diff
@@ -0,0 +1,37 @@
...
+    <group>/                      # API group (e.g. config.openshift.io, console.openshift.io)
+      <version>/                  # API version (e.g. v1, v1alpha1)
+        <Kind>.d.ts               # One file per CRD kind (e.g. Route.d.ts, ConsolePlugin.d.ts)
+```
+
+All generated types extend `K8sResourceCommon` and are organized by platform, API group, and version. Each resource kind gets its own `.d.ts` file with a `<Kind>Kind` interface name (e.g. `PodKind`, `DeploymentKind`).
+
+## Source commits
+
+The upstream commits used for type generation are tracked in [`sources.json`](./sources.json):
```
vojtechszocs:
```suggestion
The upstream git commits used for type generation are tracked in [`sources.json`](./sources.json):
```

packages/lib-types/tsconfig.json
logonoff: Any code imported here gets deleted during compilation because it's all types

packages/lib-types/collect-crds.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env sh
+
+set -euo pipefail
+
+OPENSHIFT_API_COMMIT=$(node -p "require('./sources.json')['openshift/api']")
```
vojtechszocs: Nit - for simple JSON parsing we could use `jq``
```sh
jq -r '."openshift/api"' sources.json
```

packages/lib-types/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@openshift/dynamic-plugin-sdk-types",
```
vojtechszocs, line 2: I think package name should reflect the fact that we're providing TS types for OpenShift and K8s APIs.

We should double check this before we start to publish and use this package.

packages/lib-types/src/smoke.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  it('dist/index.d.ts has the build banner', () => {
+    const content = readFileSync(resolve(distDir, 'index.d.ts'), 'utf-8');
+    expect(content).toMatch(/^\/\*\*/);
+    expect(content).toContain('OpenShift Dynamic Plugin SDK');
+    expect(content).toContain('packageName');
+  });
+});
+
+describeIfDist('build-metadata.json', () => {
```
vojtechszocs, line 51: In future, we can move this test to `common` and ensure that all `lib-*` packages have the expected `dist/build-metadata.json` structure.

packages/sample-app/src/cypress/component-index.html
```diff
@@ -1,8 +1,8 @@
-<!DOCTYPE html>
+<!doctype html>
 <html>
   <head>
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta charset="utf-8" />
```
vojtechszocs, line 4: HTML `<meta>` is a [void element](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) and therefore does not need to be "closed" as XML elements (?)

vojtechszocs (changes requested):

packages/sample-app/src/cypress/component-index.html
logonoff: Prettier did that

packages/lib-types/package.json
logonoff: we could do `@openshift/dynamic-plugin-sdk-k8s-types`..? or maybe even `@openshift/k8s-resource-types` if you're willing to not prefix the package with `dynamic-plugin-sdk`

packages/sample-app/src/cypress/component-index.html
vojtechszocs: OK, let's keep it consistent with Prettier conventions.

PR #319, opened 2026-06-04
Small tweaks to generated API type package

vojtechszocs:
- rename `packages/lib-types` to `packages/lib-api-types`
- package name changed to `@openshift/api-types`


logonoff (commented):
/lgtm

PR #325, opened 2026-07-01
Handle plugin manifest object as immutable

vojtechszocs:
Calling `PluginStore.add{Pending,Loaded,Failed}Plugin` methods will deep freeze the manifest object.

This reflects the expectation that plugin manifest should be treated as read-only (immutable) by consumers.

Attempts to modify parts of the plugin manifest should now fail as per [`Object.freeze`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) contract.

Objects returned from `PluginStore.getPluginInfo` have the `manifest` typed as `DeepReadonly<PluginManifest>`.

#### `DeepReadonly` type check
```ts
type Foo = { a: string; b: { c: number; d: number[] } };
const foo: DeepReadonly<Foo> = { a: 'test', b: { c: 0, d: [1, 2] } };

foo.a = 'x';    // Cannot assign to 'a' because it is a read-only property.
foo.b = {};     // Cannot assign to 'b' because it is a read-only property.
foo.b.c = 1;    // Cannot assign to 'c' because it is a read-only property.
foo.b.d = [];   // Cannot assign to 'd' because it is a read-only property.
foo.b.d[0] = 1; // Index signature in type 'DeepReadonlyArray<number>' only permits reading.
```


packages/common/src/types/objects.ts
logonoff, line 1: Much more simple 👍

logonoff (approved):

PR #326, opened 2026-07-09
Replace Yup with Zod

vojtechszocs:
The project uses [Yup](https://github.com/jquense/yup) version `1.7.1` for runtime validation of plugin manifests and build configurations. Yup lacks native support for map/record structures, forcing a verbose custom `recordStringStringSchema` definition using `test()` with manual key/value iteration. [Zod](https://github.com/colinhacks/zod) has first-class `record()` support, better TypeScript inference, and is the more actively maintained choice. The schemas are **internal** (not exported from package index), so this is not a breaking public API change.

```ts
// Before (Yup)
schema.validateSync(obj, { strict: true, abortEarly: false });

// After (Zod)
schema.strict().parse(obj);
```
- Yup's `strict: true` prevents type coercion; Zod never coerces by default.
- Yup's `abortEarly: false` collects all errors; Zod always collects all errors.
- Using `strict()` on the Zod schema rejects unknown keys.


README.md
```diff
@@ -5,31 +5,36 @@
...
 Both host applications and plugins can be built, released and deployed independently from each
 other. This reduces the coupling between an application and its plugins and allows individual
 plugins to be updated and deployed more frequently.
 
-Tools provided by this SDK are [React](https://reactjs.org/) focused. Host applications are React
-web applications built with [webpack](https://webpack.js.org/). Plugins are built with webpack and
-use `DynamicRemotePlugin` to generate their assets. Both host applications and plugins use React APIs
-(hooks, components, etc.) to extend the functionality of host application(s) and/or other plugins.
+Tools provided by this SDK are [React](https://reactjs.org/) focused and support [webpack][webpack]
+and [rspack][rspack] module bundlers. Plugins use `DynamicRemotePlugin` to generate their assets.
```
logonoff:
```suggestion
Tools provided by this SDK are [React](https://reactjs.org/) focused and support [webpack]
and [rspack] module bundlers. Plugins use `DynamicRemotePlugin` to generate their assets.
```

logonoff (approved):
/lgtm

PR #333, opened 2026-07-29
Remove uuid dependency in lib-core and lib-utils

vojtechszocs:
## Summary

Remove [uuid](https://github.com/uuidjs/uuid) dependency in lib-core and lib-utils packages.

`uuid` method `v4` is replaced with [`crypto.getRandomValues`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) based utility function.

## Motivation

`uuid` has a known security vulnerability `CVE-2026-41907` ([1](https://github.com/advisories/GHSA-w5hq-g745-h8pq)) ([2](https://nvd.nist.gov/vuln/detail/CVE-2026-41907)) which impacts its `v3`, `v5` and `v6` [API methods](https://github.com/uuidjs/uuid#api-summary).

Both lib-core and lib-utils packages currently use `uuid` method `v4` only - not impacted by this CVE.

It's best to remove `uuid` dependency and use standard Crypto API to generate UUID v4 compatible values.


vojtechszocs:
Actually we don't need `uuid` and can use different methods which would allow us to drop this dependency.

/hold

logonoff (approved):
/lgtm

logonoff (approved):
/lgtm

PR #334, opened 2026-07-30
Remove absolute URL validation for RemotePluginManifest.baseURL

vojtechszocs:
We need to support relative URL paths in `RemotePluginManifest.baseURL` values.

`ZodString.url()` only supports absolute URL values as per `new URL()` constructor.

Writing a custom `absoluteOrRelativeURLSchema` with implementation like
```ts
(value) => URL.canParse(value) || URL.canParse(value, 'http://localhost/')
```
seems to be pointless - when a base URL is provided, `URL.canParse(value, 'http://localhost/')` is extremely permissive, even strings like `foo bar` are accepted and auto-encoded as pathname `/foo%20bar`.


logonoff (approved):
/lgtm

